PR #16, opened 2024-05-09
Add weka as a new S3-like scheme

2015aroras:
Weka is the S3-like storage usage on the Jupiter cluster. This PR adds support for it in the same way that R2 is supported.


epwalsh (approved):
LGTM. We should replace the corresponding functions in `olmo.util` at some point with these.

PR #43, opened 2024-09-06
Add Lion optimizer and a "skip step" version of it

epwalsh:
I don't really intend to use the Lion optimizer, this is more meant as a proof of concept for two things:
- an efficient (no host-device sync) implementation of the "skip step" optimizer idea that @2015aroras is investigating, and
- compiling the optimizer step

### DIFF
--- a/src/olmo_core/optim/skip_step_optimizer.py
+++ b/src/olmo_core/optim/skip_step_optimizer.py
@@ -78,13 +78,12 @@ class SkipStepOptimizer(Optimizer):
         if len(self._losses) < max(20, self.rolling_interval_length // 2):
             return torch.tensor(1.0).to(device=self.latest_loss.device, non_blocking=True)
 
-        loss_std = torch.std(torch.stack(self._losses[:-1]))
+        loss_std, loss_mean = torch.std_mean(torch.stack(self._losses[:-1]))
 
         if self._grad_norms:
-            grad_norm_std = torch.std(torch.stack(self._grad_norms[:-1]))
-            return (
-                self.latest_loss <= self.sigma_factor * loss_std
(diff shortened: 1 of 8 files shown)


src/olmo_core/optim/skip_step_optimizer.py
```diff
@@ -0,0 +1,90 @@
...
+        """
+        if len(self._losses) < max(20, self.rolling_interval_length // 2):
+            return torch.tensor(1.0).to(device=self.latest_loss.device, non_blocking=True)
+
+        loss_std = torch.std(torch.stack(self._losses[:-1]))
+
+        if self._grad_norms:
+            grad_norm_std = torch.std(torch.stack(self._grad_norms[:-1]))
+            return (
+                self.latest_loss <= self.sigma_factor * loss_std
```
2015aroras: The Karpathy thing actually filters out when loss/grad norm is `sigma_factor` standard deviations above the mean (this does above 0). A number that is `s` standard deviations above the mean is said to have a "z-score" of `s`.

2015aroras (commented):
If I want to log the step_factor, do I lose all the benefits of avoiding host-device sync?

src/olmo_core/optim/skip_step_optimizer.py
epwalsh: Oh, duh, that's embarrassing after having spent 4 years in a stats PhD program

epwalsh: 82d2c52

2015aroras (approved):

src/olmo_core/optim/skip_step_optimizer.py
epwalsh: Forgot the `.abs()` but just pushed a fix for that.

PR #60, opened 2024-10-02
Add support for MoE models

epwalsh:


src/olmo_core/nn/moe/config.py
```diff
@@ -0,0 +1,211 @@
...
+    The number of experts to select for each token.
+    """
+    capacity_factor: int = 1
+    """
+    The capacity factor to use in the MoE block. Only applies if not using :data:`MoEType.dropless`.
+    """
+    bias: bool = True
+    """
+    Include bias terms.
+    """
```
Muennighoff, line 127: We didn't use bias in OLMoE (like in the dense models) so could consider setting this to False

src/olmo_core/nn/moe/layers.py
```diff
@@ -0,0 +1,62 @@
...
+        from megablocks.layers.router import (  # type: ignore
+            batched_router_zloss,
+            clear_router_zloss,
+        )
+
+        _, expert_scores = zip(*get_load_balancing_loss())
+        tokens = expert_scores[0].shape[0]
+
+        loss = batched_load_balancing_loss(self.args)
+        assert loss is not None
```
Muennighoff: I think we should also allow training without the load balancing loss but e.g. only using the z loss or neither; the OLMo PR allows that

src/scripts/train/OLMoE-1B-7B.py
```diff
@@ -0,0 +1,120 @@
...
+
+def build_trainer_config(common: CommonComponents) -> TrainerConfig:
+    return (
+        TrainerConfig(
+            save_folder=common.save_folder,
+            rank_microbatch_size=2 * 4096,
+            save_overwrite=True,
+            metrics_collect_interval=10,
+            cancel_check_interval=1,
+            z_loss_multiplier=1e-5,
```
Muennighoff, line 79: Is this the z-loss for the softmax at the output? OLMoE trained without that if so

Muennighoff (approved):
Looks amazing! Some small comments

src/olmo_core/nn/moe/layers.py
epwalsh: Fixed! I think the z-loss denominator was wrong, fixed that also.

src/scripts/train/OLMoE-1B-7B.py
epwalsh: Yes, I think we should try with both.

src/olmo_core/nn/moe/config.py
epwalsh: We do set this to false in the actual 1B-7B config I cooked up, I just set the class default to `True` here to be consistent with the default for other model config classes.

PR #62, opened 2024-10-07
include clone commands in setup steps

epwalsh:


epwalsh:
Alternative to #61 that's a little more general

AkshitaB (changes requested):
Yeah, that's helpful. We should remove the private repo error (or change it to a warning or something)

epwalsh:
Done: f1ca461

src/olmo_core/launch/beaker.py
```diff
@@ -84,6 +84,16 @@ class BeakerWekaBucket(Config):
     mount: str
 
 
+DEFAULT_SETUP_STEPS = (
+    'git clone "${REPO_URL}" .',
+    'git checkout "${GIT_REF}"',
+    "git submodule update --init --recursive",
+    "conda shell.bash activate base",
```
AkshitaB, line 91: Looks like we do expect conda. If so, can we just define `PRIVATE_REPO_SETUP_STEPS`, and then it will Just Work ™ ?

epwalsh: But this is just the default because we know conda will be on the default images.

AkshitaB: Ok, fair enough.

AkshitaB (approved):

PR #67, opened 2024-10-11
Do a dry-run batch before starting training

epwalsh:
Fixes a discrepancy after restarts caused by compiling on the first batch.


AkshitaB (approved):

epwalsh:
Well there is still some discrepancy after restarts with compile and also the "fused ops" (involving a bunch of triton kernels), but I think this is still a good feature. At least non-compiled, non-triton models restart perfectly. I'm going to continue looking into this.

PR #70, opened 2024-10-19
Adds mixing loader for FSL datasets

undfined:
This change set introduces a new mechanism for building a `NumpyFSLDataset` from a predetermined ratio of tokens per source. The UX is to supply a `SourceMixtureDatasetConfig` class to  `NumpyFSLDatasetMixtureConfig` that at data loading time counts tokens for all the sources/files provided, validates the intended ratios can be met from the source files, and builds the mixture of paths and number of instances to retain from each file that is then used to construct a typical `NumpyFSLDataset` with some modifications:
- We [limit the "available" instances per file](https://github.com/allenai/OLMo-core/pull/70/files#diff-9f53f13f74da044437605f8504f45ea399195f7410fa93208ad6da1558f8a709R1703-R1705) in the dataset to the token count per path provided from the `SourceMixtureDataset`.
- Modify how [file size is calculated](https://github.com/allenai/OLMo-core/pull/70/files#diff-9f53f13f74da044437605f8504f45ea399195f7410fa93208ad6da1558f8a709R1751-R1757) by retrieving the file size from a pre-built index from `SourceMixtureDataset`.
- Sample random instances from segmented documents [here](https://github.com/allenai/OLMo-core/pull/70/files#diff-fb05d0988fbf7dd3d1832bb18fca478e9df58c1f7c3928f28b17912ab99ef38d)

Lots of tests were added for the new mixture class and some downstream tests in numpy_dataset as well.


src/olmo_core/data/source_mixture.py
```diff
@@ -0,0 +1,307 @@
...
+from olmo_core.exceptions import OLMoConfigurationError
+from olmo_core.io import get_file_size
+
+__all__ = [
+    "SourceMixtureConfig",
+    "SourceMixtureDataset",
+    "SourceMixtureDatasetConfig",
+]
+
+logging.basicConfig(level=logging.INFO)
```
epwalsh: We shouldn't be configuring logging globally. We do have a function (`olmo_core.utils.setup_logging`) that handles this, but it's only meant to be called within the `__main__` block of a script.

src/olmo_core/data/source_mixture.py
```diff
@@ -0,0 +1,307 @@
...
+    if name in (
+        "boto",
+        "urllib3",
+        "s3transfer",
+        "boto3",
+        "botocore",
+        "aiobotocore",
+        "nose",
+    ):
+        logging.getLogger(name).setLevel(logging.CRITICAL)
```
epwalsh: This should be moved to `olmo_core.utils.setup_logging()`.

epwalsh: I would keep them at `ERROR` though instead of `CRITICAL like we do already with `urllib3`.

src/olmo_core/data/source_mixture.py
```diff
@@ -0,0 +1,307 @@
...
+        if (total := sum([source.target_ratio for source in self.source_configs])) != 1.0:
+            raise OLMoConfigurationError(f"target_ratios must sum to 1, got {total}")
+
+    def build(self) -> SourceMixtureDataset:
+        self.validate()
+        random.seed(self.seed)
+        available_tokens_by_source: Dict[str, int] = {}
+
+        print("---------------------------------------------------------")
+        print("Generating a source mixture from configurations:")
```
epwalsh: Logging > printing
```suggestion
        log.info("---------------------------------------------------------")
        log.info("Generating a source mixture from configurations:")
```

src/olmo_core/data/source_mixture.py
```diff
@@ -0,0 +1,307 @@
...
+        for source_config in self.source_configs:
+            pprint(source_config)
```
epwalsh: Our logging will actually pretty print config objects.
```suggestion
            log.info(source_config)
```

src/olmo_core/data/source_mixture.py
```diff
@@ -0,0 +1,307 @@
...
+                )
+            )
+
+        log.info("Outcome by source => ")
+        print(
+            tabulate.tabulate(
+                [item.for_table(self.max_tokens) for item in tokens_details_by_source],
+                headers="keys",
+                tablefmt="pretty",
+            ),
```
epwalsh: You can create nice looking tables with `rich`, which is already a dependency. I would prefer not to add another dependency just for pretty printing some stuff.

src/olmo_core/data/source_mixture.py
```diff
@@ -0,0 +1,307 @@
...
+
+        with ThreadPoolExecutor(max_workers=self.processes) as executor:
+            futures = []
+            for path in paths:
+                futures.append(executor.submit(self._count_tokens_for_file, path))
+
+            return sum(
+                [
+                    future.result()
+                    for future in tqdm(
```
epwalsh: You can use `rich.progress.track()` instead of `tqdm`.

src/olmo_core/data/utils.py
```diff
@@ -419,8 +421,10 @@ def segment_documents_into_instances(
     Returns the number of original documents and the number of resulting instances documents.
     """
     total_og_docs = 0
-    indices = []
+    indices: List[int] = []
     for start_idx, end_idx in iter_document_indices(path, eos_token_id=eos_token_id, dtype=dtype):
+        if max_instances and len(indices) // 2 >= max_instances:
```
epwalsh: Minor nit, but I think the semantics are more clear this way:
```suggestion
        if max_instances is not None and len(indices) // 2 >= max_instances:
```

pyproject.toml
```diff
@@ -22,6 +22,9 @@ dependencies = [
     "omegaconf",
     "safetensors",
     "importlib_resources",
+    "s3fs", # REMOVE THIS IN FAVOR OF SOMETHING CONSISTENT ELSEWHERE
+    "tabulate",
+    "tqdm",
```
epwalsh: From my other comments I don't think these are necessary.

pyproject.toml
```diff
@@ -22,6 +22,9 @@ dependencies = [
     "omegaconf",
     "safetensors",
     "importlib_resources",
+    "s3fs", # REMOVE THIS IN FAVOR OF SOMETHING CONSISTENT ELSEWHERE
```
epwalsh: If this is only used in the example then it shouldn't be a main dependency

src/examples/train_with_mixture.py
```diff
@@ -0,0 +1,242 @@
+"""
```
epwalsh: As far as I can tell the only difference between this example and the other one is the dataset? If so I don't think it's worth having this full example. Instead I would recommend just adding usage examples to the docstrings of these new classes.

epwalsh: Mainly worried that adding more example scripts will make them harder to maintain in the long run.

src/olmo_core/data/types.py
```diff
@@ -0,0 +1,40 @@
+from typing import Type, Union
+
+import numpy as np
+
+from olmo_core.config import StrEnum
+
+SupportedDType = Union[Type[np.uint8], Type[np.uint16], Type[np.uint32], Type[np.uint64]]
```
epwalsh: Maybe rename to something like `NumpyUIntTypes`?

src/olmo_core/data/numpy_dataset.py
```diff
@@ -270,9 +271,9 @@ def map(
 
         with executor_class(max_workers=max_workers) as executor:
             path_to_future = {}
-            for path in paths:
+            for idx, path in enumerate(paths):
                 if path not in path_to_future:
-                    path_to_future[path] = executor.submit(func, path)
+                    path_to_future[path] = executor.submit(func, (path, idx))
```
epwalsh: If passing the index with the path now then each `(path, idx)` pair will be unique, so we should probably process all paths and not skip duplicates?

src/olmo_core/data/numpy_dataset.py
```diff
@@ -1638,3 +1608,246 @@ def build(self) -> NumpyDatasetBase:
...
+        try:
+            path, idx = path_index
+            # Get size in bytes from tokens in the supplied index * itemsize
+            return self._path_offset_index[(str(path), idx)] * self.dtype(0).itemsize
+        except KeyError:
+            raise OLMoEnvironmentError(f"Item not found in path index @ {path_index}")
+
+
+@dataclass
+class NumpyFSLDatasetMixtureConfig(Config):
```
epwalsh: Is it necessary to have a separate config class, or can you just merge this into `NumpyDatasetConfig`?

epwalsh (commented):
Not a complete review yet but here are some high-level comments

src/olmo_core/data/numpy_dataset.py
undfined: I assume this can be merged. On it.

src/olmo_core/data/numpy_dataset.py
undfined: Yeah, ok I see that now. I was trying to maintain backwards compat here in the case a subclass wants to keep the functionality. I guess could parameterize whether to skip dupe paths in this map method if you think it's worth keeping?

src/olmo_core/data/source_mixture.py
undfined: I just left them untouched, not _that_ noisy.

src/olmo_core/data/numpy_dataset.py
```diff
@@ -503,6 +506,156 @@ def _get_file_size_and_length(self, path, dtype=None) -> Tuple[int, int]:
             raise RuntimeError("invalid 'max_target_sequence_length' or 'sequence_length'")
 
 
+class NumpyFSLDatasetMixture(NumpyFSLDataset):
+    """
+    A version of :class:`NumpyFSLDataset` built from a mixture of sources and their expected token ratios relative to each other. A path_offset_index is used to determine the number of instances to retain from a path when constructing the local indices.
```
epwalsh: This would render poorly in the docs without the double backticks:
```suggestion
    A version of :class:`NumpyFSLDataset` built from a mixture of sources and their expected token ratios relative to each other. A ``path_offset_index`` is used to determine the number of instances to retain from a path when constructing the local indices.
```

epwalsh: BTW you can build and serve the docs locally by running `make docs`.

src/olmo_core/data/numpy_dataset.py
```diff
@@ -1519,6 +1652,8 @@ def build(self) -> NumpyDatasetBase:
                 paths.extend(matches)
         elif self.paths:
             paths = self.paths
+        elif self.source_mixture_config:
+            log.info("Building dataset from source mixture...")
```
epwalsh: `self.source_mixture_config` is only used when `self.type == NumpyDatasetType.fsl`, right? We should check for that, otherwise looks like we could get a silent failure where `paths` just ends up empty.

epwalsh (commented):
Looks really good! A couple more minor comments

src/olmo_core/data/numpy_dataset.py
undfined: Yep, good catch. On it.

src/olmo_core/data/utils.py
```diff
@@ -412,25 +412,32 @@ def segment_documents_into_instances(
...
         length = end_idx - start_idx
         indices.append(start_idx)
         indices.append(start_idx + min(length, max_sequence_length))
         start_idx += length
 
+    if sample is not None:
+        max_instances, seed = sample
+        rng = get_rng(seed)
+        indices = (
+            rng.choice(np.array(indices).reshape(-1, 2).tolist(), size=max_instances).flatten()
```
epwalsh: I think this would be pretty slow and memory-intensive relative to sampling with numpy directly. You shouldn't need to call `tolist()` I would think.

src/olmo_core/data/utils.py
```diff
@@ -411,23 +412,33 @@ def segment_documents_into_instances(
...
-        indices.append(start_idx + min(length, max_sequence_length))
-        start_idx += length
+    idx_gen = (
+        idx
+        for start_idx, end_idx in iter_document_indices(
+            path, eos_token_id=eos_token_id, dtype=dtype
+        )
+        for idx in (start_idx, start_idx + min(end_idx - start_idx, max_sequence_length))
+    )
+    indices = np.fromiter(idx_gen, dtype=indices_dtype)
```
epwalsh, line 433: Nice!

src/olmo_core/data/source_mixture.py
```diff
@@ -0,0 +1,303 @@
...
+    """
+    A configuration class for building a dataset from a fractionalized mixture of sources.
+    """
+
+    max_tokens: int
+    source_configs: List[SourceMixtureConfig]
+    sequence_length: int
+    dtype: NumpyDatasetDType
+    processes: int = 1
+    seed: int = 42
```
epwalsh: Could you add doc strings for these attributes? Even if most are obvious, our documentation build process won't show attributes that aren't documented.

src/olmo_core/data/source_mixture.py
```diff
@@ -0,0 +1,303 @@
...
+
+
+@dataclass
+class SourceMixtureConfig(Config):
+    source_name: str
+    target_ratio: float
+    paths: List[PathOrStr]
+    # 1.0 will result in a maximum of 1 repitition of the source data per epoch
+    max_repetition_ratio: float = 1.0
+    max_source_fraction: float = 1.0
```
epwalsh: Same comment about documenting these.

src/olmo_core/data/source_mixture.py
```diff
@@ -0,0 +1,303 @@
...
+
+@dataclass
+class SourceTokenDetails:
+    """
+    A class to hold intermediate selection details for a mixture source.
+    """
+
+    config: SourceMixtureConfig
+    population: int
+    num_selected: int
```
epwalsh: Same comment about documenting these.

CHANGELOG.md
```diff
@@ -9,6 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `SourceMixtureDataset` for composing a training mixture based on ratios of source datasets.
+- Added `NumpyFSLDatasetMixture` for constructing a `NumpyDatasetBase` from a `SourceMixtureDataset`. Note this is only supported for FSL datasets.
+- Added tests for `SourceMixture*` and `NumpyFSLDatasetMixture`.
+
+### Changed
+- Moved some types into `olmo_core.data.types` to avoid some circular dependencies.
 - Added `DownstreamEvaluatorCallbackConfig` class for running in-loop downstream eval via [OLMo-in-loop-evals](https://github.com/allenai/OLMo-in-loop-evals).
```
epwalsh: can you move this up to the "Added" section? I guess it got messed up in the merge.

src/olmo_core/data/numpy_dataset.py
```diff
@@ -1364,6 +1485,14 @@ class NumpyDatasetConfig(Config):
     """
     The type of dataset.
     """
+    bust_index_cache: bool = False
+    """
+    Whether or not to bust the index cache.
+    """
```
epwalsh: It looks like this is only used in the `NumpyFSLDatasetMixture`? Is it still needed or was it just for debugging? Overwriting the cache files could create a race condition if they're in-use by another training process, so I'd prefer to remove this option.

src/olmo_core/data/numpy_dataset.py
```diff
@@ -235,7 +242,7 @@ def _warmup_clients(self):
 
     def map(
         self,
-        func: Callable[[PathOrStr], T],
+        func: Callable[[Tuple[PathOrStr, int]], T],
```
epwalsh: Can you update the docstring of this method to reflect the new signature? Also, why not `func: Callable[[PathOrStr, int], T]` instead of `Callable[[Tuple[PathOrStr, int], T]`?

epwalsh (commented):
A few more comments

CHANGELOG.md
undfined: Yeah funky merge, on it

src/olmo_core/data/numpy_dataset.py
undfined: Yeah this was really just for debugging, the race condition is reason enough to drop it.

src/olmo_core/data/numpy_dataset.py
undfined: No reason, personal preference :) I switched it back to a list of positional args.

src/olmo_core/data/utils.py
undfined: 🫡

epwalsh (approved):
LGTM!

PR #74, opened 2024-10-30
Make GCS client more robust

epwalsh:


src/olmo_core/io.py
```diff
@@ -409,14 +410,28 @@ def _get_gcs_client():
     return gcs.Client()
 
 
+def _gcs_is_retriable(exc: Exception) -> bool:
+    from google.api_core.retry import if_transient_error
+
+    return if_transient_error(exc) or isinstance(exc, requests.exceptions.Timeout)
+
+
+def _get_gcs_retry():
```
dirkgr, line 419: Why not just create a retry object once? Why wrap it in a function and create a new one every time?

dirkgr (changes requested):

src/olmo_core/io.py
epwalsh: cuz GCS is an optional dependency

PR #98, opened 2024-11-11
replace printing table with logging table

epwalsh:
I noticed this table gets printed during unit tests, muddling the pytest output. If we log the table instead of print it, it should still be rendered in the same way (due to our custom rich log handler), but will respect other logging settings.


undfined (approved):
lgtm!

PR #114, opened 2024-11-22
Add model ladder building blocks

epwalsh:
Porting over from https://github.com/allenai/OLMo/blob/ladder-1xC/scripts/ladder_peteish.py.

### Example

You can run a model size (e.g. the 190M) for the peteish ladder on Beaker as follows:

```fish
python src/scripts/train/OLMo2-ladder.py launch 190M ai2/jupiter-cirrascale-2
```

### Notes for reviewers

The key file to review is `src/olmo_core/model_ladder.py`.


epwalsh:
Merging now because I need some of these changes for other things, but plz leave comments anyway.

PR #119, opened 2024-12-05
Add learning rate schedulers

2015aroras:
This PR adds a few standard LR schedulers. Unlike in the original OLMo, these do not have warmup inbuilt.

- `LinearScheduler` - Standard linear LR schedule. LR is kept constant once the end of the decay is reached.
- `InvSqrtScheduler` - Inverse square root LR schedule.

I have also added some schedulers that I think are convenient.

- `LinearWarmupDecoratorScheduler` - Wraps an existing scheduler, adding linear warmup to it.
- `SequentialScheduler` - Allows you to chain schedules sequentially, so that one is called after another. The initial LR of a scheduler is that final LR of the previous scheduler. Last scheduler is assumed to run indefinitely.

As an example, the following scheduler warms up for 100 steps, then does inverse square root to 1/2 max LR for 150 steps, then linear decay to 0 for 200 steps, and then constant forever after (since linear schedule is constant after decay is done).
```
SequentialScheduler(
    schedulers=[
        LinearWarmupDecoratorScheduler(
            warmup_steps=100,
            inner=InvSqrtScheduler(
                alpha_f=0.5,
                step_offset=100,
            ),
        ),
        LinearScheduler(alpha_f=0, t_max=200),
    ],
    schedulers_max_steps=[250],
)
```

### DIFF
--- a/src/olmo_core/optim/__init__.py
+++ b/src/olmo_core/optim/__init__.py
@@ -2,7 +2,15 @@ from .adam import AdamConfig
 from .adamw import AdamWConfig
 from .config import OptimConfig, OptimGroupOverride
 from .lion import Lion, LionConfig, SkipStepLion, SkipStepLionConfig
-from .scheduler import ConstantScheduler, CosWithWarmup, Scheduler
+from .scheduler import (
+    ConstantScheduler,
+    CosWithWarmup,
+    InvSqrtScheduler,
+    LinearWarmupDecoratorScheduler,
+    LinearScheduler,
+    Scheduler,
+    SequentialScheduler,
+)
 from .skip_step_optimizer import SkipStepOptimizer
 
 __all__ = [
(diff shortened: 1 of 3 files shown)


2015aroras:
The example is currently running on 1 GPU on W&B: https://wandb.ai/ai2-llm/OLMo-core-1B/runs/zlta7l2a

epwalsh (commented):
The warmup wrapper is nice from a code perspective but less so from a configuration perspective since you would have to import two different scheduler classes now just to specify a single schedule. It's also inconsistent with the existing `CosWithWarmup` schedule. So I recommend we just make these consistent with `CosWithWarmup`.

2015aroras:
> The warmup wrapper is nice from a code perspective but less so from a configuration perspective since you would have to import two different scheduler classes now just to specify a single schedule. It's also inconsistent with the existing `CosWithWarmup` schedule. So I recommend we just make these consistent with `CosWithWarmup`.

https://github.com/allenai/OLMo-core/pull/119/commits/605994793b2b248e438a0853b9d7f3e2cc1dd981 Done, though it's probably easier to review the PR from scratch than to look at this commit.

epwalsh (approved):
LGTM

PR #121, opened 2024-12-10
32 b

dirkgr:


src/olmo_core/train/callbacks/evaluator_callback.py
```diff
@@ -130,7 +130,7 @@ def build(self, trainer: "Trainer") -> Optional[Callback]:
         eval_batch_size = (
             self.eval_batch_size
             if self.eval_batch_size is not None
-            else trainer.rank_microbatch_size * get_world_size(trainer.dp_process_group)
+            else 2 * trainer.rank_microbatch_size * get_world_size(trainer.dp_process_group)
```
2015aroras: nit: you could instead passed an updated evaluator callback in `OLMo2-32B.py`:
```
.with_callback(
    "lm_evaluator",
    LMEvaluatorCallbackConfig(
        eval_batch_size=<whatever you want>,
        eval_dataset=NumpyDatasetConfig.from_data_mix(
            DataMix.v3_small_ppl_validation,
            name=NumpyDatasetType.padded_fsl,
            mix_base_dir=root_dir,
            sequence_length=dataset_config.effective_sequence_length,
            tokenizer=tokenizer_config,
            work_dir=get_work_dir(root_dir),
        ),
        eval_interval=1000,
    ),
```

dirkgr:
Yeah, but I think this is better. I think we can default to 2x the training
batch size. It should always work.

src/scripts/train/OLMo2-32B.py
```diff
@@ -52,20 +53,23 @@ def build_optim_config(common: CommonComponents) -> AdamWConfig:
 
 
 def build_trainer_config(common: CommonComponents) -> TrainerConfig:
+    project_name = "peteish32"
     return (
         TrainerConfig(
-            save_folder=common.save_folder,
+            save_folder=f"gs://ai2-llm/checkpoints/{project_name}/",
```
epwalsh, line 74: Why change this?

src/olmo_core/nn/functional/cross_entropy_loss.py
```diff
@@ -53,11 +53,11 @@ def cross_entropy_loss(
...
 
 try:
-    import olmo_core.triton.cross_entropy_loss as triton_ce_loss
+    # import olmo_core.triton.cross_entropy_loss as triton_ce_loss
+    #_fused_cross_entropy_loss = triton_ce_loss.cross_entropy_loss
 
-    #  import flash_attn.ops.triton.cross_entropy as flash_attn_ce  # type: ignore
-
-    _fused_cross_entropy_loss = triton_ce_loss.cross_entropy_loss
+    import flash_attn.ops.triton.cross_entropy as flash_attn_ce  # type: ignore
```
epwalsh: Our in-house triton CE loss was copied directly from the flash-attn repo, so I don't see the point of this.

src/olmo_core/nn/transformer/config.py
```diff
@@ -460,19 +460,22 @@ def olmo2_13B(cls, vocab_size: int, **kwargs) -> "TransformerConfig":
         )
 
     @classmethod
-    def olmo2_26B(cls, vocab_size: int, **kwargs) -> "TransformerConfig":
+    def olmo2_32B(cls, vocab_size: int, **kwargs) -> "TransformerConfig":
         """
-        A 26B OLMo model config.
+        A 32B OLMo model config.
         """
+        d_model = 5120
```
epwalsh: this is a very narrow model then... are you sure about that?

src/scripts/train/OLMo2-32B.py
```diff
@@ -52,20 +53,23 @@ def build_optim_config(common: CommonComponents) -> AdamWConfig:
...
             rank_microbatch_size=4 * 4096,
             save_overwrite=True,
             metrics_collect_interval=10,
-            cancel_check_interval=1,
+            cancel_check_interval=10,
             z_loss_multiplier=1e-5,
-            compile_loss=True,
+            fused_loss=True,
+            compile_loss=False,
```
epwalsh: I understand the trepidation about the different loss implementations, but the way it was before was the most performant. This way will be slower and have a higher memory footprint.

src/scripts/train/OLMo2-32B.py
```diff
@@ -85,10 +89,57 @@ def build_trainer_config(common: CommonComponents) -> TrainerConfig:
             WandBCallback(
                 name=common.run_name,
                 entity="ai2-llm",
-                project="OLMo-core-26B",
+                project=project_name,
                 enabled=False,
                 cancel_check_interval=10,
             ),
+        ).with_callback(
```
epwalsh: We should just add this to the common callbacks.
https://github.com/allenai/OLMo-core/blob/5de774f3046b21e3e4619426f4de5d460367a3e4/src/olmo_core/internal/experiment.py#L163

src/scripts/train/OLMo2-32B.py
dirkgr: It defaults to something under my name? Not what we want for an official run?

dirkgr: Especially if we swap babysitting responsibilities during the run

src/olmo_core/nn/transformer/config.py
dirkgr: It's a clone of Qwen 32. The tradeoffs are, narrow d_model, wide FFN, GQA, lots of layers.

src/scripts/train/OLMo2-32B.py
dirkgr: Can we have some certainty that this will do the right thing? What happens if we take the 13B from a late checkpoint and run it?

src/olmo_core/nn/functional/cross_entropy_loss.py
dirkgr: Ok, I took this back out.

src/scripts/train/OLMo2-32B.py
dirkgr: I don't know that we want these for everything. Default should probably be only the new, blessed ones.

src/olmo_core/train/callbacks/evaluator_callback.py
dirkgr: Yeah, but I think this is better. I think we can default to 2x the training batch size. It should always work.

src/olmo_core/nn/functional/cross_entropy_loss.py
dirkgr: Do I want compiling and fused loss at the same time?

src/scripts/train/OLMo2-32B.py
```diff
@@ -85,10 +94,57 @@ def build_trainer_config(common: CommonComponents) -> TrainerConfig:
             WandBCallback(
                 name=common.run_name,
                 entity="ai2-llm",
-                project="OLMo-core-26B",
+                project=project_name,
                 enabled=False,
```
2015aroras, line 113: Intentionally disabled still? Just checking

src/olmo_core/io.py
```diff
@@ -590,7 +594,7 @@ def _gcs_get_bytes_range(bucket_name: str, key: str, bytes_start: int, num_bytes
     )
 
 
-@retriable()
+@retriable(retry_condition=_gcs_is_retriable)
```
2015aroras: This general approach sort of blows up our retry time from 10 mins to 30 mins. Sort of not a fan.

2015aroras: But at least it looks like it works.

epwalsh: We could always reduce the deadline/timeout

dirkgr:
Was merged piecemeal.

PR #123, opened 2024-12-16
Add option to pre-download checkpoint to load

epwalsh:
This should help with the situation @dirkgr, @2015aroras, and I discussed. You can activate this with:

```
--trainer.checkpointer.pre_download=true
```

### DIFF
--- a/src/olmo_core/train/checkpoint.py
+++ b/src/olmo_core/train/checkpoint.py
@@ -178,6 +178,7 @@ class Checkpointer:
             process_group=self.process_group,
             key_mapping=key_mapping,
             pre_download=is_url(dir) and self.pre_download,
+            work_dir=self.work_dir,
         )
 
         return trainer_state
--- a/src/olmo_core/train/checkpoint.py
+++ b/src/olmo_core/train/checkpoint.py
@@ -162,7 +162,7 @@ class Checkpointer:
             log.info(f"Pre-downloading checkpoint from '{dir}' to '{target}'...")
             if get_fs_local_rank() == 0:
                 copy_dir(dir, target, save_overwrite=self.save_overwrite)
-            barrier()
+            barrier(self.process_group)
             return self.load(
                 target,
                 model,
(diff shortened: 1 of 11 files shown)


src/olmo_core/train/checkpoint.py
```diff
@@ -128,6 +154,20 @@ def load(
         """
         dir = normalize_path(dir)
 
+        if is_url(dir) and self.pre_download:
+            target = self.work_dir / "load" / os.path.basename(dir)
+            if get_fs_local_rank() == 0:
+                copy_dir(dir, target, save_overwrite=self.save_overwrite)
+            barrier()
```
dirkgr: This is means that if the copy takes longer than 10 minutes, we will definitely timeout. Is there a way to extend the timeout, or have the other ranks participate in the copying somehow?

epwalsh: In your PR for the 32B config you extended the timeout to 30 minutes. Is that enough?

dirkgr: Yeah, but 30 minutes is really bad when you don't need it. Might make a 1B model stall out for 30 minutes when it doesn't need to.

epwalsh: There's a way to decrease the timeout later. Let me see if I can get that working.

epwalsh: actually we can just increase the timeout for the checkpointing PG: d67f767

2015aroras:
While this works, it seems like a short-term solution. It makes a model copy per node. This is good when doing HSDP with a replica per node (which is our current need), but bad for FSDP. Ideally we would only download the files needed by the rank or node.

epwalsh:
> While this works, it seems like a short-term solution. It makes a model copy per node. This is good when doing HSDP with a replica per node (which is our current need), but bad for FSDP.

If I understood correctly, the original issue this is trying solve is loading a checkpoint that was saved with different sharding configuration (e.g. saved with FSDP, load with HSDP) because that leads to too many small requests to GCS. If that's the case then there's no point in having this toggled on when you're loading a checkpoint saved with the same sharding configuration.

> Ideally we would only download the files needed by the rank or node.

This is basically what happens without this option toggled on. Except due to the way `torch.distributed.checkpoint` works, we're downloaded slices of files, not whole files. In any case, each rank only downloads the data it needs to fill its local shard of parameters.

2015aroras (approved):

PR #124, opened 2024-12-18
Pre-download only relevant files per rank.

2015aroras:
This builds on #123 and tries to change it in ways listed below. Completely untested and just an idea.

- Instead of pre-downloading the whole checkpoint, download a checkpoint file when a range query is about to made on it for the first time. `resource_path` will (hopefully) prevent any concurrency issues. As a bonus, this also splits up the work between the various processes.
- Instead of saving each non-sharded data item on the rank with the smallest save payload (thus spreading the items out across a bunch of files/ranks), save them all to the lowest rank (empirically, rank 0). This will limit how many files we have to download/read from, but makes rank 0 bigger (I think only by like 1 MB though).
  - This is the solution for https://github.com/pytorch/pytorch/issues/125740

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,7 @@ license = { file = "LICENSE" }
 dependencies = [
     "numpy<2.0",
     "torch>=2.5.1",
-    "cached-path",
+    "cached-path@git+https://github.com/allenai/cached_path@shanea/cache-gs-client",
     "requests",
     "packaging",
     "rich",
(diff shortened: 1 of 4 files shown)


epwalsh (approved):
This is good, thanks for implemented it

epwalsh:
@2015aroras ready to merge into the other branch?

2015aroras:
[c99da44](https://github.com/allenai/OLMo-core/pull/124/commits/c99da4478e3c8b0b7ba73c23b514e8845d119663) I need to cache the GCS client in `cached_path`, otherwise we get an error we saw in the old trainer until we did a similar fix.

PR #125, opened 2024-12-19
Add a callback for sending Slack notifications

epwalsh:
Requires the env var `SLACK_WEBHOOK_URL`.  I've already added this env var to our `ai2/32B` Beaker workspace which will post to the `#oe-training-on-call` channel.


2015aroras (approved):
Having a channel option in some form would be nice, whether that is passing in a different webhook URL or passing it in as a field

PR #127, opened 2024-12-22
Skip step AdamW

2015aroras:
This PR implements `SkipStepAdamW`. It's not ideal since it doesn't support fused ops or similar at present. I changed default AdamW to also use the same underlying implementation

### DIFF
--- a/src/olmo_core/optim/adamw.py
+++ b/src/olmo_core/optim/adamw.py
@@ -4,11 +4,12 @@ from typing import Optional, Tuple, Type
 
 import torch
 import torch.nn as nn
+from torch.optim.optimizer import Optimizer
 
 from .config import OptimConfig
+from .skip_step_optimizer import SkipStepOptimizer
 
 
-# TODO: use this when we implement a "skip step" version of AdamW.
 def adamw_step(
     p: nn.Parameter,
     *,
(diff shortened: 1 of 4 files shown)


src/olmo_core/optim/adamw.py
```diff
@@ -47,6 +48,129 @@ def adamw_step(
...
+
+    def __init__(
+        self,
+        params,
+        lr: float = 1e-3,
+        betas: Tuple[float, float] = (0.9, 0.999),
+        eps: float = 1e-8,
+        weight_decay: float = 1e-2,
+        foreach: Optional[bool] = None,
+        fused: Optional[bool] = None,
```
dirkgr, line 64: These two are ignored?

src/olmo_core/optim/adamw.py
```diff
@@ -47,6 +48,129 @@ def adamw_step(
...
+        rolling_interval_length: int = 128,
+        sigma_factor: int = 6,
```
dirkgr, line 119: Can you check whether this would have caught our spikes?

src/olmo_core/optim/adamw.py
2015aroras: Yes

src/olmo_core/optim/adamw.py
```diff
@@ -61,5 +185,25 @@ class AdamWConfig(OptimConfig):  # NOTE: omagaconf doesn't like "OptimConfig[tor
...
+    """
+    Configuration class for building a :class:`SkipStepAdamW` optimizer.
+    """
+
+    lr: float = 1e-3
+    betas: Tuple[float, float] = (0.9, 0.999)
+    eps: float = 1e-8
+    weight_decay: float = 1e-2
+    foreach: Optional[bool] = None
+    fused: Optional[bool] = None
```
epwalsh, line 203: These aren't used, so let's remove them.

src/olmo_core/optim/adamw.py
```diff
@@ -61,5 +185,25 @@ class AdamWConfig(OptimConfig):  # NOTE: omagaconf doesn't like "OptimConfig[tor
     fused: Optional[bool] = None
 
     @classmethod
-    def optimizer(cls) -> Type[torch.optim.AdamW]:
-        return torch.optim.AdamW
```
epwalsh, line 65: Why change this? We lose foreach and fused options unnecessarily.

epwalsh (changes requested):
> It's not ideal since it doesn't support fused ops or similar at present. I changed default AdamW to also use the same underlying implementation.

I don't see a good reason for this. Let's default to the built-in AdamW so we can still use the `foreach` and `fused` options.

2015aroras:
At this point I've made so many other relevant changes to make this work that it might be easier to just merge the 32B when we're back

PR #128, opened 2024-12-23
Fix GCS upload retries

2015aroras:
Issue: Our GCS uploads do not retry on transient errors. This is because the conditional retry logic expects a blob "generation" to be passed in. This generation lets the operation be sure that what it is doing is idempotent.

Fix: Pass in the generation to GCS upload


src/olmo_core/io.py
```diff
@@ -2,6 +2,7 @@
 import logging
 import os
 import pickle
+import random
```
epwalsh, line 5: unused import?
```suggestion
```

epwalsh (approved):
LGTM other than the random random import

2015aroras:
This is already in main. Probably added as part of merging 32B into main

PR #131, opened 2025-01-02
Make training on the MPS device work

dirkgr:


dirkgr:
@epwalsh , some thoughts? This now has a `train_single` command that does almost everything for you. Most of the time you should be able to take whatever command you have failing on the big cluster, replace `launch` with `train_single`, and repro the error on your Mac. You'll have to do some stuff to make it fit into memory, but that can't be automated.

However, another view could be that "your local machine" is a "cluster" like any other. Then we don't need another command. But it's a little weird that the special automatic munging of configs we do only happens when your cluster is set to "local". Also hard to write the code for that.

epwalsh:
I think having the separate command is fine

src/scripts/train/OLMo2-1B.py
```diff
@@ -76,48 +73,6 @@ def build_trainer_config(common: CommonComponents) -> TrainerConfig:
                 cancel_check_interval=10,
             ),
         )
-        .with_callback(
```
epwalsh, line 79: Is this change intentional?

epwalsh (commented):
LGTM, other than the changes to the 1B config

src/scripts/train/OLMo2-1B.py
dirkgr: Yes. That callback doesn't initialize properly, because some of those datasets don't exist, and none of the other sizes have it.

epwalsh (approved):

PR #132, opened 2025-01-06
Change the way `load_path` is handled

epwalsh:


src/olmo_core/train/common.py
```diff
@@ -75,17 +75,19 @@ class LoadStrategy(StrEnum):
 
     if_available = "if_available"
     """
-    Only load from the load path if a checkpoint exists there.
+    The trainer will attempt to load a checkpoint from the save folder or load path (in that order)
```
2015aroras, line 78: Thanks for updating these! I started reviewing before you updated them and got confused.

src/olmo_core/train/trainer.py
```diff
@@ -538,20 +540,29 @@ def check_if_canceled(self):
...
 
         # Maybe load a checkpoint.
-        if not self.checkpoint_loaded:
-            load_path = self.load_path if self.load_path is not None else self.save_folder
-            if self.load_strategy == LoadStrategy.always:
-                self.load_checkpoint(load_path)
-            elif self.load_strategy == LoadStrategy.if_available:
-                self.maybe_load_checkpoint(load_path)
+        if not self.checkpoint_loaded and self.load_strategy != LoadStrategy.never:
+            if not self.maybe_load_checkpoint(self.save_folder) and self.load_path is not None:
```
2015aroras: nit: I'm not a fan of `self.maybe_load_checkpoint(self.save_folder)` being in the if statement due to its side-effect nature of setting `self._checkpoint_loaded`.

Moreover, I'm not a fan of `self.maybe_load_checkpoint` returning a bool that isn't `self._checkpoint_loaded`. Imo we should either
1. remove `self._checkpoint_loaded`
2. (my favorite) make `self.maybe_load_checkpoint` not return anything and rely explicitly on `self.checkpoint_loaded`
3. make `self.maybe_load_checkpoint` return `self.checkpoint_loaded`

2015aroras (approved):
Thank you!

src/olmo_core/train/trainer.py
epwalsh: Fair point about the if statement, changed in d01aafc.

But I do think it's a useful part of the API for `maybe_load_checkpoint()` to return a boolean, and this shouldn't necessarily be the value of `self._checkpoint_loaded`, since `self._checkpoint_loaded` tells you whether any checkpoint has been loaded at any point in time, whereas the return value of `maybe_load_checkpoint` tells you if a checkpoint was loaded this time from the given load path.

dirkgr (approved):
I like the change, but didn't read the code closely, since @2015aroras is already on it.

PR #134, opened 2025-01-07
32B with Hybrid Sharding

dirkgr:


dirkgr:
We no longer need this.

PR #138, opened 2025-01-11
Allow the dataset configuration to be overridden in experiment.

soldni:
matches behavior of other configurations.


AkshitaB:
I require this for my workflow too.

PR #142, opened 2025-01-21
Add option to throttle checkpoint uploads to one rank from each node at a time

epwalsh:


src/olmo_core/distributed/utils.py
```diff
@@ -421,3 +422,36 @@ def get_local_tensor(x: torch.Tensor) -> torch.Tensor:
         return x.to_local()
     else:
         return x
+
+
+def do_n_at_a_time(
```
2015aroras, line 427: nit: one way to avoid having to use `partial` when invoking this is to pass `*args` and `**kwargs` as parameters. We do lose type-checking though.

src/test/distributed/utils_test.py
```diff
@@ -18,3 +20,20 @@ def scatter_object():
...
+    def func(rank: int):
+        nonlocal times_called
+        times_called += 1
+        calling_ranks.add(rank)
+
+    for rank in range(world_size):
+        dist_utils.do_n_at_a_time(partial(func, rank), n=n, world_size=world_size, local_rank=rank)
+
+    assert times_called == world_size
+    assert calling_ranks == set(list(range(world_size)))
```
2015aroras: nit: `list` is not needed

2015aroras (approved):

src/olmo_core/distributed/checkpoint/filesystem.py
```diff
@@ -201,22 +208,36 @@ def gen_file_name() -> str:
...
+                    futures.append(executor.submit(write_items, [bucket]))
+
+                for f in as_completed(futures):
+                    try:
+                        results.extend(f.result())
+                    except BaseException:
+                        # NOTE: we might get an error here that can't be pickled, which causes a different failure
+                        # later when PyTorch tries to reduce that error across ranks. So here we just make
+                        # sure we're raising a simple error type that can be pickled.
+                        raise OLMoCheckpointError(f"Original error:\n{traceback.format_exc()}")
```
dirkgr: This exception wrapping doesn't have to be pulled into `write_items()`?

epwalsh: good catch. 5aeca1a

src/test/distributed/utils_test.py
epwalsh: fixed: bb56008

src/olmo_core/distributed/utils.py
epwalsh: also those kwargs could conflict with kwargs to this function.

2015aroras (approved):

dirkgr (approved):
Beautiful

PR #143, opened 2025-01-22
save max steps to trainer state

epwalsh:
Convenient to have this information in the checkpoint somewhere.

### DIFF
--- a/src/olmo_core/train/trainer.py
+++ b/src/olmo_core/train/trainer.py
@@ -75,6 +75,7 @@ T = TypeVar("T")
 class TrainerStateDict(TypedDict):
     global_step: int
     global_train_tokens_seen: int
+    max_steps: int
     data_loader: Dict[str, Any]
     epoch: int
     world_size: int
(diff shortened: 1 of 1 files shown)


src/olmo_core/train/trainer.py
```diff
@@ -619,6 +620,7 @@ def state_dict(self) -> TrainerStateDict:
         return {
             "global_step": self.global_step,
             "global_train_tokens_seen": self.global_train_tokens_seen,
+            "max_steps": self.max_steps,
```
2015aroras, line 623: What happens to `def max_steps(self) -> int:` if this is here?

2015aroras: My brain is a bit mush clearly, but I'm just thinking about how this field is completely ignored when loading training state.

epwalsh: it is ignored, and that's fine

epwalsh:
@2015aroras are you okay with this?

src/olmo_core/train/trainer.py
2015aroras: Maybe add a comment that this is ignored for now?

2015aroras (approved):

PR #144, opened 2025-01-23
WIP: Guard against indices with 0 samples

undfined:


src/olmo_core/data/utils.py
```diff
@@ -438,10 +438,14 @@ def segment_documents_into_instances(
...
 
-    with memmap_to_write(target, dtype=indices_dtype, shape=(indices.size,)) as indices_mmap:
-        indices_mmap[:] = indices
+    # NOTE: It's possible to sample 0 instances from small source files. Rather than try to write this empty array we conditionally skip if indices_out is 0.
+    indices_out = len(indices) // 2
 
-    return total_og_docs, len(indices) // 2
+    if indices_out > 0:
+        with memmap_to_write(target, dtype=indices_dtype, shape=(indices.size,)) as indices_mmap:
+            indices_mmap[:] = indices
```
epwalsh: Is it going to cause issues elsewhere when we never write to the target memmap file?

PR #145, opened 2025-01-23
Add more options to the `unshard_checkpoint` function to help scale

epwalsh:
I was getting worried about unsharding really big checkpoints like for the 32B, which we'll need to do soon. The main issue at the moment is that in order to unshard we need to load the entire model (or optimizer state) in memory, which clearly isn't scalable. So I've added an option to unshard the checkpoint into chunks of a given size which helps scale because only a single chunk (which could be as small as a single tensor) needs to load into memory at a time. Each chunk gets written to a unique file. I think HuggingFace does something similar.

Note: this is not supported for optimizer state yet. But, speaking of optimizer state, this PR also adds a function called `load_keys()` for loading (and unsharding) specific keys from a checkpoint. So if you want to inspect part of the optimizer state, you could use that function without having to unshard the whole optimizer state.

### DIFF
--- a/src/olmo_core/distributed/checkpoint/__init__.py
+++ b/src/olmo_core/distributed/checkpoint/__init__.py
@@ -29,7 +29,7 @@ import logging
 from concurrent.futures import Future
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, Generator, Iterable, List, Optional, Tuple
 
 import torch
 import torch.distributed as dist
@@ -53,6 +53,7 @@ __all__ = [
     "async_save_model_and_optim_state",
     "load_model_and_optim_state",
     "unshard_checkpoint",
+    "load_keys",
     "get_checkpoint_metadata",
     "UnshardStrategy",
     "UnshardStrategyType",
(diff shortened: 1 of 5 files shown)


src/olmo_core/distributed/checkpoint/__init__.py
```diff
@@ -329,82 +403,154 @@ def unshard_checkpoint(
...
+
+        if unshard_strategy.name == UnshardStrategyType.one_file:
+            path = target_dir / f"{prefix}.{ext}"
+            return path, [(path, [prefix])]
+        elif unshard_strategy.name == UnshardStrategyType.one_file_per_tensor:
+            path = target_dir / prefix
+            chunks = []
+            for chunk_num, key in enumerate(metadata.state_dict_metadata.keys()):
+                if key.startswith(f"{prefix}."):
+                    chunks.append((path / f"chunk-{chunk_num:05d}.{ext}", [key]))
```
2015aroras: thought: it could be nice to name it as `chunk-{key}`, maybe with some cleanup of the key. Then one could easily examine the weights of specific parameters. Would possibly be slightly cleaner for hypothetical future converter logic too.

2015aroras (approved):

src/olmo_core/distributed/checkpoint/__init__.py
epwalsh: done: bc79ab5

PR #150, opened 2025-01-29
Fix device in mock batch

AkshitaB:
Fixed bug in `get_mock_batch()` by adding `non_cuda` option so that mock batches can be loaded on cuda devices in addition to MPS devices.


epwalsh:
Fixed in 7ec9114. These batches should just be created on CPU anywa.

PR #163, opened 2025-02-11
reset initial LR to configured value after loading

epwalsh:


src/olmo_core/optim/config.py
```diff
@@ -57,6 +69,12 @@ class OptimConfig(Config, Generic[Opt], metaclass=ABCMeta):
         due to the LR being restored to a float instead of a tensor.
     """
 
+    fixed_fields: Tuple[str, ...] = ("initial_lr",)
```
dirkgr, line 72: This is going to be tricky. What we really need, in full generality, is merging configs where parts of the config are allowed to be unset and then they go to whatever default is there (like, values from a previous run). But that sounds like it would be a big change.

dirkgr (approved):

PR #164, opened 2025-02-14
Learn2code | Review of the love2code configs

revbucket:
Dirk, help?

### DIFF
--- a/src/olmo_core/nn/transformer/config.py
+++ b/src/olmo_core/nn/transformer/config.py
@@ -821,3 +821,18 @@ class TransformerConfig(Config):
             init_method=InitMethod.normalized,
             **kwargs,
         )
+
+
+
+    @classmethod
+    def love2code_3B(cls, vocab_size: int, **kwargs) -> "TransformerConfig":
+        """ OLMo-core repro of GraniteCode 3B model """
+        return cls.llama_like(
+            d_model=2560,
+            hidden_size_multiplier=1.5,
+            vocab_size=vocab_size,
+            n_layers=32,
+            n_heads=32,
+            **kwargs
+        )
+        # num_params: 3607267840
--- a/src/scripts/train/learn2code/love2code-3B.py
+++ b/src/scripts/train/learn2code/love2code-3B.py
@@ -4,10 +4,11 @@ Train a 3B for a first coding run. Run this script without any arguments to see 
 
 from olmo_core.config import DType
 from olmo_core.distributed.parallel import DataParallelType
-from olmo_core.internal.experiment import CommonComponents, main, build_common_components
-from olmo_core.internal.common import Duration, get_work_dir, get_root_dir
+from olmo_core.internal.experiment import CommonComponents, main, build_common_components, SubCmd, ExperimentConfig
+from olmo_core.internal.common import get_work_dir, get_root_dir
+from olmo_core.train.common import Duration
 from olmo_core.nn.transformer import TransformerConfig, TransformerDataParallelConfig
-from olmo_core.optim import AdamWConfig, OptimGroupOverride
+from olmo_core.optim import AdamWConfig, OptimGroupOverride, OptimConfig
 from olmo_core.data import (
     DataMixBase,
     NumpyDataLoaderConfig,
@@ -17,8 +18,8 @@ from olmo_core.data import (
 )
 from olmo_core.train import TrainerConfig
 from olmo_core.train.callbacks import CheckpointerCallback, CometCallback, WandBCallback
-
-
+from typing import Callable, Dict, List, Tuple, Optional, cast
+import sys
 
 SEQUENCE_LENGTH = 2048 
 CHINCHILLA_5X_DURATION = Duration.tokens(36072678400 * 20 * 5)
@@ -57,7 +58,7 @@ class Love2CodeDataMix(DataMixBase):
 
 def build_love2code_common(
     script: str,
-    cmd: Subcmd,
+    cmd: SubCmd,
     run_name: str,
     overrides: List[str],
     *,
@@ -68,7 +69,7 @@ def build_love2code_common(
     I took the original internal.experiment.build_common_components and ran it
     and then built the dataset config and hotswapped it. Maybe not canonical, but seems easiest?
     """
-    og_common = build_common_components(script, cmd, run_name, cluster, overrides, *, global_batch_size=global_batch_size)
+    og_common = build_common_components(script, cmd, run_name, cluster, overrides, global_batch_size=global_batch_size)
     tokenizer_config = og_common.tokenizer
     root_dir = get_root_dir(cluster)
 
--- a/src/scripts/train/learn2code/love2code-3B.py
+++ b/src/scripts/train/learn2code/love2code-3B.py
@@ -1,0 +1,300 @@
+"""
+Train a 3B for a first coding run. Run this script without any arguments to see usage info.
+"""
+
+from olmo_core.config import DType
+from olmo_core.distributed.parallel import DataParallelType
+from olmo_core.internal.experiment import CommonComponents, main, build_common_components
+from olmo_core.internal.common import Duration, get_work_dir, get_root_dir
+from olmo_core.nn.transformer import TransformerConfig, TransformerDataParallelConfig
+from olmo_core.optim import AdamWConfig, OptimGroupOverride
+from olmo_core.data import (
+    DataMixBase,
+    NumpyDataLoaderConfig,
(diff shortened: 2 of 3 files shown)


src/olmo_core/nn/transformer/config.py
```diff
@@ -821,3 +821,18 @@ def ngpt_like(
...
         )
+
+
+
+    @classmethod
+    def love2code_3B(cls, vocab_size: int, **kwargs) -> "TransformerConfig":
+        """ OLMo-core repro of GraniteCode 3B model """
+        return cls.llama_like(
+            d_model=2560,
+            hidden_size_multiplier=1.5,
```
dirkgr, line 962: This isn't right. Based on the Granite paper, this should be `10240 // 2560`, but it also has to be a multiple of 128.

src/olmo_core/nn/transformer/config.py
```diff
@@ -821,3 +821,18 @@ def ngpt_like(
...
+            vocab_size=vocab_size,
+            n_layers=32,
+            n_heads=32,
+            **kwargs
+        )
+        # num_params: 3607267840
```
dirkgr: Where did you get this number?

src/scripts/train/learn2code/love2code-3B.py
```diff
@@ -0,0 +1,301 @@
...
+    NumpyDatasetConfig,
+    TokenizerConfig,
+    TokenizerName,
+)
+from olmo_core.train import TrainerConfig
+from olmo_core.train.callbacks import CheckpointerCallback, CometCallback, WandBCallback
+from typing import Callable, Dict, List, Tuple, Optional, cast
+import sys
+
+SEQUENCE_LENGTH = 2048
```
dirkgr: Short is OK for ablations? Also good for perf?

src/scripts/train/learn2code/love2code-3B.py
```diff
@@ -0,0 +1,301 @@
...
+    TokenizerConfig,
+    TokenizerName,
+)
+from olmo_core.train import TrainerConfig
+from olmo_core.train.callbacks import CheckpointerCallback, CometCallback, WandBCallback
+from typing import Callable, Dict, List, Tuple, Optional, cast
+import sys
+
+SEQUENCE_LENGTH = 2048 
+CHINCHILLA_5X_DURATION = Duration.tokens(36072678400 * 20 * 5)
```
dirkgr: Is that the real number of params? Should be non-embedding params. There might be a way to get this programmatically.

src/scripts/train/learn2code/love2code-3B.py
```diff
@@ -0,0 +1,301 @@
...
+
+    dataset = NumpyDatasetConfig.from_data_mix(
+        Love2CodeDataMix.love2code_mix,
+        tokenizer=tokenizer,
+        mix_base_dir=root_dir,
+        sequence_length=SEQUENCE_LENGTH,
+        work_dir=get_work_dir(root_dir)
+        )
+    data_loader_config = NumpyDataLoaderConfig(
+        global_batch_size=global_batch_size, seed=34521, num_workers=4)
```
dirkgr: Guessing you need more data loading workers. A small model will run fast, so the data needs to come in faster. I'd start with 8 or 16.

src/scripts/train/learn2code/love2code-3B.py
```diff
@@ -0,0 +1,301 @@
...
+def build_model_config(common: CommonComponents) -> TransformerConfig:
+    """
+    Note for DG/PW: Created a custom model (basically just copied what Pete sent me in slack)
+    """
+    return TransformerConfig.love2code_3B(
+        vocab_size=common.tokenizer.padded_vocab_size(),
+        compile=True,
+        dp_config=TransformerDataParallelConfig(
+            name=DataParallelType.hsdp, param_dtype=DType.bfloat16, reduce_dtype=DType.float32
+        ),
```
dirkgr, line 125: Set number of replicas to number of nodes.

src/scripts/train/learn2code/love2code-3B.py
```diff
@@ -0,0 +1,301 @@
...
+    del common
+    return AdamWConfig(
+        lr=4e-4,
+        weight_decay=0.1,
+        betas=(0.9, 0.95),
+        group_overrides=[
+            OptimGroupOverride(params=["embeddings.weight"], opts=dict(weight_decay=0.0))
+        ],
+        fused=True,
+    )
```
dirkgr, line 145: Does Granite have any guidelines here? 

LRs it still a black art. 7B was fine with up to `12e-4`. So I think you can go higher here. Let's do `12e-4` for this one as well. You're going to train these to the end before thinking about the results, right?

src/scripts/train/learn2code/love2code-3B.py
```diff
@@ -0,0 +1,301 @@
...
+
+def build_trainer_config(common: CommonComponents) -> TrainerConfig:
+    """ Note for DG/PW: The only think I changed here was to add the max_duration in TrainerConfig """
+    return (
+        TrainerConfig(
+            save_folder=common.save_folder,
+            rank_microbatch_size=8 * 4096,
+            save_overwrite=True,
+            metrics_collect_interval=10,
+            cancel_check_interval=1,
```
dirkgr: Set to 10. Might not be necessary, but just in case.

src/scripts/train/learn2code/love2code-3B.py
```diff
@@ -0,0 +1,301 @@
...
+                ephemeral_save_interval=1000,
+                save_async=True,
+            ),
+        )
+        .with_callback(
+            "comet",
+            CometCallback(
+                name=common.run_name,
+                workspace="ai2",
+                project="OLMo-core-1B",
```
dirkgr: Put it into a different project?

src/scripts/train/learn2code/love2code-3B.py
```diff
@@ -0,0 +1,301 @@
...
+                enabled=True,
+                cancel_check_interval=10,
+            ),
+        )
+        .with_callback(
+            "wandb",
+            WandBCallback(
+                name=common.run_name,
+                entity="ai2-llm",
+                project="OLMo-core-1B",
```
dirkgr: same

dirkgr (changes requested):
Didn't get all the way though. Will do more later.

src/olmo_core/nn/transformer/config.py
revbucket: Au contraire. Setting hidden_size_multiplier to 1.5 actually means a 4x upscale for FFN because there's an implicit multiplication by 8/3 in here. Don't @me here, this is a pete decision

src/olmo_core/nn/transformer/config.py
revbucket: I built the model locally and asked it how many parameters it had

src/scripts/train/learn2code/love2code-3B.py
revbucket: Copy from granite. We should go bigger, but trying to be faithful in the architecture repro

src/olmo_core/nn/transformer/config.py
dirkgr: Can you go fancy and do this calculation inside of this script?

dirkgr:
Actually, the rest of the script was just boiler plate, so I guess I'm done after all.

src/scripts/train/learn2code/love2code-3B.py
revbucket: > For the phase-1 pretraining, the learning rate follows a cosine schedule starting from 3 × 10−4 which decays to 3× 10−5 with an initial linear warmup step of 2k iterations.

From Granite. So I guess we jut replicate?

src/scripts/train/learn2code/love2code-3B.py
revbucket: This may or may not be necessary. `num_replicas` seems to default to num_nodes

Asked Pete about it, so we'll see when he gets back

src/scripts/train/learn2code/love2code-3B.py
revbucket: Okay, switched to 12e-4, linearly decaying to 12e-5, with 2k steps of warmup.

src/scripts/train/learn2code/love2code-3B.py
dirkgr: You are 100% correct.

PR #169, opened 2025-02-19
allow dynamic batch sizes

epwalsh:
This would allow us to change the batch size in the middle of a run (without a restart) through a callback that just sets `trainer.data_loader.global_batch_size` to something else.


epwalsh:
Closing in favor of #170, since we can just as well support this in v1.

PR #174, opened 2025-02-21
Add support for context parallelism

epwalsh:
To enable, run with `--train_module.cp_config='{degree: 2}'` (adjust the degree as needed). Probably don't try this with TP at the same time for now.

### DIFF
--- a/src/olmo_core/distributed/parallel/__init__.py
+++ b/src/olmo_core/distributed/parallel/__init__.py
@@ -125,7 +125,7 @@ def build_device_mesh(
     if cp is not None:
         if cp.degree < 1 or dp_world_size % cp.degree != 0:
             raise OLMoConfigurationError(
-                f"{tp.__class__.__name__}.degree must be at least 1 and divide into the world size"
+                f"{cp.__class__.__name__}.degree must be at least 1 and divide into the world size"
             )
         dp_world_size //= cp.degree
     if tp is not None:
--- a/src/olmo_core/distributed/parallel/__init__.py
+++ b/src/olmo_core/distributed/parallel/__init__.py
@@ -10,8 +10,10 @@ from olmo_core.utils import get_default_device
 
 from .context_parallel import (
     ContextParallelConfig,
+    RingAttentionRotateMethod,
     context_parallel_enabled,
     context_parallel_manager,
+    set_ring_attention_rotate_method,
 )
 from .data_parallel import DataParallelConfig, DataParallelType, DPMeshDimName
 from .expert_parallel import ExpertParallelConfig
@@ -41,8 +43,10 @@ __all__ = [
     "PipelineScheduleType",
     "PipelineSchedule",
     "ContextParallelConfig",
+    "RingAttentionRotateMethod",
     "context_parallel_manager",
     "context_parallel_enabled",
+    "set_ring_attention_rotate_method",
 ]
 
 log = logging.getLogger(__name__)
--- a/src/olmo_core/distributed/parallel/__init__.py
+++ b/src/olmo_core/distributed/parallel/__init__.py
@@ -8,7 +8,11 @@ from olmo_core.distributed.utils import get_num_nodes, get_world_size
 from olmo_core.exceptions import OLMoConfigurationError
 from olmo_core.utils import get_default_device
 
-from .context_parallel import ContextParallelConfig, create_context_parallel_ctx
+from .context_parallel import (
+    ContextParallelConfig,
+    context_parallel_enabled,
+    context_parallel_manager,
+)
 from .data_parallel import DataParallelConfig, DataParallelType, DPMeshDimName
 from .expert_parallel import ExpertParallelConfig
 from .pipeline_parallel import (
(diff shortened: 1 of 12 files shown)


src/olmo_core/distributed/parallel/__init__.py
```diff
@@ -112,6 +122,12 @@ def build_device_mesh(
                 f"{pp.__class__.__name__}.degree must be at least 1 and divide into the world size"
             )
         dp_world_size //= pp.degree
+    if cp is not None:
+        if cp.degree < 1 or dp_world_size % cp.degree != 0:
+            raise OLMoConfigurationError(
+                f"{tp.__class__.__name__}.degree must be at least 1 and divide into the world size"
```
2015aroras:
```suggestion
                f"{cp.__class__.__name__}.degree must be at least 1 and divide into the world size"
```

src/olmo_core/train/train_module/transformer_pipeline.py
```diff
@@ -792,16 +792,22 @@ def model_forward(
...
+        # recompiling later.
+        # In theory this could harm performance a bit when input sizes are actually static
+        # but so far I haven't noticed any dip in throughput with the models I've tested.
+        mark_dynamic(batch["input_ids"], (0, 1))
+        if "doc_lens" in batch:
+            mark_dynamic(batch["doc_lens"], (0, 1))
+        for b in attn_buffers.values():
+            mark_dynamic(b, 0)
+
+        with self._model_forward_context():
```
2015aroras, line 830: It looks like PP doesn't support CP yet. Maybe cause a config error in that case.

src/olmo_core/train/train_module/transformer.py
```diff
@@ -560,9 +592,14 @@ def train_batch(self, batch: Dict[str, Any], dry_run: bool = False):
 
         # Train one micro-batch at a time.
         for micro_batch_idx, micro_batch in enumerate(micro_batches):
-            with self._train_microbatch_context(micro_batch_idx, num_micro_batches):
+            attn_buffers = self.model.get_attn_buffers(
```
2015aroras: Is there a need to pass the attention buffers back through `forward` in this manner? I thought the context parallelism would modify them in-place. I would prefer not doing this if it's not needed.

src/olmo_core/nn/transformer/model.py
```diff
@@ -255,6 +264,7 @@ def forward(
         input_ids: torch.Tensor,
         doc_lens: Optional[torch.Tensor] = None,
         max_doc_lens: Optional[Sequence[int]] = None,
+        **attn_buffers,
```
2015aroras: I think it would be better to not have `attn_buffers` swallow up all the kwargs. I feel like this will make it easier to do something wrong. My thoughts for alternatives are:
1. `attn_buffers: dict[str, Tensor] = None`
2. not an argument (as I don't think it's needed)

src/olmo_core/nn/transformer/model.py
```diff
@@ -161,6 +161,15 @@ def compute_auxiliary_metrics(
     def reset_auxiliary_metrics(self):
         pass
 
+    def get_attn_buffers(
```
2015aroras: I'm sort of not a fan of `get_attn_buffers` naming since it restricts our buffer retrieval to just attention. This is what we want for now, but at a high-level CP is looking for sequence-wide buffers rather than attention's buffers.

src/olmo_core/nn/transformer/model.py
```diff
@@ -161,6 +161,15 @@ def compute_auxiliary_metrics(
...
+        self, max_seq_len: int, device: Optional[torch.device] = None
+    ) -> Dict[str, torch.Tensor]:
+        device = device or self.device
+        for block in self.blocks.values():
+            block = cast(TransformerBlockBase, block)
+            return block.get_attn_buffers(max_seq_len, device) or {}
```
2015aroras: thought: What if we retrieve buffers from `BufferCache` instead? Maybe we can also mark which buffers are CP-suitable in that class. Then CP can retrieve the buffers it needs in an architecture-independent manner (which isn't super important but is nice imo).

2015aroras (approved):
I don't see any reason why this wouldn't work, but some parts might be less simple/natural than they can be. Approving so you can be unblocked for now.

src/olmo_core/train/train_module/transformer.py
epwalsh: This API needs some improvement but there's a couple reasons I chose to pass these buffers through the forward method:
1. This allows the `TransformerTrainModule` to mark their sizes as dynamic for `torch.compile()`. This is irrelevant to context parallelism but it's a good thing to do in general. Without this I think we were getting some unnecessary re-compilations during in-loop evals.
2. It guarantees we don't create duplicate versions of these buffers, which could happen if the `BufferCache` owned by each `Attention` module was a different instance. In our usual constructors we make sure the cache is shared, but if someone instantiated their model differently, the cache might not be shared.

epwalsh: Alright refactored this so that the buffers are managed completely by the model.

PR #177, opened 2025-02-26
Add a `BeakerCallback`

epwalsh:
Adds a callback that updates the description of the corresponding beaker experiment with a progress indicator and a URL to the experiment on W&B or Comet.

image

### DIFF
--- a/src/olmo_core/train/callbacks/beaker.py
+++ b/src/olmo_core/train/callbacks/beaker.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import time
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, ClassVar, Optional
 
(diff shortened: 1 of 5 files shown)


src/olmo_core/train/callbacks/beaker.py
```diff
@@ -0,0 +1,118 @@
...
+
+@dataclass
+class BeakerCallback(Callback):
+    """
+    Adds metadata to the Beaker experiment description when running as a Beaker batch job.
+    """
+
+    priority: ClassVar[int] = min(CometCallback.priority - 1, WandBCallback.priority - 1)
+    experiment_id: Optional[str] = None
+    update_interval: Optional[int] = None
```
soldni, line 31: I don't know enough about beaker APIs, but do we have to worry about how often it updates? I can imagine for small 1b runs, this might update multiple time per second, and maybe beaker APIs are unhappy about that.

soldni (approved):

src/olmo_core/train/callbacks/beaker.py
epwalsh: Good point. 4bece53

PR #190, opened 2025-03-06
Steps per epoch convenience

Hgherzog:
- convenience method to translate number of epochs to number of steps 
- for our workflow we train for many epochs and the number of steps often changes with different datasets so this calculation is used to in loop evals etc


epwalsh (approved):
LGTM

PR #207, opened 2025-03-12
Assorted improvements

dirkgr:
These were going stale, so rather than have that happen again, let's merge them.

### DIFF
--- a/src/olmo_core/internal/common.py
+++ b/src/olmo_core/internal/common.py
@@ -2,6 +2,7 @@ import logging
 from functools import lru_cache
 from typing import List, Optional
 
+import torch
 from beaker import Beaker, BeakerError
 
 from olmo_core.io import is_url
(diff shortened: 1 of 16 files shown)


src/olmo_core/internal/common.py
```diff
@@ -132,6 +133,8 @@ def build_launch_config(
 def get_gpu_type(cluster: str) -> str:
     if cluster in CLUSTER_TO_GPU_TYPE:
         return CLUSTER_TO_GPU_TYPE[cluster]
+    elif cluster == "local":
+        return str(torch.get_default_device()).split(":")[0]
```
epwalsh: Looks like this does the same thing as:
```suggestion
        return torch.get_default_device().type
```

src/olmo_core/internal/model_ladder.py
```diff
@@ -57,13 +60,26 @@ def run(self, config: LadderRunConfig):
...
+        elif self in (SubCmd.train, SubCmd.train_single):
+            if self == SubCmd.train_single:
+                for parallelism_strategy in {"d", "t", "e", "c"}:
+                    setting_name = f"{parallelism_strategy}p_config"
+                    log.warning(
+                        "%s is set to %s, but you can't use data parallelism when running on a single node. Disabling.",
+                        setting_name,
+                        getattr(config.train_module, setting_name),
+                    )
+                    setattr(config.train_module, setting_name, None)
```
epwalsh: So you'll still get a warning even if all of those were `None` in the first place?

dirkgr: Argh, messed that up during refactoring. Should be fixed.

epwalsh (approved):
LGTM

PR #208, opened 2025-03-12
upload data to r2 and updated their paths

aman-17:
Uploaded dolmino and jallyrun data to r2 and updated the paths in .txt


epwalsh (approved):
Thanks Aman!

PR #214, opened 2025-03-12
Add a public "official" version of our 32B train script

epwalsh:
Changes:
- Created a new subdirectory `src/scripts/official` to put official public versions of training scripts. These scripts should be usable by non-Ai2 people out-of-the-box.
- Following that I've added a public copy of the 32B training script to that folder. This is just a refactored version of `src/scripts/train/OLMo2-32B.py` that's completely self-contained (doesn't depend on `olmo_core.internal`) and uses the public paths to the data.
- I also removed the throughput table in the README so external people don't try to run those scripts.


aman-17 (approved):
Thanks for adding these Pete. Makes my job easier.

PR #219, opened 2025-03-13
Ignore `group_overrides` for frozen params

AkshitaB:


src/olmo_core/optim/config.py
```diff
@@ -109,11 +112,20 @@ def build_groups(
...
-                        raise OLMoConfigurationError(msg)
-                    else:
+                    frozen_param_matches = False
+                    for name in frozen_params:
+                        if fnmatch(name, pattern):
+                            frozen_param_matches = True
+                            break
+                    if frozen_param_matches:
+                        msg = f"optim group {g_idx} override pattern '{pattern}' matches a frozen parameter and will be ignored"
                         log.warning(msg)
```
epwalsh:
```suggestion
                    for name in frozen_params:
                        if fnmatch(name, pattern):
                            log.warning(f"optim group {g_idx} override pattern '{pattern}' matches a frozen parameter and will be ignored")
                            break
```

PR #226, opened 2025-03-19
OLMo Core to HF conversion refactor

2015aroras:
Issue: Our conversion logic between OLMo Core and HF transformers formats is not very flexible. Every time we want to support a new architecture, we have to write new logic and test from scratch. The ideal is to have an easy way to specify how OLMo Core states maps to/from HF, and have our code use that mapping information to perform the conversion.

Fix: This PR refactors the logic around conversion to make it easier to convert between the formats. It enables us to specify mappings between states of the 2 formats that support placeholders for things like expert id and layer id (in `src/olmo_core/nn/hf/convert.py` using `StateMappingTemplate`s). The mappings can be many-to-many and allow us to specify some operations that need to be performed as part of conversion, like permuting dimensions or merging/unmerging layers. The operations currently supported are the minimum needed to support OLMo2 and the current MoE implementation.

### DIFF
--- a/src/examples/huggingface/convert_checkpoint_from_hf.py
+++ b/src/examples/huggingface/convert_checkpoint_from_hf.py
@@ -8,165 +8,368 @@ HuggingFace.
 
 import json
 import logging
-import os
+import types
+from argparse import ArgumentParser
+from functools import partial
+from pathlib import Path
+from tempfile import TemporaryDirectory
+from typing import Any, Dict, Optional
 
 import torch
+from cached_path import cached_path
+from torch.distributed import DeviceMesh
 from transformers import AutoModelForCausalLM
 
+from olmo_core.aliases import PathOrStr
 from olmo_core.data.tokenizer import TokenizerConfig
-from olmo_core.distributed.checkpoint import load_model_and_optim_state, save_state_dict
-from olmo_core.io import clear_directory, dir_is_empty
-from olmo_core.nn.rope import RoPEScalingConfig
-from olmo_core.nn.transformer import TransformerConfig
+from olmo_core.io import file_exists
+from olmo_core.nn.conversion.state_mapping import TemplatePlaceholder
+from olmo_core.nn.hf.checkpoint import load_hf_model
+from olmo_core.nn.hf.convert import get_converter_from_hf
+from olmo_core.nn.transformer.config import TransformerConfig
+from olmo_core.nn.transformer.model import Transformer
+from olmo_core.optim.adamw import AdamWConfig
+from olmo_core.train.checkpoint import CheckpointerConfig
+from olmo_core.train.train_module.transformer import TransformerTrainModuleConfig
 from olmo_core.utils import get_default_device, prepare_cli_environment
 
 log = logging.getLogger(__name__)
 
-HF_MODEL = "allenai/OLMo-2-1124-7B"
-# HF_MODEL = "allenai/OLMo-2-1124-7B-Instruct"
-# HF_MODEL = "allenai/OLMo-2-1124-13B-Instruct"
-# HF_MODEL = "meta-llama/Llama-3.2-1B"
-# HF_MODEL = "meta-llama/Llama-3.2-8B"
 
-SAVE_PATH = f"/tmp/checkpoints/{HF_MODEL}"
-SAVE_OVERWRITE = False
-
-TOKENIZER_CONFIG = TokenizerConfig.from_hf(HF_MODEL)
-MODEL_CONFIG: TransformerConfig
-if HF_MODEL == "meta-llama/Llama-3.2-1B":
-    MODEL_CONFIG = TransformerConfig.llama3_1B(
-        TOKENIZER_CONFIG.vocab_size,
-        fused_ops=False,
-        use_flash=False,
-        rope_scaling=RoPEScalingConfig(),
-    )
-elif HF_MODEL.startswith("allenai/OLMo-2-1124-7B"):
-    MODEL_CONFIG = TransformerConfig.olmo2_7B(
-        TOKENIZER_CONFIG.vocab_size,
-        fused_ops=False,
-        use_flash=False,
-    )
-elif HF_MODEL.startswith("allenai/OLMo-2-1124-13B"):
-    MODEL_CONFIG = TransformerConfig.olmo2_13B(
-        TOKENIZER_CONFIG.vocab_size,
-        fused_ops=False,
-        use_flash=False,
-    )
-else:
-    raise NotImplementedError(HF_MODEL)
-
-
-def convert_checkpoint() -> AutoModelForCausalLM:
-    log.info(f"Loading HF checkpoint '{HF_MODEL}'")
-    hf_model = AutoModelForCausalLM.from_pretrained(HF_MODEL)
-    print(hf_model)
-
-    if not dir_is_empty(SAVE_PATH):
-        if SAVE_OVERWRITE:
-            log.warning(f"Clearing existing checkpoint at '{SAVE_PATH}'")
-            clear_directory(SAVE_PATH)
-        else:
-            log.warning(f"Using existing checkpoint at '{SAVE_PATH}'")
-            return hf_model
-
-    n_layers = len(hf_model.model.layers)
-    state_dict = hf_model.state_dict()
-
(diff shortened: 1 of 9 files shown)


src/olmo_core/nn/conversion/state_mapping.py
```diff
@@ -0,0 +1,250 @@
...
+    EXPERT = "[expert]"
+
+
+@dataclass
+class StateMappingTemplate:
+    """
+    The template for a mapping from state from one format to another format (e.g. OLMo Core to HF).
+    These mappings are 'templates' since they support keys and other metadata having placeholders
+    for information like the layer number or number of MoE experts. This class can be converted
+    to a `StateMapping` by providing the placeholder information information.
```
epwalsh: minor typos:
```suggestion
    The template for a mapping state from one format to another format (e.g. OLMo Core to HF).
    These mappings are 'templates' since they support keys and other metadata having placeholders
    for information like the layer number or number of MoE experts. This class can be converted
    to a `StateMapping` by providing the placeholder information.
```

src/olmo_core/nn/conversion/state_mapping.py
```diff
@@ -0,0 +1,250 @@
...
+    The key or keys of the state(s) being mapping from.
+    """
+    dest_template_keys: str | Tuple[str, ...]
+    """
+    The key or keys of the state(s) being mapping to.
+    """
+
+    source_key_per_placeholder: TemplatePlaceholder | None = None
+    """
+    A placeholder in `source_template_keys` for which this mapping should map all valid placeholder
```
epwalsh: Annoyingly ReStructured Text requires double backticks for code formatting.

```suggestion
    A placeholder in ``source_template_keys`` for which this mapping should map all valid placeholder
```

src/olmo_core/nn/conversion/state_converter.py
```diff
@@ -0,0 +1,130 @@
...
+)
+
+
+class StateConverter:
+    """
+    A class for converting state from one format to another format (e.g. OLMo Core to HF).
+    """
+
+    def __init__(self, mapping_templates: List[StateMappingTemplate]) -> None:
+        self.mapping_templates = mapping_templates
```
epwalsh: Could just make this a dataclass to avoid the `__init__()` ?

src/olmo_core/nn/hf/checkpoint.py
```diff
@@ -0,0 +1,159 @@
...
+    barrier(group=process_group)
+
+    hf_model = AutoModelForCausalLM.from_pretrained(model_name_or_path)
+    log.info(f"Loaded hf model: {hf_model}")
+
+    converted_state_dict: Dict[str, torch.Tensor] = convert_state_from_hf(
+        hf_model.config, hf_model.state_dict(), model_id=model_id
+    )
+
+    for key, state in converted_state_dict.items():
```
epwalsh: Since there's potentially some distributed communication here you'll need to make sure the keys are traversed in the same order on each rank.
```suggestion
    for key in sorted(converted_state_dict.keys()):
        state = converted_state_dict[key]
```

src/examples/huggingface/convert_checkpoint_from_hf.py
```diff
@@ -8,165 +8,372 @@
...
+        self.to_empty(device=device)
+
+        for module in self.modules():
+            if hasattr(module, "reset_parameters"):
+                module.to_empty(device=device)  # type: ignore
+
+        seed = self.init_seed
+        if pp_mesh is not None:
+            seed += pp_mesh.get_local_rank()
+        return torch.Generator(device).manual_seed(seed)
```
epwalsh: I don't think this is necessary. You could just call `model.to_empty(device=device)`.

src/examples/huggingface/convert_checkpoint_from_hf.py
```diff
@@ -8,165 +8,372 @@
...
+
+    model.init_weights = types.MethodType(init_weights, model)
+
+    device = device or get_default_device()
+
+    train_module = TransformerTrainModuleConfig(
+        rank_microbatch_size=max_sequence_length,
+        max_sequence_length=max_sequence_length,
+        optim=AdamWConfig(),
+    ).build(model, device=device)
```
epwalsh: I know technically the train module decides the format of the checkpoint, but it seems overkill to have to instantiate a train module here. In this case I think it's fair to make assumptions about the format and just use the corresponding functions from `olmo_core.distributed.checkpoint` directly.

epwalsh (commented):
Looks good overall, especially the `nn` API. I left a couple suggestions about how to simplify the example conversion scripts, and I have a couple questions:
- Does this cover removing padding from the embeddings?
- Can you add tests for the conversion scripts on a tiny model?

src/olmo_core/nn/conversion/state_converter.py
2015aroras: Done https://github.com/allenai/OLMo-core/pull/226/commits/197bcca0ff42d125403722b663ea98b7d300f8aa

src/examples/huggingface/convert_checkpoint_from_hf.py
2015aroras: This was necessary due to the use of `TransformerTrainModule`, but since that is gone (https://github.com/allenai/OLMo-core/pull/226/commits/8028909162812d74de16525f7132f0472f0cc4c0) I have fixed this too.

src/examples/huggingface/convert_checkpoint_from_hf.py
2015aroras: Done https://github.com/allenai/OLMo-core/pull/226/commits/8028909162812d74de16525f7132f0472f0cc4c0

2015aroras:
> Does this cover removing padding from the embeddings?

WIP, since this was mentioned fairly recently

> Can you add tests for the conversion scripts on a tiny model?

👍

src/olmo_core/nn/conversion/state_mapping.py
```diff
@@ -0,0 +1,250 @@
...
+    EXPERT = "[expert]"
+
+
+@dataclass
+class StateMappingTemplate:
+    """
+    The template for a mapping state from one format to another format (e.g. OLMo Core to HF).
+    These mappings are 'templates' since they support keys and other metadata having placeholders
+    for information like the layer number or number of MoE experts. This class can be converted
+    to a `StateMapping` by providing the placeholder information.
```
epwalsh:
```suggestion
    to a :class:`StateMapping` by providing the placeholder information.
```

src/olmo_core/nn/conversion/state_mapping.py
```diff
@@ -0,0 +1,250 @@
...
+
+    The most standard mapping is a one-to-one state mapping, which corresponds to a single
+    string entry for both `source_template_keys` and `dest_template_keys`. The class also supports
```
epwalsh: I still see a bunch of instances like this that should have double backticks. You can check how this renders by running `make docs`, though you may need to add a page for this submodule in `docs/source/nn/` (and link to it in `docs/source/nn/index.rst`).

```suggestion
    string entry for both ``source_template_keys`` and ``dest_template_keys``. The class also supports
```

epwalsh (commented):
LGTM, just a couple more doc issues

src/olmo_core/nn/conversion/state_mapping.py
2015aroras: Done a mass update of docs
[b1b6e7e](https://github.com/allenai/OLMo-core/pull/226/commits/b1b6e7e4c40291fef1a03643a003207c2a21208e)

2015aroras:
@epwalsh Gentle ping.

pyproject.toml
```diff
@@ -20,6 +20,7 @@ dependencies = [
     "packaging",
     "rich",
     "omegaconf",
+    "transformers",
```
epwalsh: Can we make this as an optional dependency? I think the HF stuff is pretty self-contained, right?

2015aroras: https://github.com/allenai/OLMo-core/pull/226/commits/ddca1b0892d006d332d3c8308ccf933d5dd90eb0

PR #228, opened 2025-03-20
Update float8 API for latest torchao

epwalsh:


epwalsh:
@AkshitaB are you using FP8 right now? This slightly changes how it's configured, but after going through torchao code again today I'm less excited about using it in general...

AkshitaB:
> @AkshitaB are you using FP8 right now? This slightly changes how it's configured, but after going through torchao code again today I'm less excited about using it in general...

@epwalsh Yes, we do use FP8 . What would you recommend as an alternative?

epwalsh:
@AkshitaB safest to just turn it off for now. My main concern is that the default settings use tensor-wise scaling, which may lead to performance degradation with large models. It would be better to use row-wise scaling, which is an option, but I don't think it plays well with FSDP at the moment.

PR #230, opened 2025-03-22
Pick the correct remote

dirkgr:


src/olmo_core/launch/utils.py
```diff
@@ -27,7 +27,19 @@ def ensure_repo(allow_dirty: bool = False) -> Tuple[str, str, str, bool]:
...
-    account, repo = parse_git_remote_url(repo.remote().url)
+
+    remote = repo.remote()
+    # Try to find a remote based on the current tracking branch.
+    try:
+        branch = repo.active_branch
+    except TypeError:
+        branch = None
+    if branch is not None:
+        branch = branch.tracking_branch()
```
epwalsh, line 38: This is a bit confusing, what about:
```suggestion
        branch = repo.active_branch.tracking_branch()
    except TypeError:
        branch = None
```

dirkgr: `TypeError` is very general. I don't want to catch stray type errors from the other line.

epwalsh (approved):

PR #234, opened 2025-03-25
Port the 1B from old OLMo

dirkgr:


src/olmo_core/nn/transformer/config.py
```diff
@@ -643,13 +643,17 @@ def llama2_271M(cls, vocab_size: int, **kwargs) -> "TransformerConfig":
...
+        d_model = kwargs.pop("d_model", 2048)
         return cls.llama_like(
-            d_model=2048,
+            d_model=d_model,
             vocab_size=vocab_size,
-            n_layers=kwargs.pop("n_layers", 18),
+            n_layers=kwargs.pop("n_layers", 16),
             n_heads=kwargs.pop("n_heads", 16),
             rope_theta=kwargs.pop("rope_theta", 10_000),
+            hidden_size_multiplier=kwargs.pop("hidden_size_multiplier", 8192 / (8 * d_model / 3)),
```
epwalsh: You can just do this:
```suggestion
            hidden_size_multiplier=kwargs.pop("hidden_size_multiplier", 1.5),
```

See the `llama3_1B()` method.

src/olmo_core/optim/adamw.py
```diff
@@ -147,6 +147,7 @@ class SkipStepAdamWConfig(OptimConfig):
     betas: Tuple[float, float] = (0.9, 0.999)
     eps: float = 1e-8
     weight_decay: float = 1e-2
+    fused: Optional[bool] = None
```
epwalsh: There's no fused option for skip step

epwalsh (approved):
LGTM other than the change to `SkipStepAdamWConfig`

src/olmo_core/optim/adamw.py
dirkgr: It takes that as a parameter though.

epwalsh: Oh, it shouldn't because it doesn't actually use that param, as well as `foreach`.

dirkgr: done

PR #241, opened 2025-04-02
Add support for label masks with numpy datasets

epwalsh:
Ports over support for custom label masks from old OLMo. You can enable this by setting `label_mask_paths` to a list of paths/URLs to numpy bool files in your `NumpyDatasetConfig`. Currently only FSL and padded FSL datasets support this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `auto_resume` option to `CometCallback` for resume an existing run.
 - (BETA) Added methods `load_hf_model` and `save_hf_model` for saving supported OLMo Core models to HF transformers format.
 Also added lower-level methods for converting state between the formats.
-- Added support for label mask files with numpy datasets.
+- Added support for label mask files with numpy FSL datasets.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `auto_resume` option to `CometCallback` for resume an existing run.
 - (BETA) Added methods `load_hf_model` and `save_hf_model` for saving supported OLMo Core models to HF transformers format.
 Also added lower-level methods for converting state between the formats.
+- Added support for label mask files with numpy datasets.
 
 ### Changed
 
--- a/src/olmo_core/data/numpy_dataset.py
+++ b/src/olmo_core/data/numpy_dataset.py
@@ -823,6 +823,9 @@ class NumpyPaddedFSLDataset(NumpyFSLDataset):
         return data
 
     def _get_instance_indices_path(self, path: PathOrStr) -> Path:
+        # NOTE: the instance indices file names are based on the corresponding source (token IDs) file name,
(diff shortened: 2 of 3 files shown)


src/olmo_core/data/numpy_dataset.py
```diff
@@ -1723,6 +1783,7 @@ def build(self) -> NumpyDatasetBase:
             paths, labels = mix.build(self.mix_base_dir, self.tokenizer.identifier)
             if metadata is None:
                 metadata = [{"label": label} for label in labels]
+            label_mask_paths = cast(Optional[List[PathOrStr]], self.label_mask_paths)
```
2015aroras, line 1789: thought: Is it possible to make the label mask paths support (non-fractionalized) mix-based construction too?

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `auto_resume` option to `CometCallback` for resume an existing run.
 - (BETA) Added methods `load_hf_model` and `save_hf_model` for saving supported OLMo Core models to HF transformers format.
 Also added lower-level methods for converting state between the formats.
+- Added support for label mask files with numpy datasets.
```
2015aroras: nit: specify the FSL requirement here

src/olmo_core/data/numpy_dataset.py
```diff
@@ -766,22 +804,27 @@ def prepare(self):
...
         indices = load_array_slice_into_tensor(
             indices_path, index * 2, index * 2 + 2, self.indices_dtype
         )
         start_idx, end_idx = indices
-        data = load_array_slice_into_tensor(path, int(start_idx), int(end_idx), self.dtype)
+        data = load_array_slice_into_tensor(path, int(start_idx), int(end_idx), dtype or self.dtype)
         return data
 
     def _get_instance_indices_path(self, path: PathOrStr) -> Path:
+        if path in self._label_mask_path_to_source_path:
```
2015aroras, line 829: What scenario needs this? I was scared that this would mess up the initial mask setup, but the dictionary is created after the setup.

2015aroras (approved):

src/olmo_core/data/numpy_dataset.py
epwalsh: This is kind of ugly, but it's because the "instance indices" files are named based on the corresponding source file path. So to get the right instance indices file for a label mask, we need to map the label mask path to the corresponding source file path. I'll add a comment explaining this.

src/olmo_core/data/numpy_dataset.py
epwalsh: Probably yes but I think I'll hold off on that until we need it.

src/olmo_core/data/numpy_dataset.py
2015aroras: Why support this? Why not throw an error if we try to get instance indices for a label mask path? (I don't know the dataloader logic very well).

soldni (approved):
LGTM!

src/olmo_core/data/numpy_dataset.py
epwalsh: We need to get instance indices for label masks in the same way we need them for token IDs. See the `__getitem__` method.

PR #243, opened 2025-04-03
fix in-loop normalization with v2

davidheineman:
Merge and update in-loop evals to match https://github.com/allenai/OLMo-in-loop-evals/pull/6

Description from the `ai2-olmo-eval==0.7.1` PR:
> our current in-loop does not match the OLMES standard (in our downstream evals), [the downstream normalization is here](https://github.com/allenai/oe-eval-internal/blob/main/oe_eval/metrics/metric.py#L229-L231). It appears the -1 term is an artifact from when we originally implemented in-loop evals.

This will keep the original version of the metrics (`_v1`) and the version with the fixed normalization (`_v2`)

### DIFF
--- a/src/olmo_core/train/callbacks/evaluator_callback.py
+++ b/src/olmo_core/train/callbacks/evaluator_callback.py
@@ -246,14 +246,23 @@ class LMEvaluatorCallbackConfig(CallbackConfig):
 
 class DownstreamEvaluator(Evaluator):
     metric_type_to_label = {
-        "f1": "F1 score",
-        "acc": "accuracy",
-        "len_norm": "length-normalized accuracy",
-        "pmi_dc": "PMI-DC accuracy",
-        "ce_loss": "CE loss",
-        "bpb": "BPB",
-        "soft": "soft loss",
-        "soft_log": "log soft loss",
+        "f1_v1": "F1 score v1",
+        "acc_v1": "accuracy v1",
+        "len_norm_v1": "length-normalized accuracy v1",
+        "pmi_dc_v1": "PMI-DC accuracy v1",
+        "ce_loss_v1": "CE loss v1",
+        "bpb_v1": "BPB v1",
+        "soft_v1": "soft loss v1",
+        "soft_log_v1": "log soft loss v1",
+
+        "f1_v2": "F1 score v2",
+        "acc_v2": "accuracy v2",
+        "len_norm_v2": "length-normalized accuracy v2",
+        "pmi_dc_v2": "PMI-DC accuracy v2",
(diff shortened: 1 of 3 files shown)


src/olmo_core/train/callbacks/evaluator_callback.py
```diff
@@ -246,14 +246,22 @@ def build(self, trainer: "Trainer") -> Optional[Callback]:
...
-        "soft": "soft loss",
-        "soft_log": "log soft loss",
+        "f1_v1": "F1 score v1",
+        "acc_v1": "accuracy v1",
+        "len_norm_v1": "length-normalized accuracy v1",
+        "pmi_dc_v1": "PMI-DC accuracy v1",
+        "ce_loss_v1": "CE loss v1",
+        "bpb_v1": "BPB v1",
+        "soft_v1": "soft loss v1",
+        "soft_log_v1": "log soft loss v1",
```
epwalsh: Would be good to keep the original names for logging purposes.

src/olmo_core/train/callbacks/evaluator_callback.py
```diff
@@ -246,14 +246,22 @@ def build(self, trainer: "Trainer") -> Optional[Callback]:
...
-        "f1": "F1 score",
-        "acc": "accuracy",
-        "len_norm": "length-normalized accuracy",
-        "pmi_dc": "PMI-DC accuracy",
-        "ce_loss": "CE loss",
-        "bpb": "BPB",
-        "soft": "soft loss",
-        "soft_log": "log soft loss",
+        "f1_v1": "F1 score",
+        "acc_v1": "accuracy",
```
OyvindTafjord, line 250: Can the "acc" metric depend on this prefix space?

davidheineman: No, `acc_v1` will be the same as `acc_v2` but I assume it's just better to update everything to avoid confusion (as others will be different)?

dirkgr: Can we leave the old names alone? That way they will line up correctly in WandB & CometML.

dirkgr (changes requested):

src/olmo_core/train/callbacks/evaluator_callback.py
davidheineman: This will leave the names in WandB the same. The values of this dict are what is reported as the metric by OLMo-core.

dirkgr: Ah, got it.

dirkgr (approved):

PR #244, opened 2025-04-03
Add option to cancel run after first evals

epwalsh:
You can use this feature to only run in-loop evals without training any longer. For example, add these flags to your launch command:

```bash
  --trainer.callbacks.lm_evaluator.eval_on_startup=true \
  --trainer.callbacks.lm_evaluator.cancel_after_first_eval=true \
  --trainer.callbacks.downstream_evaluator.eval_on_startup=true \
  --trainer.callbacks.downstream_evaluator.cancel_after_first_eval=true
```

This also includes some minor fixes to how we log the speed of in-loop evals.

### DIFF
--- a/src/olmo_core/train/callbacks/evaluator_callback.py
+++ b/src/olmo_core/train/callbacks/evaluator_callback.py
@@ -126,13 +126,6 @@ class EvaluatorCallback(Callback):
                     with cuda_sync_debug_mode(0):
                         evaluator.update_metrics(batch, ce_loss, logits)
 
-                if (
-                    not self.cancel_after_first_eval
-                ):  # guard here in case we have multiple evaluators doing this
-                    if self.trainer.is_canceled:
-                        self._log_progress(evaluator, eval_step)
-                        return
-
                 if self.eval_duration.due(step=eval_step, tokens=eval_tokens, epoch=1):
                     self._log_progress(evaluator, eval_step)
                     break
--- a/src/olmo_core/train/callbacks/evaluator_callback.py
+++ b/src/olmo_core/train/callbacks/evaluator_callback.py
@@ -152,7 +152,7 @@ class EvaluatorCallback(Callback):
                     self.trainer.record_metric(f"eval/{evaluator.name}/{name}", value)
 
             evaluator_times.append(time.monotonic() - start_time)
-            evaluator_names.append(evaluation_names)
+            evaluator_names.append(evaluator.name)
             evaluator_bs.append(eval_step)
 
             log.info(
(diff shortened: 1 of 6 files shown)


src/olmo_core/train/trainer.py
```diff
@@ -529,15 +529,21 @@ def training_progress(self) -> TrainingProgress:
...
+        :param no_sync: Set this to ``True`` only if you're calling this from all ranks at the same
+            time, otherwise you'll get a distributed deadlock.
         """
-        #  self._canceled = True  # NOTE: important not to set this!! Leads to distributed hang.
         self._canceling_rank = get_rank()
         self._cancel_reason = reason
+        if no_sync:
+            self._canceled = (
+                True  # NOTE: important not to set this otherwise!! Leads to distributed hang.
+            )
```
dirkgr: This is confusing. The comment says to not set this, but it is being set?

src/olmo_core/train/callbacks/evaluator_callback.py
```diff
@@ -172,6 +179,9 @@ def _perform_eval(self):
         self.trainer.record_metric("throughput/in-loop eval time (s)", total_time)
         self.trainer.record_metric("throughput/in-loop eval batches", total_bs)
 
+        if self.cancel_after_first_eval:
+            self.trainer.cancel_run("canceled from evaluator callback", no_sync=True)
```
dirkgr: Maybe a clearer message, like "canceled because `cancel_after_first_eval` is set"?

dirkgr (changes requested):

src/olmo_core/train/callbacks/evaluator_callback.py
epwalsh: Done

src/olmo_core/train/trainer.py
epwalsh: Done

dirkgr:
Things are happening in this PR? Did you find an issue?

epwalsh:
I'm done editing, just found some other things to fix with how the in-loop eval speed is logged. Ready for review again.

src/olmo_core/train/callbacks/evaluator_callback.py
```diff
@@ -119,16 +126,18 @@ def _perform_eval(self):
...
                         evaluator.update_metrics(batch, ce_loss, logits)
 
-                if eval_step % self.trainer.cancel_check_interval == 0:
-                    self.trainer.check_if_canceled()
+                if (
+                    not self.cancel_after_first_eval
+                ):  # guard here in case we have multiple evaluators doing this
+                    if self.trainer.is_canceled:
+                        self._log_progress(evaluator, eval_step)
+                        return
```
dirkgr: I find this confusing. If multiple callbacks cancel the trainer, we still want to run them all, right? The canceling only takes effect when it's time to do the next step?

epwalsh: I just removed this block: a8de0efa

We really don't need it. I had it there initially because I thought a run could get canceled during an eval loop, in which case we should break out of the loop early. But now I think it's better to finish the eval loop.

PR #245, opened 2025-04-07
[muP] Mini optimizer group building refactor

2015aroras:
This PR modifies the logic for building optimizer parameter groups in the following manner.
1. It moves the logic for expanding parameter globs of optimizer group overrides into a separate method `_expand_param_globs`.
2. It changes the glob into the corresponding name rather than the actual parameter. Conversion to parameter happens after any other processing, which is currently nothing but will soon be muP processing.
3. It puts no-override parameters into a no-op optimizer override group, so that any new processing just runs on `OptimGroupOverride` objects.

This change sets up the addition of muP to OLMo Core. This change is simply a refactor; there should be no change in training behavior.


src/olmo_core/optim/config.py
```diff
@@ -97,42 +130,21 @@ def build_groups(
...
-
-        return param_groups
+        group_overrides = [
+            self._expand_param_globs(go, all_params, frozen_params, g_idx, strict=strict)
+            for g_idx, go in enumerate(self.group_overrides or [])
+        ]
+
+        # Treat no overrides as its own override group
+        overriden_param_names = {name for go in group_overrides for name in go.params}
+        default_override = OptimGroupOverride(list(all_params.keys() - overriden_param_names), {})
```
epwalsh: The order of parameters within a group should be deterministic to be safe, and I think this line breaks that. Could change to something like:

```suggestion
        default_override = OptimGroupOverride([name for name in all_params.keys() if name not in overriden_param_names], {})
```

epwalsh (approved):
LGTM

PR #250, opened 2025-04-11
clone single branch when possible

epwalsh:
As we're accumulating more and more branches, clone time is getting proportionately slower. This PR makes it so we only clone a single branch at a runtime on Beaker when possible, which is must faster.

- Test run: https://beaker.org/ex/01JRK5BR70QFZG4RGZYQV1XDSN
- References:
  - https://stackoverflow.com/questions/4811434/clone-only-one-branch

### DIFF
--- a/src/olmo_core/launch/beaker.py
+++ b/src/olmo_core/launch/beaker.py
@@ -96,7 +96,11 @@ class BeakerWekaBucket(Config):
 
 
 DEFAULT_SETUP_STEPS = (
-    'git clone "$REPO_URL" .',
+    'if [[ -z "$GIT_BRANCH" ]]; then',
+    '  git clone "$REPO_URL" .',
+    "else",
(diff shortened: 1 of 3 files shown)


src/olmo_core/launch/beaker.py
```diff
@@ -354,13 +358,7 @@ def build_experiment_spec(
             "mkdir -p /root/.cache/torch/kernels && export PYTORCH_KERNEL_CACHE_PATH=/root/.cache/torch/kernels",
             "mkdir -p /olmo-core-runtime",
             "cd /olmo-core-runtime",
-        ]
-        # TODO: remove once we have a base image with CUDA 12.8
```
2015aroras, line 358: I'm assuming this is just some extra cleanup you're sneaking into this PR

2015aroras (approved):

src/olmo_core/launch/beaker.py
epwalsh: Yeup

PR #251, opened 2025-04-11
Add a git config to BeakerLaunchConfig

epwalsh:
This gives us a way to record the git state in our configs.

Test run: https://beaker.org/ex/01JRKBHRCH1H61KB856X6SF1WM

### DIFF
--- a/src/olmo_core/launch/beaker.py
+++ b/src/olmo_core/launch/beaker.py
@@ -348,7 +348,10 @@ class BeakerLaunchConfig(Config):
         Get the Beaker experiment spec corresponding to this config instance.
         """
         if self.git is None:
-            raise OLMoConfigurationError(f"{self.__class__.__name__}.git field is required!")
+            raise OLMoConfigurationError(
+                f"{self.__class__.__name__}.git field is required!\n"
+                "You either need to instantiate your launch config from a valid git repository folder or set the '.git' field manually."
+            )
 
         if self.git.is_dirty and not self.allow_dirty:
             raise RuntimeError(
--- a/src/olmo_core/launch/beaker.py
+++ b/src/olmo_core/launch/beaker.py
@@ -96,12 +96,12 @@ class BeakerWekaBucket(Config):
 
 
 DEFAULT_SETUP_STEPS = (
-    'if [[ -z "$GIT_BRANCH" ]]; then',
-    '  git clone "$REPO_URL" .',
+    f'if [[ -z "${GIT_BRANCH_ENV_VAR}" ]]; then',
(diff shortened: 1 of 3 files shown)


src/olmo_core/launch/beaker.py
```diff
@@ -341,10 +351,14 @@ def build_experiment_spec(
         """
         Get the Beaker experiment spec corresponding to this config instance.
         """
-        # Get repository account, name, and current ref.
-        github_account, github_repo, git_branch, git_ref, is_public = ensure_repo(self.allow_dirty)
+        git = self.git or GitConfig.from_env()
```
2015aroras: nit: A part of me wants this to be an error instead if `self.git` is None, given the post init. As is, if we change how we setup `self.git` in the post init, we would have to do it here too.

src/olmo_core/launch/utils.py
```diff
@@ -19,35 +29,52 @@ def parse_git_remote_url(url: str) -> Tuple[str, str]:
...
 
 
-def ensure_repo(allow_dirty: bool = False) -> Tuple[str, str, Optional[str], str, bool]:
-    import requests
-    from git.repo import Repo
+@dataclass
+class GitConfig(Config):
+    repo_url: str
+    ref: str
+    is_dirty: bool
```
2015aroras: It seems a bit odd to have this configurable. Is it possible to make this a property that can be determined by `repo.is_dirty()` call? Sort of like `is_public`.

epwalsh: Yeup, good idea: 7b9d1284

src/olmo_core/launch/beaker.py
epwalsh: That's fair. I made that changed here: daa728fc

src/olmo_core/launch/beaker.py
```diff
@@ -351,14 +347,15 @@ def build_experiment_spec(
         """
         Get the Beaker experiment spec corresponding to this config instance.
         """
-        git = self.git or GitConfig.from_env()
+        if self.git is None:
+            raise OLMoConfigurationError(f"{self.__class__.__name__}.git field is required!")
```
2015aroras: thought: You could make `self.git` never be None and throw if the dir is not a repo in `from_env`. That would have a clearer error message than this.

2015aroras (approved):

src/olmo_core/launch/beaker.py
epwalsh: I had it this way because I like the flexibility of being able to use the launch config from a non-git directory. But I'll make the error message more clear: 5f24c09d

PR #255, opened 2025-04-15
make LB loss with sigmoid more robust

epwalsh:
The epsilon we used in the denominator when normalizing the sigmoid scores for the LB loss was too big (1e-20, which we took from [Megatron-LM](https://github.com/NVIDIA/Megatron-LM/blob/55c968dabb6d32cf8c21b1e3fadeea47e2ceb78d/megatron/core/transformer/moe/router.py#L195-L199)), allowing the model to hack the LB loss function by pushing sigmoid scores below epsilon.

image

### DIFF
--- a/src/olmo_core/nn/moe/router.py
+++ b/src/olmo_core/nn/moe/router.py
@@ -440,8 +440,6 @@ class MoERouter(nn.Module):
         else:
             raise NotImplementedError(self.gating_function)
 
-        scores = scores.float()  # NOTE: should be FP32 already, but just to be sure.
-
         # shape: (batch_size, seq_len, top_k)
         expert_weights, expert_indices = self.get_top_k(scores)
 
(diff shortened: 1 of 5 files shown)


src/olmo_core/nn/moe/router.py
```diff
@@ -492,7 +494,7 @@ def forward(
             if self.z_loss_weight is not None:
                 assert self.z_loss is not None
                 z_loss = router_z_loss(
-                    expert_logits=logits.float(),
```
epwalsh, line 495: This was redundant.

src/olmo_core/nn/moe/router.py
```diff
@@ -455,10 +458,9 @@ def forward(
                 )
             )
 
-        scores = scores.float()
         # Make sure scores are normalized, otherwise load balancing loss doesn't work.
         if self.gating_function == MoERouterGatingFunction.sigmoid:
-            scores = scores / (scores.sum(dim=-1, keepdim=True) + 1e-20)
+            scores = scores / (scores.sum(dim=-1, keepdim=True) + _EPSILON)
```
2015aroras: Is the issue that the epsilon was too big?

epwalsh: Yeup, surprisingly

PR #260, opened 2025-04-19
Add a numpy FSL dataset variant with document packing

epwalsh:
Based on [Fewer Truncations Improve Language Modeling](https://arxiv.org/pdf/2404.10830). I think this a good option for annealing and context-length extension, and maybe even pretraining as well. You can enable it with the flags:

```bash
--dataset.name=packed_fsl --dataset.long_doc_strategy=truncate  # or "fragment"
```

The idea is to use the best-fit-decreasing bin-packing algorithm to pack documents from each source file into instances of at most `max_sequence_length`, and use padding when needed. There are two options for handling documents that are longer than `max_sequence_length`:
- `truncate`: truncate those docs and discard the excess tokens.
- `fragment`: split those docs into multiple smaller documents so that no tokens are lost but you end up with some fragmented documents.

To test this I trained 3 tiny models on c4 for 5K steps:
- `no-packing`: our default dataset that just concats and chunks up source files to create instances.
- `pack-and-fragment`: the new packed FSL dataset with `long_doc_strategy=fragment`.
- `pack-and-truncate`: the new packed FSL dataset with `long_doc_strategy=truncate`.

All 3 models were trained with intra-document masking enabled and the in-loop c4 PPL eval set was evaluated in the same way for each: using the padded FSL dataset, which means each document is treated as its own instance, padded out to the target sequence length.

The results can be found in [this Comet project](https://www.comet.com/ai2/petew-testing/view/EOJ7jL0iY8uqXNXA0v0lyRwee/panels). In summary, both packing strategies performed significantly better than the concat+chunk baseline, with `pack-and-truncate` the overall winner.

image

Both packing strategies add a *surprisingly* small number of *padding tokens* to the training instances. On average, fewer than 1 padding token (0.1 - 0.2) is added per instance.  `pack-and-fragment` only  produced an additional 123 instances over the 666,182 instances produced from the default concat+chunk method (+0.012%). I found this counter-intuitive, as I thought there would be a lot more padding, but it agrees with the results in the paper. See Section 3.1 on **Campactness** and Table 2:

> The other important perspective of packing is how compact the resulting training sequences are. Theo- retically, both FFD and BFD are guaranteed to use no more than 11/9 of the optimal number of bins asymptotically (Johnson et al., 1974). However, practically, the majority of documents are short compared to context length, as shown by the dashed curves in Figure 4. The abundance of small- sized items makes packing much easier. Consequently, we observe that the training sequences from packing are nearly as compact as those obtained from concatenation. As shown in Table 2, Best-fit Packing yields no more than 0.01% addi- tional training sequences with either 2k or 8k context length and across NL and PL datasets. From another perspective, Table 2 also indicates that with Best-fit Packing, the amount of padding is negligibly small. The results prove that Best- fit Packing achieves roughly the same training efficiency as concatenation, as measured by the number of non-padding tokens processed using the same amount of compute.

image

---

**For PR reviewers, the main changes are:**

- In`src/olmo_core/data/numpy_dataset.py`: I moved some logic from `NumpyFSLDataset` to a new base class called `NumpyFSLDatasetBase`, and then added the new `NumpyPackedFSLDataset` which inherits from `NumpyFSLDatasetBase`. The `NumpyPackedFSLDataset` implementation is pretty light, as most of the heavy lifting is done in the `pack_documents_into_instances()` function in `src/olmo_core/data/utils.py`. The only other noteworthy change in this file is with how the dataset fingerprints are calculated. The fingerprint now includes the class name so that different subclasses will produce different fingerprints, as they should. This change is backwards compatible since I also changed the `fingerprint_version` (you'll see a warning when restarting a run from an older version, but that's okay).
- In`src/olmo_core/data/utils.py`: this is where I implemented the Optimized Best-Fit Decreasing (OBFD) bin-packing algorithm following the details in Appendix B of [the paper](https://arxiv.org/pdf/2404.10830). That section is quite easy to follow, and I recommend reading through it to see what's going on. I've added corresponding tests in `src/test/data/utils_test.py` that follow the exact example from the appendix.


src/olmo_core/data/utils.py
```diff
@@ -592,3 +651,186 @@ def find_periodic_sequences(
...
+
+@dataclass
+class SegmentTreeNode:
+    weight: int = 0
+    parent: Optional["SegmentTreeNode"] = None
+    children: Optional[Tuple["SegmentTreeNode", "SegmentTreeNode"]] = None
+    leaf_id: Optional[int] = None
+
+    @property
+    def is_root(self) -> bool:
```
2015aroras, line 670: nit: Don't set weight here, so that tree structure setup is a bit more decoupled from weight setup
```suggestion
        self.root_node = SegmentTreeNode()
```

src/olmo_core/data/utils.py
```diff
@@ -592,3 +651,186 @@ def find_periodic_sequences(
...
+
+
+class SegmentTree:
+    def __init__(self, N: int):
+        assert math.log2(N) % 1 == 0, "N should be a power of 2"
+        self.root_node = SegmentTreeNode(weight=N)
+        self.leaf_nodes: List[SegmentTreeNode] = []
+
+        max_depth = int(math.log2(N))
+        leaf_id = 0
```
2015aroras, line 695: thought: you could move this assert into the while loop as `assert weight <= node.weight`. Then you cover the root case and the general invariant that the weight should not exceed the current node (except when you hit the leaf)

src/olmo_core/data/utils.py
```diff
@@ -592,3 +651,186 @@ def find_periodic_sequences(
...
+
+    :param path: Path/URL to the source file of token IDs.
+    :param max_sequence_length: The maximum sequence length of each *instance*.
+    :param eos_token_id: The EOS token ID, used to find document boundaries.
+    :param dtype: The numpy datatype of the source file.
+    :param indices_dtype: The numpy datatype to use for document indices.
+    :param long_doc_strategy: Specifies how to handle document that are longer than ``max_sequence_length``.
+        If set to "truncate" then those documents are just truncated to ``max_sequence_length`` and
+        the excess tokens are discarded.
+        If set to "fragment" then those documents are split into smaller documents so that no tokens
```
2015aroras, line 826: nit: Sorting is a detail of the packing algorithm. The packer could instead have a `pack_documents()` method that does any desired sorting.

2015aroras: A `pack_documents` method would be less conducive of adding documents in an online fashion, which may or may not be something we want.

epwalsh:
update: I've got a full-blown 1B pretraining ablation going here https://www.comet.com/ai2/olmo-core-1b/view/3zFCkQjRaX50fvYTvuIeYXMiA/panels

2015aroras (approved):
LGTM! Just minor nits/thoughts

src/olmo_core/data/utils.py
```diff
@@ -592,3 +651,186 @@ def find_periodic_sequences(
...
+            self.weight = weight
+        else:
+            assert self.children is not None
+            self.weight = max(self.children[0].weight, self.children[1].weight)
+        if self.parent is not None:
+            self.parent.update()
+
+
+class SegmentTree:
+    def __init__(self, N: int):
```
2015aroras, line 689: thought: what about just something like `self.leaf_nodes[-1].update(N)`?

src/olmo_core/data/utils.py
epwalsh: Done: a7b18c2a

src/olmo_core/data/utils.py
epwalsh: Done: a7b18c2a

src/olmo_core/data/utils.py
epwalsh: > ... would be less conducive of adding documents in an online fashion, which may or may not be something we want.

The online packing variants perform much worse than the offline versions, so I'm not too keen on supporting that use case.

epwalsh: > The packer could instead have a pack_documents() method that does any desired sorting.

Done: 03f128fa

src/olmo_core/data/utils.py
epwalsh: nice; d18f3c04

src/olmo_core/data/utils.py
2015aroras: nit: If we're not supporting online, then maybe change `pack_document` to `_pack_document` to discourage its use.

Aside: I'm also in favor of not supporting online, since it's not really best-fit anymore.

2015aroras (approved):

src/olmo_core/data/utils.py
epwalsh: done; 5ea6d28f

2015aroras (approved):

PR #263, opened 2025-04-25
Add Interleaved Numpy Dataset

2015aroras:
This PR adds `NumpyInterleavedFSLDataset`, a numpy dataset that produces instances by splits source documents into chunks and interleaves these chunks.

The implementation inherits most of the functionality from `NumpyPaddedFSLDataset`. It modifies `NumpyPaddedFSLDataset` to produce documents of size `sequence_length // docs_per_instance` (where `docs_per_instance` is set to 1 in `NumpyPaddedFSLDataset`), and then updates `__getitem__` to get `docs_per_instance` of these smaller documents and interleave them. This approach is a bit hacky; an alternative was to put the interleaving functionality directly into `NumpyPaddedFSLDataset`, since `NumpyInterleavedFSLDataset` is a generalization of it.


src/olmo_core/data/numpy_dataset.py
```diff
@@ -1209,6 +1212,155 @@ def _pack_all_documents_into_instances(self):
...
+        )
+
+        # Add an eos token if there is space after interleaving.
+        if len(item["input_ids"]) < self.sequence_length:
+            item["input_ids"] = F.pad(item["input_ids"], (0, 1), value=self.eos_token_id)
+            item["label_mask"] = F.pad(item["label_mask"], (0, 1), value=True)
+
+        pad_shape = (0, self.sequence_length - len(item["input_ids"]))
+        item["input_ids"] = F.pad(item["input_ids"], pad_shape, value=self.pad_token_id)
+        item["label_mask"] = F.pad(item["label_mask"], pad_shape, value=False)
```
epwalsh: I'm guessing we would need to handle BOS tokens similarly?

src/olmo_core/data/numpy_dataset.py
```diff
@@ -1209,6 +1212,155 @@ def _pack_all_documents_into_instances(self):
...
+                for chunked_tensor in chunked_tensors
+            ]
+        )
+
+    def __getitem__(self, index: int) -> Dict[str, Any]:
+        index = int(index)  # in case this is a numpy int type.
+        pos_index = index if index >= 0 else len(self) + index
+
+        if self._docs_indices is None:
+            raise RuntimeError(f"{self.prepare.__name__} has not been called.")
```
epwalsh: Do you mean:
```suggestion
            raise RuntimeError(f"{self.__name__}.prepare() has not been called.")
```

src/olmo_core/data/numpy_dataset.py
```diff
@@ -1209,6 +1212,155 @@ def _pack_all_documents_into_instances(self):
...
+            self._num_instances = self.offsets[-1][1]
+        return self._num_instances // self._docs_per_instance
+
+    def prepare(self):
+        super().prepare()
+        self._docs_indices = (
+            get_rng(self._seed)
+            .permutation(len(self) * self._docs_per_instance)
+            .reshape(-1, self._docs_per_instance)
+        )
```
epwalsh: Could cache this as a memmap like we do with other files. That would save time with subsequent runs, since permuting a large array can take a couple minutes.

src/olmo_core/data/numpy_dataset.py
2015aroras: BOS token slipped my mind since we don't use it for any of the other datasets, will add.

2015aroras: Done https://github.com/allenai/OLMo-core/pull/263/commits/1878ac12dc1e568df839c7f160f1d2e229cfa484

src/olmo_core/data/numpy_dataset.py
2015aroras: Ended up removing this while caching the indices

src/olmo_core/data/numpy_dataset.py
2015aroras: Done https://github.com/allenai/OLMo-core/pull/263/commits/8ac35c8c41709d3400ad7ea3a1cfc965875e9054

epwalsh (approved):
LGTM!

PR #264, opened 2025-04-28
More fixes for `LMHead` with TP

epwalsh:
There were 2 more issues with the `LMHead` when TP is enabled:

1. An unexpected bug with the default loss implementation introduced by #261. I don't completely understand the issue, but it had to do with combining CE loss and Z loss as `DTensors` across the TP group within the `LMHead` module. When I changed that to combine them as local, non-distributed tensors, everything worked again. See [this beaker run](https://beaker.allen.ai/orgs/ai2/workspaces/long-contexts/work/01JSWWW8YSE524MRSFNQYNQTP3?taskId=01JSWWW8YZB1KE7NG5ZSPHXCDJ&jobId=01JSWWW93H0MW7NVHXJBJ7BCXQ).
2. An issue with the fused-linear loss implementation that existed prior to #261. The problem was that the `w_out` linear module wasn't treated with any TP wrapper, so FSDP would complain that the parameters of `w_out` were only placed across the DP dimension of the device mesh, while every other parameter was placed across both DP and TP dimensions. See [this beaker run](https://beaker.allen.ai/orgs/ai2/workspaces/long-contexts/work/01JSX066ZAMZEZCKH5499ZSFEV?taskId=01JSX066ZG5DPGKX0XXYBSCVSX&jobId=01JSX0672ZV9QBEWFPTPDTAXN1). This was an easy fix: https://github.com/allenai/OLMo-core/pull/264#discussion_r2064063805

### DIFF
--- a/src/olmo_core/nn/lm_head.py
+++ b/src/olmo_core/nn/lm_head.py
@@ -242,14 +242,6 @@ class LMHead(nn.Module):
                 loss = ce_loss + z_loss
             else:
                 loss = ce_loss
-            #  loss = ce_loss
-            #  ce_loss = self._finalize_loss(
-            #      ce_loss, B, loss_reduction=loss_reduction, loss_div_factor=loss_div_factor
-            #  )
-            #  if z_loss is not None:
-            #      z_loss = self._finalize_loss(
-            #          z_loss, B, loss_reduction=loss_reduction, loss_div_factor=loss_div_factor
-            #      )
         elif self.loss_implementation == LMLossImplementation.fused_linear:
             logits = None
             loss, z_loss = fused_linear_cross_entropy_loss(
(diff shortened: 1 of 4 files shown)


src/olmo_core/train/trainer.py
```diff
@@ -856,7 +856,7 @@ def record_metric(
         if not isinstance(value, torch.Tensor):
             value = torch.tensor(value)
         else:
-            value = get_local_tensor(value).float()
+            value = get_local_tensor(value.detach()).float()
```
epwalsh, line 859: This was unrelated to the bugs in the PR, but is good to have for extra safety.

src/olmo_core/nn/lm_head.py
```diff
@@ -350,7 +374,13 @@ def apply_tp(
...
             )
 
-        if self.loss_implementation != LMLossImplementation.fused_linear:
+        if self.loss_implementation == LMLossImplementation.fused_linear:
+            parallelize_module(
+                module=self.w_out,
+                device_mesh=tp_mesh,
+                parallelize_plan=SequenceParallel(),
+            )
+        else:
```
epwalsh, line 383: This is the one change needed to fix issue (2), everything else was to fix issue (1).

epwalsh (commented):
NOTE: I'm not updating the CHANGELOG because the lines I added in #261 are still relevant, and those changes haven't been released.

src/test/nn/lm_head_test.py
```diff
@@ -118,6 +118,14 @@ def run_lm_head_tp(
         z_loss_multiplier=z_loss_multiplier,
     )
     _, local_loss, local_ce_loss, local_z_loss = local_output
+
+    # The loss for optimizing ('local_loss') will have been reduced across the TP group, but not the losses
```
2015aroras, line 122: Does this break prior to issue (1) being fixed? Just wondering if our tests would catch this bug if it happened again.

epwalsh: This test was passing even with (1) being present, so it must be an interaction with FSDP + TP, and this test is only TP. It's on my TODO list to expand the tests.

PR #269, opened 2025-05-01
Add an example for using open_instruct data

vwxyzjn:
Hi @epwalsh and @soldni, here is an example to create fine-tuning data with the open instruct utility.

I am a bit unsure about several things:

1. how to integrate attention mask / labels in olmo-core
2. where you'd like to store this `data_transformation.py`

Instructions on how to use 

```bash
python src/examples/llama/oi_sft_data.py
torchrun --nproc-per-node=1 src/examples/llama/train_sft.py haha
```

image

### DIFF
--- a/src/oi_data/data_transformation.py
+++ b/src/oi_data/data_transformation.py
@@ -1,0 +1,1465 @@
+"""
+This script is taken from https://github.com/allenai/open-instruct/blob/main/open_instruct/dataset_transformation.py
+"""
+
+# 1. PPO (prompt)
+# 2. SFT (prompt + demonstration), there is also packing.
+# 3. ✅ RM / DPO (chosen and rejected)
+# 4. ✅ Visualization of length distributions?
+# 5. ✅ Filter?
+# 6. ✅ dataset_num_proc
+# 7. ✅ check EOS token
+# 8. dataset mixer?
+# 9. ✅ pretty print that show tokenization?
+# 10. ✅ hashable tokneization?
+# 11. inputs / labels / attention_mask
+# 12. ✅ always set a `tokenizer.pad_token_id`?
+# 13. a new DataCollatorForLanguageModeling?
+# 14. ✅ `add_bos_token` and `add_eos_token`? E.g., LLAMA models
+# 15. ✅ generate properties: has eos_token, bos_token (through chat template)
+
+# ✅ get tokenizer revision
+# ✅ get dataset revision
+# create a cached tokenized dataset, with tokenized revision, dataset revision, tokenization function name.
+
+# too many names related to "maximum length":
+# * `max_seq_length` in SFT
+# * `max_length`, `max_target_length` in RM / DPO,
+# * `max_prompt_length` in DPO
+
+# TODO: note that tokenizer doesn't change but model name does change. Should be mindful of this.
+"""
+This file contains the utility to transform and cache datasets with different configurations.
+The main things we are looking for are:
+* handle dataset mixing
+* handle different tokenization functions
+* **cache** the tokenized dataset so we don't have to re-tokenize every time
+    * This is especially important when we have 405B SFT models: 32 nodes are just spending like
+    5 minutes to tokenize the dataset. This translates to 32 * 5 * 8 = 1280 minutes = 21 hours of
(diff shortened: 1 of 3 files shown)


src/oi_data/data_transformation.py
epwalsh, line 1: I think this could live in `src/olmo_core/data/` as a submodule there, and then we just add `datasets` and `huggingface-hub` as additional optional dependencies (I think those are the only new ones) to `pyproject.toml`.

src/olmo_core/data/oi_data/data_transformation.py
```diff
@@ -0,0 +1,1531 @@
+# this file deals with dataset pre-processing before training
```
tyler-romero: note: exact copy from open-instruct

tyler-romero:
Script is pretty much ready for integration testing now!

Realistic command to launch SFT:
```
uv run python src/scripts/train/sft/OLMo2-7B-sft.py launch olmo2-7B-sft-take3-8gpu /weka/oe-training-default/ai2-llm/checkpoints/dustins/lc_7b_cont_pretrain_final_anneal/step11921 ai2/augusta-google-1 --trainer.callbacks.wandb.enabled=True --launch.num_gpus=8 --launch.priority=high
```

src/olmo_core/train/trainer.py
```diff
@@ -743,10 +743,19 @@ def load_checkpoint(self, dir: PathOrStr, *, load_trainer_state: Optional[bool]
...
+        error: Optional[Exception] = None
         if get_rank() == 0 and not self.checkpointer.dir_is_checkpoint(dir):
-            # Try to find the latest checkpoint in the directory.
-            dir = self.checkpointer.latest_checkpoint(dir)
-        dir = scatter_object(dir)
+            try:
+                dir_to_scatter = self.checkpointer.latest_checkpoint(dir)
+            except FileNotFoundError as e:  # defer raising until after the scatter
+                dir_to_scatter, error = None, e
+        dir_to_scatter = scatter_object(dir_to_scatter)
```
tyler-romero, line 765: This avoids a 30 min hang (rank 0 throws exception while other ranks are waiting) when the checkpoint isn't found

natolambert (approved):
Did a quick skim, generally looks fine, some TODOs and placeholders. how is the bos token before a sequence handled in olmo_core? One of the flags we need in open instruct is --add_bos for olmo's

src/olmo_core/train/trainer.py
epwalsh: Oh, this is a great fix!

src/scripts/train/sft/OLMo2-7B-sft.py
```diff
@@ -0,0 +1,390 @@
...
+            .with_callback("gpu_monitor", GPUMemoryMonitorCallback())
+            .with_callback("config_saver", ConfigSaverCallback())
+            .with_callback("garbage_collector", GarbageCollectorCallback())
+            .with_callback(
+                "checkpointer",
+                CheckpointerCallback(
+                    save_interval=1000, ephemeral_save_interval=500, save_async=True
+                ),
+            )
+            .with_callback(
```
epwalsh: Consider using `.with_recommended_evals()` instead.
https://github.com/allenai/OLMo-core/blob/796c2dc2b57985b160395b29855bc4320ac4eea1/src/olmo_core/train/config.py#L69

epwalsh (commented):
Overall looks pretty good, only a couple minor comments. Also, have we ablated document packing + intra-document masking in this scenario yet? I suspect it works well, but we should confirm if we haven't already.

tyler-romero:
@natolambert bos is supported in olmo-core, since your review I added better support for reading the same tokenizer setup that was used to tokenize the SFT data.

tyler-romero:
> Also, have we ablated document packing + intra-document masking in this scenario yet? I suspect it works well, but we should confirm if we haven't already.

Running that ablation overnight!

tyler-romero:
> how is the bos token before a sequence handled in olmo_core?

bos is supported explicitly, just need to ensure that the tokenizer that is specified is configured with the right token id.

PR #271, opened 2025-05-06
Sliding Window Attention

dirkgr:


src/olmo_core/nn/attention/__init__.py
```diff
@@ -137,6 +157,8 @@ def num_params(self, d_model: int) -> int:
     def build(
         self,
         d_model: int,
+        layer_idx: int,
+        n_layers: int,
```
epwalsh, line 162: Can we make these key-word only (put them after the `*`)? Would be easy to accidentally swap them.

epwalsh (approved):
LGTM. Only one minor suggestion

src/olmo_core/nn/attention/__init__.py
dirkgr: Done

PR #272, opened 2025-05-09
Add `BatchSizeSchedulerCallback`

epwalsh:
Implements the batch size scheduler defined by Dirk in [this document](https://docs.google.com/document/d/1D_LCmHdJcTAVsiLveDdik1F2VwuSlsnm3A9ulnNXQA8/edit?usp=sharing).

### DIFF
--- a/src/olmo_core/train/callbacks/batch_size_scheduler.py
+++ b/src/olmo_core/train/callbacks/batch_size_scheduler.py
@@ -2,7 +2,7 @@ import dataclasses
 import logging
 import math
 from dataclasses import dataclass
-from typing import List, Optional, Tuple
+from typing import List, Optional, Tuple, Union
 
 import torch
 
@@ -124,11 +124,17 @@ class BatchSizeSchedulerCallback(Callback):
             log.info(
                 f"Adjusting base learning rate by a factor of {lr_adjustment_factor:.4f} = sqrt({ratio})"
             )
-            for optim in optimizers:
-                for group in optim.param_groups:
+            for optim_idx, optim in enumerate(optimizers):
+                for group_idx, group in enumerate(optim.param_groups):
+                    new_lr: Union[float, torch.Tensor]
                     if scheduler is not None:
                         if group.get(scheduler.initial_lr_field) is None:
                             group[scheduler.initial_lr_field] = group[scheduler.lr_field]
                         group[scheduler.initial_lr_field] *= lr_adjustment_factor
+                        new_lr = group[scheduler.initial_lr_field]
                     else:
                         group["lr"] *= lr_adjustment_factor
+                        new_lr = group["lr"]
+                    log.info(
+                        f"Set base LR for optimizer {optim_idx+1}, group {group_idx+1} to {float(new_lr):.8f}"
+                    )
--- a/src/olmo_core/train/callbacks/batch_size_scheduler.py
+++ b/src/olmo_core/train/callbacks/batch_size_scheduler.py
@@ -83,7 +83,8 @@ class BatchSizeSchedulerCallback(Callback):
     def post_attach(self):
         self._maybe_update_batch_size_and_lr()
 
-    def post_checkpoint_loaded(self):
+    def post_checkpoint_loaded(self, *args):
+        del args
         self._maybe_update_batch_size_and_lr()
 
     def pre_load_batch(self):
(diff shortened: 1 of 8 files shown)


src/olmo_core/train/callbacks/batch_size_scheduler.py
```diff
@@ -0,0 +1,140 @@
...
+                elif target_batch_size < self.current_batch_size:
+                    assert self.current_batch_size % target_batch_size == 0
+                    self._update_batch_size_and_lr(target_batch_size)
+
+                break
+
+    def _update_batch_size_and_lr(self, batch_size: int):
+        log.info(f"Changing global batch size to {batch_size:,d}...")
+        ratio = batch_size / self.current_batch_size
+        lr_adjustment_factor = math.sqrt(ratio)
```
dirkgr, line 124: It's worth noting that this is true for Adam, but may not be true for other optimizers. I don't know how to package that fact in the code. Can we check somehow whether we have Adam, and throw an error if we don't?

epwalsh: 29a9ae02

epwalsh:
I have some short test runs here: https://www.comet.com/ai2/petew-test-bz-sched/view/new/panels

src/olmo_core/train/callbacks/batch_size_scheduler.py
```diff
@@ -0,0 +1,171 @@
...
+    def __post_init__(self):
+        if len(self.batch_sizes) != len(self.schedule):
+            raise OLMoConfigurationError(
+                "batch_sizes and schedules should have the same number of items"
+            )
+
+        if not self.schedule:
+            return
+
+        if len(set([duration.unit for duration in self.schedule])) > 1:
```
dirkgr: Python has set comprehensions too. `{x for x in something}`.

src/olmo_core/train/callbacks/batch_size_scheduler.py
```diff
@@ -0,0 +1,171 @@
...
+                f"Unable to adjust learning rate for {self.trainer.train_module.__class__.__name__} train module class"
+            )
+            return
+
+        for optim in optimizers:
+            if not isinstance(optim, (torch.optim.Adam, torch.optim.AdamW, SkipStepAdamW)):
+                log.warning(
+                    f"Unable to adjust learning rate for {optim.__class__.__name__} optimizer"
+                )
+                return
```
dirkgr: It falls back to doing nothing? I'd rather it fails. Otherwise we end up running 500B tokens worth of baselines, only to discover afterwards that something we configured didn't actually run.

src/olmo_core/train/callbacks/batch_size_scheduler.py
```diff
@@ -0,0 +1,171 @@
...
+
+                if group.get(initial_lr_field) is None:
+                    group[initial_lr_field] = group[lr_field]
+
+                group[initial_lr_field] *= lr_adjustment_factor
+
+                # Only set the actual LR if there's no scheduler. Schedulers update the LR based
+                # on the initial LR.
+                if scheduler is None:
+                    group[lr_field] = group[initial_lr_field]
```
dirkgr, line 165: LR schedulers don't overwrite this every step?

dirkgr (changes requested):
What LR will we log? The one from the LR scheduler, or the overall one?

src/olmo_core/train/callbacks/batch_size_scheduler.py
epwalsh: They overwrite "lr" but not "initial_lr". So we only touch "lr" when there's no scheduler.

src/olmo_core/train/callbacks/batch_size_scheduler.py
epwalsh: done: ff895e8f

src/olmo_core/train/callbacks/batch_size_scheduler.py
epwalsh: done: ff895e8f

epwalsh:
> What LR will we log? The one from the LR scheduler, or the overall one?

We always log (to comet/W&B) the LR that's actually used in the optimizer step.

dirkgr (approved):

PR #273, opened 2025-05-11
make indices filename robust to change in batch size

epwalsh:


src/olmo_core/data/data_loader.py
```diff
@@ -631,7 +631,7 @@ def _global_indices_file(self) -> Path:
             "global_indices",
             seed=self.seed if self.shuffle else None,
             epoch=self.epoch if self.shuffle else None,
-            size=self.total_size,
+            dataset_size=len(self.dataset),
```
epwalsh, line 634: Changing from `self.total_size` to just `len(self.dataset)`, since `self.total_size` could change with batch size if the remainder of `len(dataset) / batch size` changes.

dirkgr (approved):

PR #277, opened 2025-05-14
Support lower precisions for conversion to HF

2015aroras:
This PR adds support for converting models to HF in lower precisions. Due to https://github.com/allenai/olmo-cookbook/issues/60, it also changes the default HF precision to bf16 (so that vllm is less likely to be broken).


2015aroras:
@soldni Since we know that HF doesn't match OLMo in bf16 currently, should we instead stick with fp32 default for now?

epwalsh (approved):
LGTM

PR #281, opened 2025-05-19
Fast in-loop MCQA

davidheineman:
Bump in-loop evals to v0.8.1. This will add "fast" MCQA, which performs MC tasks in 1 forward pass instead of 4 forward passes. (We extract the A/B/C/D logits from a single pass).

https://github.com/allenai/OLMo-in-loop-evals/pull/8

This will make the MC tasks 4x faster, and produces the same numbers.

Also, added Java, Rust and C++ translated MBPP BPB.


epwalsh (approved):
Sweet!

PR #286, opened 2025-05-28
Update default cluster for src/examples/llama/train_launch.py

tyler-romero:
Update cluster to avoid error when running `src/examples/llama/train_launch.py` using defaults.


epwalsh (approved):
LGTM

PR #290, opened 2025-06-06
Make ruff line-length match black line-length

tyler-romero:
```
> make checks
======== running isort... ========
Skipped 2 files
======== running black... ========
All done! ✨ 🍰 ✨
217 files would be left unchanged.
======== running ruff... =========
All checks passed!
======== running mypy... =========
Success: no issues found in 216 source files
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ dynamic = ["version"]
 readme = "README.md"
 description = "Core training module for the Open Language Model (OLMo)"
 authors = [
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -107,7 +96,7 @@ profile = "black"
 multi_line_output = 3
 
 [tool.ruff]
-line-length = 115
+line-length = 100
```
tyler-romero, line 99: My editor would just reformat every file b/c this doenst match black's config above, which is what the makefile uses

epwalsh (approved):

PR #295, opened 2025-06-11
Bump pytorch, ring-flash-attn, and liger-kernel versions

tyler-romero:
Based on the changelogs these are mostly under the hood bugfixes.

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -43,7 +43,7 @@ RUN chmod +x ~/miniconda.sh && \
 
 # Install PyTorch core ecosystem.
 ARG CUDA_VERSION_PATH
-ARG TORCH_VERSION=2.7.0
+ARG TORCH_VERSION=2.7.1
 ARG INSTALL_CHANNEL=whl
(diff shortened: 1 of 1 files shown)


src/Dockerfile
epwalsh, line 1: You'll have to update these values in the `Makefile` as well

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed bug causing slow evals in BPB/RC in-loop evals due to fast MC
 - Changed default precision of converted HF models in `src/examples/huggingface/convert_checkpoint_to_hf.py` to bfloat16.
 - Changed default cluster to `saturn` in `src/examples/llama/train_launch.py`.
+- Bump pytorch, ring-flash-attn, and liger-kernel versions
```
epwalsh: This kind of change is okay to omit from the CHANGELOG, since it's not an API change and we're not changing any version pins in `pyproject.toml`.

src/Dockerfile
tyler-romero: Ah whoops, totally missed those

epwalsh (approved):

PR #298, opened 2025-06-16
Allow profiler to record on multiple ranks

tyler-romero:
Useful in conjunction with https://hta.readthedocs.io/en/latest/ for profiling distributed training runs.

For example can be useful for analyzing skew across ranks:
image

image

### DIFF
--- a/src/olmo_core/train/callbacks/profiler.py
+++ b/src/olmo_core/train/callbacks/profiler.py
@@ -52,21 +52,25 @@ class ProfilerCallback(Callback):
     """
     Whether to track tensor memory allocation/deallocation
     """
+    enable_cuda_sync_events: bool = False
+    """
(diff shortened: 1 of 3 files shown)


src/olmo_core/train/callbacks/profiler.py
```diff
@@ -62,21 +143,26 @@ def pre_train(self):
         if self.trainer.device.type == "cuda":
             activities.append(ProfilerActivity.CUDA)
 
+        experimental_config = None
+        if self.enable_cuda_sync_events:
+            experimental_config = _ExperimentalConfig(enable_cuda_sync_events=True)
```
tyler-romero, line 149: Useful for critical path analysis:
https://hta.readthedocs.io/en/latest/source/features/lightweight_critical_path_analysis.html

epwalsh (approved):
Excited to try out HTA at some point 😎

PR #301, opened 2025-06-20
Improve clarity of SWA config

tyler-romero:
Changes to SWA config to improve clarity. Open to other suggestions.

### DIFF
--- a/src/olmo_core/nn/attention/__init__.py
+++ b/src/olmo_core/nn/attention/__init__.py
@@ -58,39 +58,61 @@ class SlidingWindowAttentionConfig(Config):
     pattern: List[int]
     """
     The pattern of window sizes to use for attention, repeated to cover all layers.
-    -1 indicates full attention. Example: [4096, 4096, 4096, -1] means that for each set of 4 layers,
-    the first 3 layers will use a window size of 4096, and the last layer will use full attention.
+    A value of -1 indicates full attention. For example, a pattern of `[4096, 4096, 4096, -1]`
+    means that for each set of 4 layers, the first 3 will use a window size of 4096,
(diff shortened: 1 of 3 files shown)


dirkgr:
Haven't read closely yet. What about backwards compatibility?

tyler-romero:
There are no configs in main that use this class, so BW compat can be addressed when merging this change into the Olmo3 branch.

src/olmo_core/nn/attention/__init__.py
```diff
@@ -55,21 +55,60 @@
...
 @dataclass
 class SlidingWindowAttentionConfig(Config):
-    pattern: List[bool]
-    window_size: int = 4096
-    force_first: bool = True
-    force_last: bool = True
+    pattern: List[int]
+    """
+    The pattern of window sizes to use for attention, repeated to cover all layers.
+    A value of -1 indicates full attention. For example, a pattern of `[4096, 4096, 4096, -1]`
```
epwalsh: Our docs builder parses doc strings as reStructuredText (RST), which requires double backticks for inline code.

```suggestion
    A value of -1 indicates full attention. For example, a pattern of ``[4096, 4096, 4096, -1]``
```

epwalsh (commented):
LGTM other than my comment on docstring syntax 👍

epwalsh:
> Haven't read closely yet. What about backwards compatibility?
>> There are no configs in main that use this class, so BW compat can be addressed when merging this change into the Olmo3 branch.

Adding to that, we haven't published a release with SWA yet, so there's no concern from me with making breaking changes here.

epwalsh (approved):

PR #302, opened 2025-06-25
Only check for beaker secrets in non-distributed settings

2015aroras:
Checking for beaker secrets in distributed settings can result in a lot of requests to beaker. This DOS'd beaker causing it to raise 5XX errors. This PR only checks for beaker secrets in non-distributed settings.

### DIFF
--- a/src/olmo_core/internal/common.py
+++ b/src/olmo_core/internal/common.py
@@ -5,6 +5,7 @@ from typing import List, Optional
 import torch
 from beaker import Beaker, BeakerError, SecretNotFound
 
+from olmo_core.distributed.utils import is_distributed
 from olmo_core.exceptions import OLMoConfigurationError
 from olmo_core.io import is_url
 from olmo_core.launch.beaker import (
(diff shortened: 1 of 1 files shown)


src/olmo_core/internal/common.py
```diff
@@ -53,7 +54,9 @@ def beaker_secret_exists(secret: str, workspace: Optional[str] = None) -> bool:
 def _to_beaker_env_secret(
     name: str, secret: str, *, workspace: Optional[str] = None, required: bool = True
 ) -> Optional[BeakerEnvSecret]:
-    if beaker_secret_exists(secret, workspace=workspace):
+    # Assume beaker secret exists if we are in a distributed setting (e.g., during a training job)
+    # so that we don't DOS beaker.
+    if not is_distributed() or beaker_secret_exists(secret, workspace=workspace):
```
epwalsh: Shouldn't this be:
```suggestion
    if is_distributed() or beaker_secret_exists(secret, workspace=workspace):
```

2015aroras: Yep, good catch! I need more sleep.

PR #303, opened 2025-06-25
Fix fingerprints for various FSLDatasets

tyler-romero:
Various options that are used to generate cached files aren't correctly included in the fingerprint. These are soon-to-be-relevant for SFT training.

### DIFF
--- a/src/olmo_core/data/numpy_dataset.py
+++ b/src/olmo_core/data/numpy_dataset.py
@@ -867,6 +867,18 @@ class NumpyPaddedFSLDataset(NumpyFSLDataset):
         self._array_instance_offsets: Optional[Tuple[Tuple[int, int], ...]] = None
 
     @property
+    def fingerprint_fields(self) -> Tuple[str, ...]:
+        return (
+            "vocab_size",
+            "pad_token_id",
+            "eos_token_id",
+            "dtype",
+            "max_target_sequence_length",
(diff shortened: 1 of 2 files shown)


src/olmo_core/data/numpy_dataset.py
```diff
@@ -866,6 +866,18 @@ def __init__(
...
+    @property
+    def fingerprint_fields(self) -> Tuple[str, ...]:
+        return (
+            "vocab_size",
+            "pad_token_id",
+            "eos_token_id",
+            "dtype",
+            "max_target_sequence_length",
+            "bos_token_id",
+            "sequence_length",
```
tyler-romero, line 878: sequence_length is used when generating the cached instance files. The others are carried over from the parent class.

src/olmo_core/data/numpy_dataset.py
```diff
@@ -1305,6 +1318,9 @@ def fingerprint_fields(self) -> Tuple[str, ...]:
             "_docs_per_instance",
             "_seed",
             "_interleaving_exempt_paths",
+            "max_target_sequence_length",
+            "bos_token_id",
+            "sequence_length",
```
tyler-romero, line 1323: carry over fingerprints required for the parent class

epwalsh (approved):
LGTM

PR #305, opened 2025-06-25
Make async bookkeeping more robust

epwalsh:
Adds a queue for tracking bookkeeping operations and the ability to cancel in-progress operations when submitting a new one.

### DIFF
--- a/src/olmo_core/train/callbacks/wandb.py
+++ b/src/olmo_core/train/callbacks/wandb.py
@@ -130,7 +130,7 @@ class WandBCallback(Callback):
     def post_step(self):
         cancel_check_interval = self.cancel_check_interval or self.trainer.cancel_check_interval
         if self.enabled and get_rank() == 0 and self.step % cancel_check_interval == 0:
-            self.trainer.thread_pool.submit(self.check_if_canceled)
+            self.trainer.run_bookkeeping_op(self.check_if_canceled, cancel_in_progress=True)
 
     def post_train(self):
         if self.enabled and get_rank() == 0 and self.run is not None:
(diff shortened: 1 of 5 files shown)


src/olmo_core/train/callbacks/wandb.py
```diff
@@ -151,11 +151,11 @@ def check_if_canceled(self):
             try:
                 # NOTE: need to re-initialize the API client every time, otherwise
                 # I guess it return cached run data.
-                api = self.wandb.Api(api_key=os.environ[WANDB_API_KEY_ENV_VAR])
+                api = self.wandb.Api(api_key=os.environ[WANDB_API_KEY_ENV_VAR], timeout=5)
```
dirkgr, line 154: Unrelated change?

src/olmo_core/train/trainer.py
```diff
@@ -1015,26 +1020,52 @@ def _handle_os_signal(self, signalnum, stack_frame):
...
+        if op_name is None:
+            op_name = op.__name__
+        op_id = uuid.uuid4().hex
+
         if (
             self.async_bookkeeping
             and self.bookkeeping_device.type == "cpu"
             and self.bookkeeping_pg is not None
         ):
+            if cancel_in_progress:
```
2015aroras, line 1059: Just checking, does any of this stuff affect the async checkpointing?

epwalsh: No it shouldn't because we don't use `cancel_in_progress` for that.

src/olmo_core/train/callbacks/wandb.py
epwalsh: Sort of related. Good to have a timeout on this.

PR #306, opened 2025-06-26
Uv for faster startup

tyler-romero:
I was working on this to help `gantry` experiments startup faster and figured I might as well roll it in here too.


.github/workflows/docker.yml
```diff
@@ -7,15 +7,15 @@ concurrency:
...
-  # pull_request:
-  #   branches:
-  #     - main
-  #   paths:
-  #     - 'Makefile'
-  #     - 'pyproject.toml'
-  #     - 'src/olmo_core/version.py'
-  #     - 'src/Dockerfile'
-  #     - '.github/workflows/docker.yml'
+  pull_request:
```
tyler-romero, line 10: todo: revert

PR #307, opened 2025-06-26
ensure sharded parameters initialized different on different ranks

epwalsh:


src/olmo_core/nn/transformer/model.py
```diff
@@ -232,9 +229,7 @@ def init_weights(
             if hasattr(module, "reset_parameters"):
                 module.reset_parameters()  # type: ignore
 
-        seed = self.init_seed
-        if pp_mesh is not None:
-            seed += pp_mesh.get_local_rank()
+        seed = self.init_seed + get_rank()
```
dirkgr, line 232: Heh, that's a way around that.

This system means that all of our initialization schemes always need to run one parameter at a time. We can't have dependencies between parameters at initialization time. But so far, this has always been true.

dirkgr (approved):

PR #313, opened 2025-07-01
"Better" sorting of Augusta ranks at runtime

2015aroras:
Issue: Augusta is made up of 2 blocks of nodes. When a model replica in a distributed job is split across the blocks, perf goes bad. We reorder ranks to avoid this happening, but the current logic has a 1 in N chance (for replica size N) of letting this happen even in the ideal case where there is an ordering that can avoid this.

This PR uses a slightly convoluted sorting that ensures model replicas in a distributed job are not split across the blocks if it's possible. This reduces our block splitting problem to ensuring that the number of nodes from each block is a multiple of the replica size N (separate PR eventually).

### DIFF
--- a/src/olmo_core/launch/reorder_ranks_in_gcp.py
+++ b/src/olmo_core/launch/reorder_ranks_in_gcp.py
@@ -13,6 +13,11 @@ def main():
     parser.add_argument("master_addr", help="Hostname of worker 0")
     parser.add_argument("--master_port", type=int, default=29501, help="Port for TCPStore")
     parser.add_argument("--debug", action="store_true", help="Enable debug mode (outside of GCP)")
+    parser.add_argument(
+        "--verbose",
+        action="store_true",
+        help="Print the whole list of node ids in order on rank 0 to stderr",
+    )
     args = parser.parse_args()
 
     # Create or connect to the store
--- a/src/olmo_core/launch/reorder_ranks_in_gcp.py
+++ b/src/olmo_core/launch/reorder_ranks_in_gcp.py
@@ -86,7 +86,7 @@ def main():
             all_blocks[rank],
             all_subblocks[rank],
             all_machines[rank],
-            rank
+            rank,
         )
     )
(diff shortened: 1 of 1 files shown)


epwalsh:
How are subblocks defined?

dirkgr:
Sub-blocks aren't a think. Maybe we could consider one switch's worth of nodes a sub-block, but we don't care much about that.

For the purposes of this PR, blocks are defined by the first member of the physical machine ID. We never want to shard a model across two nodes that don't share a first member of the physical machine ID.

dirkgr:
>Maybe we could consider one switch's worth of nodes a sub-block

Oh, I see that the code does exactly this.

2015aroras:
> How are subblocks defined?

@epwalsh Honestly no idea. Empirically, subblocks are generally used by at most 2 hosts (but sometimes 3), and they capture a notion of closeness that is finer grained than blocks.

The main thing we need is not to go across blocks. Having subblocks/machines not split across replicas is a bonus.

2015aroras:
This is what gcloud cli has to say. So subblocks are real. But what I call a "machine" they call a "host"
```
{
  "resourceStatus": {
    "physicalHost": "/53859ce2421d144b908e7a100ec51244/78c3e570133aa388d88305dd4706a89b/a7ef1744dc23b29b3a6b1c532e527765",
    "physicalHostTopology": {
      "block": "53859ce2421d144b908e7a100ec51244",
      "cluster": "us-central1-cluster-mjho",
      "host": "a7ef1744dc23b29b3a6b1c532e527765",
      "subblock": "78c3e570133aa388d88305dd4706a89b"
    }
  }
}
```

2015aroras:
Lol the auto-merge...

epwalsh:
Oh, well don't bother reverting unless @dirkgr wants changes.

dirkgr:
I'm good with this. I sat next to Shane when he wrote it.

PR #316, opened 2025-07-03
fix rounding error with mixing datasets

mayeechen:
When handling a mixing dataset (i.e., `SourceMixtureDataset`), there is a rounding error in converting target ratios into number of tokens per path, into number of instances per path. The precise issue:

> tokens_requested_per_path = target_ratio * num_tokens_in_path
> instances_requested_per_path = tokens_requested_per_path // sequence_length 
> sum(instances_requested_per_path) / batch_size != training_steps

where `batch_size` is at the instance level (i.e. `global_batch_size / sequence_length`) and `training_steps = math.ceil(max_tokens / global_batch_size)`.  

The consequence of this: Cookbook trains for `max_tokens` duration on the dataset, but the dataset has fewer than `max_tokens` due to this rounding error. Training thus goes 5-30 steps into epoch 2. We didn't notice this error on pre-training experiments, but it becomes obvious on mid-training anneals, as the training loss suddenly drops. 

The fix: request more tokens for each path:
> tokens_requested_per_path = target_ratio x num_tokens_in_path + padding, 

where padding is the smallest constant number such that 
> sum(instances_requested_per_path) / batch_size == training_steps

This way, we handle the rounding error evenly throughout the dataset; the padding approach roughly preserves the target ratios. 

Example runs:
Before fix
https://wandb.ai/ai2-llm/olmo-cookbook/runs/jeylqj50

After fix 
https://wandb.ai/ai2-llm/olmo-cookbook/runs/nbdfadys

Note: this change does not impact non-mixing datasets.

### DIFF
--- a/src/olmo_core/data/source_mixture.py
+++ b/src/olmo_core/data/source_mixture.py
@@ -180,6 +180,10 @@ class SourceMixtureDatasetConfig(Config):
     """
     The data type of the dataset.
     """
+    global_batch_size: int
+    """
+    The global batch size for training, in tokens.
+    """
     processes: int = 1
     """
     The number of processes to use for counting tokens in parallel.
@@ -259,7 +263,7 @@ class SourceMixtureDatasetConfig(Config):
 
         # Increase the number of tokens per path until we have enough instances, handling rounding issues 
         all_tokens_per_path = [path.tokens for source_path_tokens in tokens_per_path_per_source.values() for path in source_path_tokens]
-        requested_instances = self.max_tokens // self.sequence_length
+        requested_instances = math.ceil(self.max_tokens / self.global_batch_size) * int(self.global_batch_size / self.sequence_length)
         padding = 0
         while True:
             if sum([(tokens_per_path + padding) // self.sequence_length for tokens_per_path in all_tokens_per_path]) >= requested_instances:
--- a/src/olmo_core/data/source_mixture.py
+++ b/src/olmo_core/data/source_mixture.py
@@ -249,14 +249,35 @@ class SourceMixtureDatasetConfig(Config):
             )
 
         completed: List[SourceMixtureOutcome] = []
+        tokens_per_path_per_source: Dict[str, List[SourcePathTokens]] = {}
         for source in tokens_details_by_source:
+            source_path_tokens = self.get_paths_and_tokens_for_source(
+                        source_config=source.config,
+                        token_details=source,
+            )
+            tokens_per_path_per_source[source.config.source_name] = source_path_tokens
+
+        # Increase the number of tokens per path until we have enough instances, handling rounding issues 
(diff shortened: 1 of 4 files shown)


src/olmo_core/data/source_mixture.py
```diff
@@ -249,14 +253,49 @@ def build(self) -> SourceMixtureDataset:
...
+            for path in source_path_tokens
+        ]
+        requested_instances = math.ceil(self.max_tokens / self.global_batch_size) * int(
+            self.global_batch_size / self.sequence_length
+        )
+        padding = 0
+        while True:
+            if (
+                sum(
+                    [
```
tyler-romero: It seems like this solution makes the assumption that the paths are all close to the same size, so that adding an equal amount of padding to every path doesn't change the ratios.

Thats probably a reasonable assumption to make (I haven't checked) but I think it could lead to a subtle bug that biases the source mixture if that isn't true for some reason in the future - like if one source is generated with larger file sizes for some reason, or of one source fails to fill even a single file.

It should be possible to determine the number of tokens to add per-source to preserve the ratio and then div those tokens up within the source's paths.

src/olmo_core/data/source_mixture.py
```diff
@@ -180,6 +180,10 @@ class SourceMixtureDatasetConfig(Config):
     """
     The data type of the dataset.
     """
+    global_batch_size: int
```
tyler-romero, line 183: nit: It would be nice to not have to define the global_batch_size at two different places in the config (here and in the data loader). But I took a quick look at how that could be done and it doesn't look super straightforward/clean to communicate this value from the dataloader to here.

src/olmo_core/data/source_mixture.py
mayeechen: working on doing a different adjustment per source right now! I agree constant padding could bias things.

mayeechen: fixed and updated top level comment - we adjust the number of requested instances for paths that are already close to being rounded up, which results in minimal distortion. I also added some logging to display the difference in the final mixing ratios and the original target ratios, so we can always double check that the change is not too big.

src/olmo_core/data/source_mixture.py
mayeechen: agreed it's not clear to me how to avoid this redundancy..

dirkgr:
What's the status of this?

mayeechen:
> What's the status of this?

I finished implementing a solution to handle the rounding error with minimal distortion to the mixing ratios. The example runs above suggest that it is working as expected.

However, I think there hasn't been a strong push to get this integrated. While the issue (going a few steps into epoch 2 on every run that uses a mix) does occur consistently, its impact on the training loss drop seems to be limited to runs with a very small amount of tokens, based on my experience.

PR #317, opened 2025-07-03
Add config option for to enable/disable the step-increment bugfix.

tyler-romero:
Defaults to `True` which means *fix the bug*. Flip to `False` to reintroduce the bug.

### DIFF
--- a/src/olmo_core/optim/adamw.py
+++ b/src/olmo_core/optim/adamw.py
@@ -20,6 +20,7 @@ def adamw_step(
     exp_avg_sq: torch.Tensor,
     step: torch.Tensor,
     step_factor: torch.Tensor,
+    stepfix: bool = False,
 ):
     if p.grad is None:
         return
@@ -44,7 +45,8 @@ def adamw_step(
     update = -step_size * torch.div(exp_avg, denom)
     update.mul_(step_factor)
     p.add_(update)
-    step.add_(step_factor)
+    if stepfix:
+        step.add_(step_factor)
 
 
 def foreach_adamw_step(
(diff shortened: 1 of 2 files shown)


src/olmo_core/optim/adamw.py
```diff
@@ -259,6 +267,12 @@ class SkipStepAdamWConfig(OptimConfig):
     Faster than the non-foreach version.
     """
 
+    stepfix: bool = True
+    """
+    Whether or not to fix the step-incrementing bug discovered in SkipStepAdamW.
+    TODO: remove this when the bug is no longer being depended on.
```
2015aroras: This might have to stay forever for backwards incompatibility

src/olmo_core/optim/adamw.py
```diff
@@ -20,6 +20,7 @@ def adamw_step(
     exp_avg_sq: torch.Tensor,
     step: torch.Tensor,
     step_factor: torch.Tensor,
+    stepfix: bool = True,
```
2015aroras: nit: it might be better to name this for what it does (something like `increment_step`) rather than why (`stepfix`). You can explain why in the config comment

dirkgr: Name it `bias_correction`, and it should default to `True`?

tyler-romero: Thats a bit misleading bc its bias correction + lr * 2.2

tyler-romero:
```
    step_increment_bugfix: bool = True
    """
    Whether or not to fix the step-incrementing bug discovered in SkipStepAdamW.

    If this flag is set to False, the step will not be incremented, which
    gives the optimizer an effective lr that is 2.2x higher than the specified lr,
    and no bias correction is applied.
    """
 ```

dirkgr: That's true. I like yours better.

PR #319, opened 2025-07-07
fix initialization (again)

epwalsh:
# Background
While fixing one thing, #307 introduced a different issue with HSDP. Unlike DDP, PyTorch's `fully_shard()` (FSDP) wrapper doesn't synchronize to make sure parameters are consistent across the process group. So when you init your rank-local tensors with different seeds, you end up with inconsistent parameters across replica groups. That means replica 1 has completely different parameters than replica 2, and so on. When we restarted the bad run, this actually helped it out because now the replicas all ended up with the same parameters (the params from the first replica, because that's what was saved to the checkpoint).

# Changes
This PR reverts #307 more or less, and introduces a different way of initializing DTensors in general such that:
1. We don't run into this bug with HSDP.
2. We get the same full tensors after initialization regardless of world size.

# Integration test
I tested this end-to-end with the exact same OLMo3-1B settings where we observed the issue in the first place. You can find the runs here named `OLMo3-1B-swafix-initfix`: https://wandb.ai/ai2-llm/olmo3/workspace?nw=2kjfyryx4wm

image

Interestingly we also see lower peak GPU memory usage with this method.

image

### DIFF
--- a/src/olmo_core/nn/transformer/init.py
+++ b/src/olmo_core/nn/transformer/init.py
@@ -2,12 +2,27 @@ from typing import Optional, Union, cast
 
 import torch
 import torch.nn as nn
+from torch.distributed.tensor import DTensor
 
 from olmo_core.config import StrEnum
+from olmo_core.distributed.utils import distribute_like, get_local_tensor
 
 from ..attention import Attention, AttentionBase, FusedAttention
 from ..feed_forward import FeedForward
 from ..moe import DroplessMoEMLP, MoEBase, MoELinearRouter, MoEMLP
+
+
+def _apply_init(init_fun, x: torch.Tensor, *args, **kwargs):
+    if not isinstance(x, DTensor):
+        init_fun(x, *args, **kwargs)
+
+    # Initialize full version of x locally, then apply init to that.
+    full_x = torch.zeros(x.shape, dtype=x.dtype, device=x.device)
+    init_fun(full_x, *args, **kwargs)
+    full_x = distribute_like(x, full_x)
+
+    # Now copy over the corresponding shard of `full_x` into `x`.
+    get_local_tensor(x).copy_(get_local_tensor(full_x))
 
 
 class InitMethod(StrEnum):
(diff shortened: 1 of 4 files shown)


tyler-romero (approved):
This makes sense to me as a fix for this specific issue - but I wonder if we're papering over the core issue, which is that we have different seeds across groups that are supposed to be replicas of one another. I worry that the same issue will bite us again elsewhere in the future.

dirkgr:
This also fixes the issue that different world sizes produce different models at init time?

dirkgr:
Tests?

epwalsh:
> I wonder if we're papering over the core issue, which is that we have different seeds across groups that are supposed to be replicas of one another. I worry that the same issue will bite us again elsewhere in the future.

@tyler-romero I'm not sure what you mean. With these changes we use the same seed everywhere (except for different pipeline parallel stages).

> This also fixes the issue that different world sizes produce different models at init time?

Right.

> Tests?

A regression test here would need 4 GPUs (for HSDP), but we only run on up to 2 GPUs. `fully_shard()` won't work on CPU.

epwalsh:
Ok I added a test. As I said before this currently won't run in CI since it requires 4 GPUs, but you can run it manually like this:

```bash
python src/scripts/beaker/launch_test.py --num_gpus=4 -- pytest -v src/test/nn/transformer/model_test.py -k 'init_with_hsdp'
```

tyler-romero:
With these changes we're not actually correcting the seeds though right? We're just taking the model weights initialized on rank 0 and distributing them correctly. I agree this fixes the known issue but its not actually correcting the seeds.

epwalsh:
> With these changes we're not actually correcting the seeds though right?

We are fixing the init seeds. And the global seeds have always been fixed across ranks.
https://github.com/allenai/OLMo-core/blob/0dda3ec7592593df1ddc9830ee5dc05f57fb4132/src/olmo_core/internal/experiment.py#L304

> We're just taking the model weights initialized on rank 0 and distributing them correctly.

No each rank initializes the same weights locally (since the seeds are the same). There's no communication.

tyler-romero:
Ah I see, I was missing a couple details. Thanks!

PR #322, opened 2025-07-09
Add FlexAttention support

2015aroras:
This PR adds the option to use FlexAttention in our attention. This includes sliding window support and intra-document masking. FlexAttention can run on CPU, so now these aspects can be used locally for testing & HF conversion (the motivation for doing this). I have also added some SDPA tests for checking that sliding window support and intra-document masking work as expected.

Based on small test runs, FlexAttention on and off give similar training curves. FlexAttention decreases throughput by 10%, which is expected. 

Test Runs (no sliding window or intra-document masking):
[FlexAttention off](https://www.comet.com/ai2/shanea-testing/8eff0ed9406c40f6a268eebce3608e93?compareXAxis=step&experiment-tab=panels&prevPath=%2Fai2%2Fshanea-testing%2Fview%2Fnew%2Fpanels&showOutliers=true&smoothing=0&xAxis=step)
[FlexAttention on](https://www.comet.com/ai2/shanea-testing/1420f217b66d4de48291d2ed9cb9fed5?compareXAxis=step&experiment-tab=panels&prevPath=%2Fai2%2Fshanea-testing%2Fview%2Fnew%2Fpanels&showOutliers=true&smoothing=0&xAxis=step)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added support for revisions in `convert_checkpoint_from_hf.py` and the `load_hf_model` method of `olmo_core.nn.hf.checkpoint`.
 - `foreach` support in `SkipStepAdamW`.
 - Added `budget` mode for activation checkpointing configuration.
+- Added FlexAttention support https://pytorch.org/blog/flexattention/.
 
 ### Changed
 


src/olmo_core/nn/attention/__init__.py
```diff
@@ -437,6 +455,53 @@ def sdpa(
...
+            #        (batch_size, n_kv_heads, seq_len, head_dim)
+            q, k, v = q.transpose(1, 2), k.transpose(1, 2), v.transpose(1, 2)
+
+            # SDPA uses full precision. We match it for flex attention.
+            og_dtype = q.dtype
+            q, k, v = q.float(), k.float(), v.float()
+            with torch.autocast(enabled=False, device_type=q.device.type):
+                # shape: (batch_size, n_heads, seq_len, head_dim)
+                flex_att = flex_attention(
+                    q, k, v, block_mask=block_mask, scale=scale, enable_gqa=True
```
tyler-romero, line 498: is there a reason to set enable_gqa to False if we're actually not using it?

src/olmo_core/nn/attention/__init__.py
```diff
@@ -929,3 +1004,100 @@ def repeat_kv(x: torch.Tensor, n_rep: int) -> torch.Tensor:
...
+
+        mask_mods.append(_sliding_window_mask_mod)
+
+    if doc_lens is not None:
+        if device is None:
+            raise ValueError("Device is required for intra-document masking mod")
+
+        document_ids = torch.cat(
+            [torch.full((int(doc_len),), i, device=device) for i, doc_len in enumerate(doc_lens)]
+        )
```
tyler-romero, line 1041:
```suggestion
        document_ids = torch.repeat_interleave(
	        torch.arange(len(doc_lens), device=device),
	        torch.as_tensor(doc_lens, device=device, dtype=torch.uint32)
	    )
```
nit: I think this would be faster in the case where torch.compile isnt enabled

src/olmo_core/nn/transformer/model.py
```diff
@@ -206,6 +208,36 @@ def get_rope_buffers(
...
+
+        block_masks_by_window_size = {}
+        block_masks = []
+        for block in self.blocks.values():
+            block = cast(TransformerBlock, block)
+            att = cast(Union[Attention, FusedAttention], block.attention)
+
+            block_mask = None
+            if att.use_flex_attn:
+                # Reuse block masks across layers with the same window size
```
tyler-romero, line 228: nice

src/olmo_core/nn/transformer/model.py
```diff
@@ -1000,3 +1043,20 @@ def _unhide_cpu_inputs_from_torch(m, args, kwargs) -> Optional[Tuple[Any, Dict[s
...
+    if block_mask.full_kv_indices is not None:
+        mark_dynamic(block_mask.full_kv_indices, (2, 3))
+    if block_mask.q_num_blocks is not None:
+        mark_dynamic(block_mask.q_num_blocks, 2)
+    if block_mask.q_indices is not None:
+        mark_dynamic(block_mask.q_indices, (2, 3))
+    if block_mask.full_q_num_blocks is not None:
+        mark_dynamic(block_mask.full_q_num_blocks, 2)
+    if block_mask.full_q_indices is not None:
+        mark_dynamic(block_mask.full_q_indices, (2, 3))
```
epwalsh, line 1062: Which dimensions are being marked dynamic here? Sequence I assume? What about batch dimension?

src/olmo_core/nn/attention/__init__.py
```diff
@@ -929,3 +1004,100 @@ def repeat_kv(x: torch.Tensor, n_rep: int) -> torch.Tensor:
...
+    def _causal_mask_mod(
+        B: torch.Tensor, H: torch.Tensor, q_idx: torch.Tensor, kv_idx: torch.Tensor
+    ) -> torch.Tensor:
+        return q_idx >= kv_idx
+
+    mask_mods.append(_causal_mask_mod)
+
+    if window_size is not None and window_size != (-1, -1):
+
+        def _sliding_window_mask_mod(
```
tyler-romero, line 1025: The key value prop of FlexAttn is that its easy to experiment with a variety of different attention types. I think we should try to expose a config interface that allows users to define the set of masks and the score mod they want to use.

src/olmo_core/nn/attention/__init__.py
2015aroras: I'm not aware of any reason.

src/olmo_core/nn/attention/__init__.py
tyler-romero: This is actually coming up currently with @aman-17 's work trying to add in GPT-OSS style attention sinks

2015aroras:
Too old, not needed by me anymore, and I believe Aman has a better implementation now anyways.

PR #324, opened 2025-07-15
`GenerationModule` with support for KV Caching

tyler-romero:
Add a simple GenerationModule (analogous to `TrainModule`) that can be used to configure and run autoregressive next token prediction.

As of now this supports:
	1. loading distributed checkpoints exactly as they were saved during training. 
	2. `temperature`, `top_k`, `top_p` params for generation.
	3. using FSDP to shard larger models across multiple devices (mostly as a demonstration of how other types of parallelism can be worked in).
	4. HF-style attention masks so that batched generation w/ left-padding is supported
	5. HF-like API
	6. Generation with kv caching

Across a large suite of tasks Olmo-core native generation almost exactly replicates results achieved using vLLM. Generation speed with kv-cache is fairly fast too. Much faster than HF, owing in part to some memory saving techniques allowing for larger inference batch sizes.


src/olmo_core/train/train_module/transformer/common.py
```diff
@@ -36,8 +36,8 @@ def parallelize_model(
     *,
     world_mesh: Optional[DeviceMesh],
     device: torch.device,
-    max_sequence_length: int,
-    rank_microbatch_size: int,
+    max_sequence_length: Optional[int] = None,
+    rank_microbatch_size: Optional[int] = None,
```
tyler-romero, line 40: These actually *are* optional and if provided are used to warm up the RoPE cache (max_seq_len) and an MoE component (rank_microbatch_size).

src/olmo_core/generate/generation.py
epwalsh, line 1: Nit: maybe rename this `generation_module.py` to be more specific. Also consider moving the transformer implementation to its own submodule.

src/olmo_core/generate/generation.py
```diff
@@ -0,0 +1,377 @@
...
+    get_fs_local_rank,
+    get_rank,
+    get_world_size,
+    is_distributed,
+    scatter_object,
+)
+from olmo_core.exceptions import OLMoConfigurationError
+from olmo_core.float8 import Float8Config
+from olmo_core.generate.config import GenerationConfig
+from olmo_core.generate.selection import temperature_sampling
```
epwalsh: Nit: relative imports are nice to have at least within the same submodule.
```suggestion
from .config import GenerationConfig
from .selection import temperature_sampling
```

src/olmo_core/generate/generation.py
```diff
@@ -0,0 +1,377 @@
...
+        Args:
+            checkpoint_dir: Path to checkpoint directory
+            work_dir: Working directory for caching remote checkpoints
+            process_group: Process group for distributed loading
+            pre_download: Whether to pre-download remote checkpoints
+            load_thread_count: Number of threads to use for loading the checkpoint
+
+        Raises:
+            FileNotFoundError: If checkpoint directory doesn't exist
+            RuntimeError: If checkpoint loading fails
```
epwalsh: nit: this docstring syntax is inconsistent with our other docstrings

src/olmo_core/generate/generation.py
```diff
@@ -0,0 +1,377 @@
...
+
+        # Load transformer config from checkpoint if not provided
+        if transformer_config is None and get_rank(process_group) == 0:
+            experiment_config = ExperimentConfig.from_file(checkpoint_dir / "config.json")
+            transformer_config = experiment_config.model
+
+        # Create work directory on rank 0
+        work_dir = Path(
+            work_dir or (tempfile.mkdtemp() if get_rank(process_group) == 0 else "/tmp")
+        )
```
epwalsh: This assumes all ranks share the filesystem? Which is usually only true for single-node jobs.

epwalsh: Alternatively just force the user to provide a work dir.

pyproject.toml
```diff
@@ -32,33 +31,36 @@ Changelog = "https://github.com/allenai/OLMo-core/blob/main/CHANGELOG.md"
...
     "sphinx-autodoc-typehints==1.23.3",
+    "sphinx-copybutton",
     "sphinx-inline-tabs",
+    "twine>=1.11.0",
+    "wheel",
 ]
 beaker = ["beaker-py>=1.36.0,<2.0", "GitPython>=3.0,<4.0"]
-wandb = ["wandb"]
 comet = ["comet_ml"]
+eval = ["ai2-olmo-eval==0.8.4"]
```
tyler-romero: similar to wandb, eval is only needed if the eval callback is added

pyproject.toml
```diff
@@ -32,33 +31,36 @@ Changelog = "https://github.com/allenai/OLMo-core/blob/main/CHANGELOG.md"
 
 [project.optional-dependencies]
 dev = [
-    "ruff",
-    "mypy>=1.0,<1.4",
     "black>=23.1,<24.0",
+    "boto3",
```
tyler-romero, line 35: just sorting alphabetically

pyproject.toml
tyler-romero: It also causes issues when trying to use olmo-core generation within the eval lib

src/olmo_core/nn/attention/__init__.py
```diff
@@ -373,14 +379,6 @@ def __init__(
                 raise OLMoConfigurationError(
                     f"fused RoPE is not compatible with {self.__class__.__name__}"
                 )
-
-            # On layers with sliding windows, we don't do rope extension.
```
tyler-romero, line 377: this should be configurable / is configurable via `block_overrides`

src/olmo_core/generate/generation_module/config.py
```diff
@@ -0,0 +1,151 @@
...
+            raise ValueError(f"temperature must be non-negative, got {self.temperature}")
+        if self.top_k <= 0 and self.top_k != -1:
+            raise ValueError(f"top_k must be positive or -1, got {self.top_k}")
+        if self.top_p <= 0.0 or self.top_p > 1.0:
+            raise ValueError(f"top_p must be in (0, 1], got {self.top_p}")
+
+
+@beta_feature
+@dataclass
+class TransformerGenerationModuleConfig(Config):
```
epwalsh: I think this is a duplicate of what's in `.transformer.generation_module.config`?

src/olmo_core/generate/generation_module/transformer/config.py
```diff
@@ -0,0 +1,93 @@
...
+            config_dict["autocast_precision"] = DType(autocast_precision).as_pt()
+        if (state_dict_save_opts := config_dict.pop("state_dict_save_opts", None)) is not None:
+            config_dict["state_dict_save_opts"] = dist_cp_sd.StateDictOptions(
+                **state_dict_save_opts
+            )
+        if (state_dict_load_opts := config_dict.pop("state_dict_load_opts", None)) is not None:
+            config_dict["state_dict_load_opts"] = dist_cp_sd.StateDictOptions(
+                **state_dict_load_opts
+            )
+        print(config_dict)
```
epwalsh: Might want to only print from rank 0, or local (node) rank 0.

src/olmo_core/generate/generation_module/transformer/generation_module.py
```diff
@@ -0,0 +1,528 @@
...
+            )
+
+        # Build model and generation module
+        if dtype is not None:
+            dtype = DType(dtype)
+            transformer_config.apply(
+                lambda c: setattr(c, "dtype", dtype) if hasattr(c, "dtype") else None
+            )
+        print(transformer_config)
+        print(generation_config)
```
epwalsh: Might want to only print from rank 0, or local (node) rank 0.

src/olmo_core/generate/generation_module/transformer/generation_module.py
```diff
@@ -0,0 +1,528 @@
...
+            print(f"  Sequence length: {prompt_len:,} → {prompt_len + tokens_generated:,}")
+            print(f"  Total generation time: {total_time:.3f}s")
+            # Calculate prefill and completion throughput
+            prefill_tokens_total = prompt_len * batch_size
+            completion_tokens_total = tokens_generated * batch_size
+
+            print("  Throughput:")
+            print(
+                f"    Overall: {tokens_per_sec_total:.1f} tokens/s (total) | {tokens_per_sec_per_seq:.1f} tokens/s (per seq)"
+            )
```
epwalsh: Any reason to prefer `print(...)` over `log.info(...)` here?

src/olmo_core/generate/generation_module/transformer/generation_module.py
```diff
@@ -0,0 +1,528 @@
...
+
+        # Create work directory on rank 0
+        work_dir = Path(
+            work_dir or (tempfile.mkdtemp() if get_rank(process_group) == 0 else "/tmp")
+        )
+
+        # Broadcast config and work_dir to all ranks
+        transformer_config, work_dir, tokenizer_config = scatter_object(
+            (transformer_config, work_dir, tokenizer_config)
+        )
```
epwalsh, line 465: Nice 😎

src/olmo_core/version.py
```diff
@@ -1,5 +1,5 @@
 _MAJOR = "2"
-_MINOR = "1"
+_MINOR = "2"
```
epwalsh: I typically wait until I do an actual release to change these, otherwise it's easy to accidentally increment the version twice.

src/olmo_core/nn/transformer/model.py
```diff
@@ -416,8 +430,12 @@ def _prepare_inputs(
...
             input_ids = move_to_device(input_ids, self.device)
             labels = move_to_device(labels, self.device)
-            all_block_kwargs["max_doc_len"] = max_doc_len
-            all_block_kwargs["cu_doc_lens"] = move_to_device(cu_doc_lens, self.device)
+
+            if max_doc_len is not None or cu_doc_lens is not None:
+                all_block_kwargs["max_doc_len"] = max_doc_len
+                all_block_kwargs["cu_doc_lens"] = move_to_device(cu_doc_lens, self.device)
+            elif cache_leftpad is not None:
+                all_block_kwargs["cache_leftpad"] = move_to_device(cache_leftpad, self.device)
```
epwalsh: Are these fundamentally mutually exclusive, or is it just harder to support both? In any case, should we raise an exception if both are supplied?

src/olmo_core/generate/generation_module/transformer/generation_module.py
tyler-romero: No, I meant to come back and clean this up but overlooked it

src/olmo_core/nn/transformer/model.py
tyler-romero: Good q: mutually exclusive. will raise an exception

src/olmo_core/generate/generation_module/config.py
tyler-romero: whoops - refactor mishap

src/olmo_core/generate/generation_module/transformer/config.py
tyler-romero: yeah good point

epwalsh (approved):
LGTM!

PR #329, opened 2025-07-28
Make removing ephemeral checkpoints more robust

epwalsh:


src/olmo_core/train/callbacks/checkpointer.py
```diff
@@ -155,8 +155,12 @@ def _remove_checkpoint(self, path: str):
         log.info(f"Removing old checkpoint at '{path}'...")
         if is_url(path):
             if get_rank() == 0:
+                # Remove metadata file first to invalidate the checkpoint.
+                remove_file(join_path(path, self.trainer.checkpointer.METADATA_FNAME))
                 self.trainer.run_bookkeeping_op(clear_directory, path)
         elif get_fs_local_rank() == 0:
+            # Remove metadata file first to invalidate the checkpoint.
+            remove_file(join_path(path, self.trainer.checkpointer.METADATA_FNAME))
             self.trainer.run_bookkeeping_op(clear_directory, path)
```
dirkgr: At this point in time we don't know that the file exists, and it'll fail if it doesn't.

epwalsh: fixed

abertsch72 (approved):
thank you!

PR #330, opened 2025-07-28
More Rope Scaling Implementations (PI, Yarn)

tyler-romero:
Implements
1. Absolute base frequency change
2. Position Interpolation
3. Stepwise (llama 3.1) scaling
4. Yarn

Note that per-layer rope theta modifications are already supported via the `block_overrides` arg to `TransformerConfig`

### DIFF
--- a/src/olmo_core/nn/rope.py
+++ b/src/olmo_core/nn/rope.py
@@ -40,35 +40,199 @@ class RoPEType(StrEnum):
     """
 
 
+def compute_inv_freqs(theta: int, dim: int, device: torch.device) -> "torch.Tensor":
+    inv_freq = 1.0 / (theta ** (torch.arange(0, dim, 2, device=device, dtype=torch.float) / dim))
+    return inv_freq
+
+
 @dataclass
 class RoPEScalingConfig(Config):
     """
-    Defines how to scale RoPE to longer sequence lengths.
+    Base class for RoPE scaling configs. Defines a strategy for scaling RoPE to longer sequences.
+    """
+
+    attention_rescale_factor: float = 1.0
+    """
+    Factor to rescale attention scores by when using scaled RoPE. Can be used to compensate for
+    the larger effective context. 1.0 means no rescaling.
+    """
+
+    @abstractmethod
+    def compute_scaled_inv_freq(
+        self, theta: int, dim: int, device: torch.device
+    ) -> tuple["torch.Tensor", float]:
+        """Compute the scaled inverse frequencies for RoPE, and the attention rescaling factor."""
+        raise NotImplementedError
+
+
+@dataclass
+class ABFRoPEScalingConfig(RoPEScalingConfig):
+    """Absolute base frequency scaling (ABF). Simply uses a new base frequency parameter."""
+
+    new_theta: int = 8_000_000
+
+    def compute_scaled_inv_freq(
+        self, theta: int, dim: int, device: torch.device
+    ) -> tuple["torch.Tensor", float]:
+        inv_freq = compute_inv_freqs(self.new_theta, dim, device)
+        return inv_freq, self.attention_rescale_factor
+
+
+@dataclass
+class PIRoPEScalingConfig(RoPEScalingConfig):
+    """
+    Position-Interpolation (PI) RoPE scaling from Chen et al. (https://arxiv.org/pdf/2306.15595)
+
+    Interpolate the rotary angles instead of extrapolating them when the context window at
+    inference time exceeds the window used during training. In practice, this amounts to linearly
+    *compressing* the original position indices by a constant factor ``factor``.
+    """
+
+    factor: float = 2.0
+    """Context expansion multiplier. If factor = 1, reduces to vanilla RoPE."""
+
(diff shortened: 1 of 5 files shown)


src/olmo_core/nn/rope.py
```diff
@@ -43,32 +43,191 @@ class RoPEType(StrEnum):
...
+    """
+
+    attention_rescale_factor: float = 1.0
+    """
+    Factor to rescale attention scores by when using scaled RoPE. Can be used to compensate for
+    the larger effective context.
+    """
+
+    def __post_init__(self):
+        if self.attention_rescale_factor < 1.0:
```
abertsch72: do we have to restrict attention rescaling factor to be >1?

src/olmo_core/nn/rope.py
```diff
@@ -43,32 +43,191 @@ class RoPEType(StrEnum):
...
+        # Positional-interpolation scales the *positions* by 1/factor. This is
+        # equivalent to scaling the inverse frequencies by the same amount.
+        if self.factor != 1.0:
+            inv_freq = inv_freq / self.factor
+
+        return inv_freq, self.attention_rescale_factor
+
+
+@dataclass
+class PerFrequencyRoPEScalingConfig(RoPEScalingConfig):
```
abertsch72: nit, but the name suggests to me that you get more finegrained control than high/low...

src/olmo_core/nn/rope.py
```diff
@@ -43,32 +43,191 @@ class RoPEType(StrEnum):
...
+    Denominator that determines the *low-frequency* wavelength cut-off
+    (a larger value moves the cut-off towards shorter wavelengths).
+    """
+
     high_freq_factor: float = 4.0
+    """
+    Denominator that determines the *high-frequency* wavelength cut-off
+    (a smaller value keeps more of the very short wavelengths untouched).
+    """
+
```
abertsch72, line 139: it would be nice if these params used the same name / interpretation as the YaRN params that control the same thing... minimally, I'd vote for calling them something other than factor because we also have `factor` above that does a different thing

(I know these names are probably holdovers from the methods, though-- if this is how huggingface does the two methods then maybe it's better to be consistent with them rather than internally consistent)

abertsch72 (commented):
thanks so much for standardizing this! were you also going to add a way to turn on/off sliding-window-layer scaling? this seems more important now that it seems like some of our sliding window scaling runs actually outperform global-only scaling

tyler-romero:
`PerFrequencyRoPEScalingConfig` is the same as the old rope scaling config. I think its wrong to think of that as the base case, and there are 0 dependencies on that in main, so I'm not sure its worth distorting the factorization of the code for.

src/olmo_core/nn/rope.py
tyler-romero: Yeah good point, I'll switch to calling it stepwise

abertsch72:
> the base class does basic RoPE extension like the RoPE paper says

> PerFrequencyRoPEScalingConfig is the same as the old rope scaling config

I think these are both true! The old RoPEScalingConfig implemented llama 3.1-style, which is called PerFrequency here, but almost every run was done with only scaling RoPE theta (the most basic type of extension), and not passing a RoPE config at all (which is the same as the base config here as long as you don't change attn temperature).

src/olmo_core/nn/rope.py
tyler-romero: Yeah I think I want the variables to be comperable to other oss implementations. But I agree that having so many different "factors" is annoying so I updated the args to be "proportions" instead, which I also think is a bit easier to reason about.

tyler-romero:
> were you also going to add a way to turn on/off sliding-window-layer scaling? this seems more important now that it seems like some of our sliding window scaling runs actually outperform global-only scaling

It seems as though this is already possible with the block_overrides kwarg to `TransformerConfig`.

I added block_mods arg to `llama_like` to make this a bit less cumbersome.

src/olmo_core/nn/rope.py
```diff
@@ -199,20 +369,27 @@ def _get_rotary_embedding(
...
-                self.theta
-                ** (torch.arange(0, self.dim, 2, device=device, dtype=torch.float) / self.dim)
-            )
-            if self.scaling is not None:
-                inv_freq = self.scaling.scale_inv_freq(inv_freq)
+            if self.scaling is None:
+                inv_freq = 1.0 / (
+                    self.theta
+                    ** (torch.arange(0, self.dim, 2, device=device, dtype=torch.float) / self.dim)
+                )
```
epwalsh: I think the helper function you added does the same thing, right?
```suggestion
                inv_freq = compute_inv_freqs(self.theta, self.dim, device)
```

epwalsh (commented):
I didn't review the logic in the scaling strategies closely, but the high level changes look good to me. Left one suggestion.

src/olmo_core/nn/rope.py
tyler-romero: yes missed this one, thanks!

PR #333, opened 2025-07-31
Make async bookkeeping more robust

epwalsh:


src/olmo_core/train/callbacks/beaker.py
```diff
@@ -132,7 +132,8 @@ def _update(self):
             self._set_description,
             self.trainer.training_progress,
             op_name="beaker_set_description",
-            cancel_in_progress=True,
+            allow_multiple=False,
+            distributed=False,
```
epwalsh, line 136: `allow_multiple=False` because there's no point in running multiple invocations of this op concurrently. One at a time is enough. `distributed=False` because this op doesn't require distributed communication, so it can run concurrently with other ops.

dirkgr (approved):
All bookkeeping ops have to be threadsafe. But they already had to be threadsafe before, so no change?

epwalsh:
> All bookkeeping ops have to be threadsafe. But they already had to be threadsafe before, so no change?

Right

PR #338, opened 2025-08-06
Port the WandB comparison tool from the old trainer

dirkgr:
It needed no changes except for imports.


dirkgr:
Three suggestions from Copilot. All wrong 😔

PR #340, opened 2025-08-08
Add 'namespaces' to BufferCache to avoid collisions

epwalsh:


src/olmo_core/nn/buffer_cache.py
```diff
@@ -24,3 +49,8 @@ def get_for_device(self, key: str, device: torch.device) -> Optional[torch.Tenso
             return tensor
         else:
             return None
+
+    def with_namespace(self, namespace: str) -> "BufferCache":
```
tyler-romero, line 53: Just talking through this for myself:
This creates a new BufferCache object with a pointer to the same underlying data. So that if the namespace is the same, the data is not duplicated but if the namespace is different a new entry is added.

tyler-romero (commented):
LGTM!

tyler-romero (approved):

src/olmo_core/nn/buffer_cache.py
epwalsh: Right. Actually I'm going to paraphrase what you just wrote into the docstring: a5fb42d4

PR #344, opened 2025-08-12
Support configuring `head_stride` for ring-flash-attn

tyler-romero:
Ring Flash Attn processes `num_kv_heads // head_stride` "batches" of (all_gather -> flash_attn) ops.

Increasing head_stride means that fewer, larger communication ops will be used and fewer, larger calls to flash-attn will be made. This can possibly harm overlap, but also many small communication ops can be inefficient.

### DIFF
--- a/src/olmo_core/nn/moe/router.py
+++ b/src/olmo_core/nn/moe/router.py
@@ -54,9 +54,9 @@ class _UniformExpertAssignment(torch.autograd.Function):
         return out.view(x.shape)
 
 
-_uniform_expert_assignment: Callable[[torch.Tensor, int], torch.Tensor] = (
-    _UniformExpertAssignment.apply
-)  # type: ignore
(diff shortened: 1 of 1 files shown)


src/olmo_core/nn/attention/__init__.py
```diff
@@ -423,7 +428,7 @@ def sdpa(
                 max_seqlen=max_doc_len,
                 max_seqlen_q=max_doc_len_q,
                 max_seqlen_k=max_doc_len_k,
-                heads_k_stride=1,  # TODO: should this ever not be 1?
```
epwalsh, line 426: Thank you for answering this question! 😉

epwalsh (approved):

src/olmo_core/nn/attention/__init__.py
tyler-romero: :)

PR #346, opened 2025-08-15
Add option to greedily pack documents with the padded FSL dataset

epwalsh:
Currently the padded FSL dataset creates a single instance for each document, adding padding when necessary to ensure all instances at runtime have the target sequence length. This is nice for evals, for we're thinking we could potentially use this for training if we made it a little more efficient by greedily packing documents into instances. There will still be some padding, how much depends on the composition of source files.

You can enable this in a training script by adding the options:

```
dataset.name=padded_fsl --dataset.do_greedy_packing=true
```

Note: this probably shouldn't be used when training with tensor or context parallelism. Since instances will be right-padded, the highest rank within a TP/CP group will see fewer trainable tokens on average, which essentially means those tokens will get higher weight.


epwalsh:
update: this is actually just a quick-and-dirty version of the packed FSL dataset with `long_doc_strategy=truncate`.

PR #348, opened 2025-08-18
Add `io.glob_directory` function

epwalsh:
The cookbook has this functionality so we should have it here as well.


soldni (approved):
LGTM!

PR #349, opened 2025-08-18
Option to group npy sources together for packing

epwalsh:
Adds the ability to group multiple `*.npy` source files together for packing with the numpy packed FSL dataset via the option `source_group_size`.  For example, setting `source_group_size` to 8 is equivalent to concatenating every 8 consecutive numpy sources together and passing those to the `NumpyPackedFSLDataset`.

I also added an option (`source_permutation_seed`) to automatically shuffle the source files before applying packing to get a better distribution within each group.

So to enable this with one of our training scripts, add these flags (assuming a group size of 8, which seems reasonable):

```
--dataset.name=packed_fsl --dataset.source_group_size=8 --dataset.source_permutation_seed=423423
```

### DIFF
--- a/src/olmo_core/data/numpy_dataset.py
+++ b/src/olmo_core/data/numpy_dataset.py
@@ -2418,12 +2418,12 @@ class NumpyDatasetConfig(Config):
     """
     source_group_size: Optional[int] = None
     """
-    Determines how many sources are grouped together a single source for packing with the packed
+    Determines how many sources are grouped together as a single source for packing with the packed
     FSL dataset.
     """
     source_permutation_seed: Optional[int] = None
     """
-    Used to shuffle the source files.
+    Used to shuffle the source files before handing off to the dataset class.
     """
 
     def validate(self):
--- a/src/olmo_core/data/numpy_dataset.py
+++ b/src/olmo_core/data/numpy_dataset.py
@@ -1080,14 +1080,16 @@ class NumpyPackedFSLDataset(NumpyFSLDatasetBase):
     def source_instance_offsets(self) -> Tuple[Tuple[int, int], ...]:
         if self._source_instance_offsets is None:
             item_size = self.indices_dtype(0).itemsize
-            num_instances_per_path = self.map(
-                lambda path, _: get_file_size(self._get_instance_offsets_path(path))
-                // (item_size * 2)
+            num_instances_per_group = self.map(
+                lambda path, _: get_file_size(path) // (item_size * 2),
+                _paths=[
+                    self._get_instance_offsets_path(*paths)
+                    for paths in chunked(self.paths, self.source_group_size)
+                ],
             )
             array_instance_offsets = []
             start_offset = 0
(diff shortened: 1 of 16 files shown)


src/olmo_core/io.py
```diff
@@ -895,6 +900,10 @@ def _s3_get_bytes_range(
     except ClientError as e:
         if e.response["ResponseMetadata"]["HTTPStatusCode"] == 404:
             raise FileNotFoundError(f"{scheme}://{bucket_name}/{key}") from e
+        elif e.response["Error"]["Code"] == "InvalidRange":
+            raise OLMoInvalidRangeRequestError(
+                f"Invalid range request to '{scheme}://{bucket_name}/{key}' ({bytes_start=}, {num_bytes=})"
+            )
```
epwalsh, line 906: ran into this due to a bug in my packing

src/olmo_core/data/utils.py
```diff
@@ -671,6 +680,17 @@ def find_periodic_sequences(
...
 
 
+T = TypeVar("T")
+
+
+def _take(n: int, iterable: Iterable[T]) -> List[T]:
+    return list(islice(iterable, n))
+
+
+def chunked(iterable: Iterable[T], n: int) -> Iterable[List[T]]:
```
epwalsh, line 690: same as the function from `more-itertools`, I just didn't want to pull in an additional dependency

src/olmo_core/train/train_module/transformer/train_module.py
```diff
@@ -336,10 +336,16 @@ def train_batch(self, batch: Dict[str, Any], dry_run: bool = False):
...
             )
 
-        # Calculate how many tokens are going to be used in the loss.
+        # Calculate and record how many tokens are going to be used in the loss.
+        batch_num_tokens = batch["labels"].numel()
         batch_num_tokens_for_loss = move_to_device(
             (batch["labels"] != self.label_ignore_index).sum(), self.device
         )
+        self.record_metric(
+            "train/masked labels (%)",
```
tyler-romero, line 345: nice

src/olmo_core/data/numpy_dataset.py
```diff
@@ -2335,6 +2416,15 @@ class NumpyDatasetConfig(Config):
     """
     Determines how long documents are handled with the packed FSL dataset.
     """
+    source_group_size: Optional[int] = None
```
tyler-romero, line 2419: for the future: it would be nice to wrap groups of options for specific dataloaders like FSL into their own config classes or something to make it more obvious which ones are valid / can be used in combination.

tyler-romero (approved):
LGTM!

src/olmo_core/data/numpy_dataset.py
epwalsh: Totally agree. In fact, I wrote a little library to make doing that sort of thing a whole lot easier: https://github.com/epwalsh/dataclass-extensions?tab=readme-ov-file#registrable-subclasses

On my TODO list to integrate that into OLMo-core at some point.

PR #351, opened 2025-08-20
Add option to `Trainer.load_checkpoint()` to ignore optim state

epwalsh:


dirkgr (commented):
If it's `None`, it should do what it does now. If it's `True` or `False`, it should throw an exception when it can't do the thing. Is that what happens in this?

epwalsh:
> If it's None, it should do what it does now. If it's True or False, it should throw an exception when it can't do the thing. Is that what happens in this?

The default behavior (when `None`) stays the same. If `True` but there's no optim state, we raise an exception. If `False`, there's no case where an exception should be raised unless some other part of the checkpoint is invalid.

PR #353, opened 2025-08-21
Fix TP when headwise QK norm are applied

tyler-romero:


pyproject.toml
```diff
@@ -137,7 +137,7 @@ reportPrivateImportUsage = false
 ignore_missing_imports = true
 no_site_packages = true
 check_untyped_defs = true
-no_namespace_packages = true
```
tyler-romero, line 140: unrelated, fixed nonexistent setting

src/olmo_core/nn/rope.py
```diff
@@ -445,14 +445,6 @@ def forward(
                 pos_sin, pos_cos = self._get_rotary_embedding(k_len, q_.device)
             pos_sin, pos_cos = pos_sin.type_as(q_), pos_cos.type_as(q_)
 
-            head_dim_local = q_.shape[-1]
-            if pos_sin.size(-1) < head_dim_local or pos_cos.size(-1) < head_dim_local:
-                raise RuntimeError(
```
tyler-romero, line 450: TP shouldn't shard across the head_dim :)

src/test/nn/attention_test.py
```diff
@@ -377,3 +387,150 @@ def test_sliding_window_attention_config_invalid_pattern_error():
...
+    ],
+)
+@pytest.mark.parametrize(
+    "load_balancer_type",
+    [
+        pytest.param(RingAttentionLoadBalancerType.zig_zag, id="zig_zag"),
+        pytest.param(RingAttentionLoadBalancerType.llama3, id="llama3"),
+    ],
+)
+@pytest.mark.skip("known precision issues with ring-flash-attn")
```
tyler-romero, line 505: todo: remove and adjust tolerance

src/test/nn/attention_test.py
```diff
@@ -198,7 +212,6 @@ def test_attention_with_intra_document_masking():
     torch.testing.assert_close(y1, y2)
     torch.testing.assert_close(y1_fused, y2_fused)
     torch.testing.assert_close(y1, y1_fused)
-    torch.testing.assert_close(y2, y2_fused)
```
2015aroras: nit: Why remove this?

src/olmo_core/nn/rope.py
2015aroras: nit/thought: Maybe keep this check with an updated error then?

2015aroras (commented):
I'm not very knowledgeable here but nothing obviously wrong

2015aroras (approved):

src/olmo_core/nn/rope.py
tyler-romero: I dont think it makes sense, there are plenty of dims that TP shouldnt shard across throughout a transformer model and we dont check all of them. It'll raise a shape error anyways if it is being sharded across the wrong dim.

src/test/nn/attention_test.py
tyler-romero: whoops, my b

PR #355, opened 2025-08-22
Refine hostname constraints for beaker experiments on Google clusters

2015aroras:
This PR cleans up the beaker host selection logic for runs on Google clusters. Options for disabling hostname constraints and for specifying the number of 'execution units' (e.g. model replicas) are added. Host selection logic is switched from hardcoded on to opt-in, since host selection logic is basically incompatible with preemption (if one of your nodes is taken by a higher prio job, you might not be able to relaunch even though nodes free up later on).

### DIFF
--- a/src/olmo_core/launch/beaker.py
+++ b/src/olmo_core/launch/beaker.py
@@ -33,7 +33,7 @@ from ..exceptions import BeakerExperimentFailedError, OLMoConfigurationError
 from ..train.callbacks.beaker import BEAKER_RESULT_DIR
 from ..utils import LOG_FILTER_TYPE_ENV_VAR, LogFilterType
 from ..version import VERSION
-from .select_beaker_hosts import get_host_name_constraints
+from .select_beaker_hosts import get_beaker_hostname_constraints
 from .utils import GIT_BRANCH_ENV_VAR, GIT_REF_ENV_VAR, GIT_REPO_URL_ENV_VAR, GitConfig
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


dirkgr:
Hostname selection also breaks Beaker restarts, so we should probably not allow both at the same time.

src/olmo_core/launch/beaker.py
```diff
@@ -242,6 +242,22 @@ class BeakerLaunchConfig(Config):
     The directory of the Beaker results dataset.
     """
 
+    use_hostname_constraints: bool = False
+    """
+    Uses hostname constraints to restrict the hostnames on which the experiment runs. This is currently
+    only supported for Augusta clusters, and can benefit performance by forcing the use of colocated nodes.
+
+    This is NOT recommended to be used lower priority preemptible jobs, since hostname constraints are not
```
epwalsh:
```suggestion
    This is NOT recommended to be used with lower priority preemptible jobs, since hostname constraints are not
```

src/olmo_core/launch/beaker.py
```diff
@@ -242,6 +242,22 @@ class BeakerLaunchConfig(Config):
...
+    updated on preemption.
+    """
+
+    num_execution_units: Optional[int] = None
+    """
+    Number of \"execution units\", defaults to ``max(1, num_nodes // 32)``. An \"execution unit\" is abstraction
```
epwalsh: I don't think the backslashes are necessary
```suggestion
    Number of "execution units", defaults to ``max(1, num_nodes // 32)``. An "execution unit" is abstraction
```

epwalsh (approved):
In our internal experiment config, can we automatically set the default `num_execution_units` to the number of replicas when HSDP is enabled? Or at issue a warning if they differ? Otherwise LGTM

2015aroras:
> In our internal experiment config, can we automatically set the default `num_execution_units` to the number of replicas when HSDP is enabled? Or at issue a warning if they differ? Otherwise LGTM

https://github.com/allenai/OLMo-core/pull/355/commits/f56fbf8ab246ca57633e120b82ef4265a1115162 I have (hackishly) set it as a train time warning because the number of replicas is not necessarily known at launch time (say, if you use shard degree to specify the dp degree). Not yet tested

2015aroras:
https://github.com/allenai/OLMo-core/pull/355/commits/ae244b65802cf1e76256960cb0b524f85ba76cd8 I've changed the logic so that it (still hackily) determines the number of execution units from the dp config.

PR #364, opened 2025-09-08
redo base dir calculation for new cluster names

abertsch72:
Because the canonical cluster names no longer have `-cirrascale` or `-google`, the logic here for detecting the base dir is wrong. this causes runs on augusta that use this logic to infer the wrong location for pretraining data.

Listing out the cluster names isn't the most elegant solution-- not sure if there's a better way to do this?


dirkgr:
Update on this? Are we merging it?

PR #366, opened 2025-09-11
Don't do RoPE extension in sliding window layers

dirkgr:


src/olmo_core/nn/attention/__init__.py
```diff
@@ -378,6 +378,14 @@ def __init__(
                 raise OLMoConfigurationError(
                     f"fused RoPE is not compatible with {self.__class__.__name__}"
                 )
+
+            # On layers with sliding windows, we don't do rope extension.
+            uses_full_attention = self.window_size == (-1, -1)
+            uses_sliding_window = not uses_full_attention
+            if uses_sliding_window and rope.scaling is not None:
+                rope = rope.replace(scaling=None)
```
tyler-romero, line 386: nit: I think it would be good to log a warning here

dirkgr:
Tyler has a better solution.

PR #367, opened 2025-09-11
Optionally don't load optimizer state

dirkgr:


dirkgr:
Are we going to regret not separating trainer state and optimizer state?

src/olmo_core/train/trainer.py
```diff
@@ -172,6 +172,16 @@ class Trainer:
...
     """
 
+    load_trainer_state: Optional[bool] = None
+    """
+    Whether to load trainer and optimizer state from the checkpoint.
+    
+    When this is False, we will only load weights.
+    When this is True, we will load trainer and optimizer state, and fail loudly if we can't.
+    When this is None, we will try to load the trainer and optimizer state, but fall back quietly to just loading
+    weights if we can't.
```
epwalsh, line 182: Check out the signature of `Trainer.load_checkpoint()`. There's a separate option to control loading optimizer state, so there should also be a separate option here.

epwalsh:
> Are we going to regret not separating trainer state and optimizer state?

They are separate

PR #369, opened 2025-09-11
Makes it possible to override the common config builder

dirkgr:


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Increased NCCL_FASTRAK_PLUGIN_ACCEPT_TIMEOUT_MS from 10 minutes to 30 minutes.
+- Makes it possible to override the common config builder when specifying experiments
```
epwalsh: I would just leave this out of the changelog since it only impacts an internal API

dirkgr: Then the check is going to yell at me?

epwalsh: It's not a required check, don't let it bully you

PR #370, opened 2025-09-12
Set accum_grad to fp32 for fused_linear_cross_entropy_loss, bump liger-kernel version

tyler-romero:
This setting to fp32 rather than the default bf16 is both faster (yes actually!) and more stable. But costs slightly more memory than the bf16 setting (still far less than using a default linear -> cross_entropy) setup.

### DIFF
--- a/src/olmo_core/launch/beaker.py
+++ b/src/olmo_core/launch/beaker.py
@@ -57,7 +57,7 @@ __all__ = [
 
 BeakerPriority = Priority
 
-_DEFAULT_TORCH = "2.7.0".replace(".", "")
+_DEFAULT_TORCH = "2.7.1".replace(".", "")
 _DEFAULT_CUDA = "12.8".replace(".", "")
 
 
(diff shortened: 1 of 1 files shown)


src/Dockerfile
```diff
@@ -66,7 +66,7 @@ ARG RING_FLASH_ATTN_VERSION=0.1.5
 RUN pip install --no-build-isolation --no-cache-dir ring-flash-attn==${RING_FLASH_ATTN_VERSION}
 
 # Install liger-kernel.
-ARG LIGER_KERNEL_VERSION=0.5.10
+ARG LIGER_KERNEL_VERSION=0.6.2
```
epwalsh, line 69: This needs to get updated in the Makefile primarily, though good to update here as well

tyler-romero:
We have a pretty good test for this functionality already (https://github.com/allenai/OLMo-core/blob/tyler/liger-kernel-stability/src/test/nn/lm_head_test.py#L33) that didn't notice this - and it seems its fine in most scenarios according to the maintainers of liger. Not exactly sure how to test this effectively.

More details in thread here: https://github.com/linkedin/Liger-Kernel/issues/512

But it seems this is similar to casting the logits to fp32 before CE.

PR #371, opened 2025-09-12
Add the ability to send Slack notifications from `launch.beaker`

epwalsh:
This adds the option to send direct Slack messages from our `olmo_core.launch.beaker` module 
when following a job simply by setting the environment variable `SLACK_WEBHOOK_URL` locally. With this we'll get more reliable notifications on failure.

Example notification when a run has launched:

image

This also adds more notification events to our `SlackNotifierCallback`, so we'll get updates each time a checkpoint is saved or an epoch is completed.

### DIFF
--- a/src/olmo_core/train/callbacks/slack_notifier.py
+++ b/src/olmo_core/train/callbacks/slack_notifier.py
@@ -1,14 +1,18 @@
+import logging
 import os
 from dataclasses import dataclass
 from typing import Optional
 
 import requests
 
+from olmo_core.aliases import PathOrStr
 from olmo_core.config import StrEnum
 from olmo_core.distributed.utils import get_rank
 from olmo_core.exceptions import OLMoEnvironmentError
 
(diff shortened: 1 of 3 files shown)


src/olmo_core/train/callbacks/slack_notifier.py
```diff
@@ -126,4 +144,7 @@ def _post_message(self, msg: str):
         if BEAKER_JOB_ID_ENV_VAR in os.environ:
             msg = f"{msg}\n*Beaker job:* https://beaker.org/job/{os.environ[BEAKER_JOB_ID_ENV_VAR]}"
 
-        requests.post(webhook_url, json={"text": msg})
+        try:
+            requests.post(webhook_url, json={"text": msg})
+        except Exception as e:
+            log.exception(f"Failed to send Slack notification: {e}")
```
epwalsh, line 150: Just in case

PR #372, opened 2025-09-16
Avoid DDOS-ing Beaker for real

epwalsh:
We had a fix before to avoid spamming Beaker with secret read requests, but that fix relied on checking if the distributed backend was initialized. The problem with that is we build the config (which triggers the secret requests) before initializing the distributed backend, so we still ended up spamming Beaker. This PR changes the logic to instead check if we're running inside a Beaker batch job.

There's still some scenarios that could slip through the cracks here, but I think they're relatively unlikely. For example, if we were running a large distributed job _without_ Beaker (like with Slurm). But in that case we shouldn't be building a Beaker launch config anyway.


2015aroras (approved):

2015aroras:
Thank you!

PR #375, opened 2025-09-16
Treat cordoned beaker hosts as occupied

2015aroras:
This PR improves beaker host selection:

1. Cordoned nodes are considered occupied and their jobs aren't retrieved from the beaker APIs.
2. Beaker jobs are retrieved per node (instead of getting all cluster jobs).

This seems to be running faster locally, but that could be because there are less jobs on beaker now due to Dirk's distributed job.

Unfortunately no nodes are cordoned currently, so I can't test this fully.

### DIFF
--- a/src/olmo_core/launch/select_beaker_hosts.py
+++ b/src/olmo_core/launch/select_beaker_hosts.py
@@ -72,24 +72,31 @@ def get_occupied_beaker_hosts(
     occupied_hosts = set()
 
     cluster = beaker.cluster.get(beaker_cluster)
-    jobs = beaker.job.list(cluster=cluster)
-    for job in jobs:
-        if job.node is None:
+    nodes = beaker.cluster.nodes(cluster)
+    for node in sorted(nodes, key=lambda node: node.hostname):
+        host = node.hostname
+        assert host not in occupied_hosts, f"Host {host} is somehow already in occupied hosts"
+        if host not in hosts_metadata:
+            log.warning(f"No metadata found for beaker host {host}")
             continue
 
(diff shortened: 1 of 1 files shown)


src/olmo_core/launch/select_beaker_hosts.py
```diff
@@ -72,24 +72,31 @@ def get_occupied_beaker_hosts(
...
+        assert host not in occupied_hosts, f"Host {host} is somehow already in occupied hosts"
+        if host not in hosts_metadata:
+            log.warning(f"No metadata found for beaker host {host}")
             continue
 
-        host = beaker.node.get(job.node).hostname
-        if host not in hosts_metadata or host in occupied_hosts:
+        if node.cordoned is not None:
+            # Treat cordoned node as occupied since it might be uncordoned later.
+            occupied_hosts.add(host)
```
epwalsh, line 85: can skip pulling jobs for the node below in this case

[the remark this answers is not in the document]
epwalsh: the type is `Optional[datetime]`, not `bool`, silly

epwalsh (approved):
LGTM, just one suggestion

src/olmo_core/launch/select_beaker_hosts.py
2015aroras: There's a `continue` on the next line that should skip the job pulling?

epwalsh: Oh, didn't see that for some reason

PR #377, opened 2025-09-18
Add `flash_3` attention backend

tyler-romero:
Benchmark:
```
S = 4k, no CP, on Augusta
Olmo2 7B
1 node/8 GPUs

torch SDPA:   10,164 TPS
flash-attn 2: 10,231 TPS
TE attn:      10,846 TPS
flash-attn 3: 11,066 TPS
```


src/olmo_core/nn/attention/backend.py
```diff
@@ -34,11 +38,17 @@ class AttentionBackendName(StrEnum):
     PyTorch's built-in SDPA ➡️ :class:`TorchAttentionBackend`
     """
     flash = "flash"
+    flash_2 = "flash"
```
tyler-romero: thoughts on supporting both names? I'd probably prefer just `flash_2`

src/olmo_core/testing/utils.py
```diff
@@ -11,19 +11,30 @@
...
+has_flash_attn_3 = False
 has_torchao = False
 has_grouped_gemm = False
 has_te = False
 
 try:
     import flash_attn  # type: ignore
 
-    has_flash_attn = True
+    has_flash_attn_2 = True
```
tyler-romero, line 23: good idea to rename?

src/olmo_core/nn/attention/backend.py
epwalsh: Let's go with just `flash_2` ✅

src/olmo_core/testing/utils.py
epwalsh: 👍

src/olmo_core/nn/attention/__init__.py
```diff
@@ -30,6 +30,7 @@
 from .backend import (
     AttentionBackend,
     AttentionBackendName,
+    FlashAttention3Backend,
     FlashAttentionBackend,
```
epwalsh: Rename to `FlashAttention2Backend?`

epwalsh (approved):
Looks good! Just a couple comments.

You might want to update the Beaker image we use for GPU tests so that the flash3 tests will run in CI: https://github.com/allenai/OLMo-core/blob/1aeb369c596b040fce3329e8f94807f74cdd8c55/.github/workflows/main.yml#L125

PR #378, opened 2025-09-18
Port task groups from cookbook to OlmoCore

tyler-romero:
Getting ready to get rid of this in cookbook: https://github.com/allenai/olmo-cookbook/blob/olmo3-anneals/src/cookbook/model/evaluators.py#L6

### DIFF
--- a/src/olmo_core/eval/task_groups.py
+++ b/src/olmo_core/eval/task_groups.py
@@ -1,0 +1,184 @@
+from typing import Dict, List
+
+# For training runs where we don't expect the model to acquire MC (e.g., 1B-5xC, short 7B training runs)
+FULL_TASKS_SMALL_COMPUTE = [
+    # OLMES Core 9(-ish) RC
+    "arc_challenge_test_rc_5shot",
+    "arc_easy_test_rc_5shot",
+    "hellaswag_rc_5shot",  # 1K subset of HellaSwag
+    "winogrande_val_rc_5shot",  # Helpful after 750M-5xC scale
+    "csqa_val_rc_5shot",
+    "piqa_val_rc_5shot",
+    "socialiqa_val_rc_5shot",
+    # Too noisy to be worth tracking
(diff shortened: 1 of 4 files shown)


src/olmo_core/eval/task_groups.py
```diff
@@ -0,0 +1,184 @@
+from typing import Dict, List
+
+# For training runs where we don't expect the model to acquire MC (e.g., 1B-5xC, short 7B training runs)
+FULL_TASKS_SMALL_COMPUTE = [
```
tyler-romero, line 4: I don't love this hardcoding of big lists but it is simple/explicit

src/olmo_core/eval/task_groups.py
```diff
@@ -0,0 +1,184 @@
...
+    "mt_mbpp_cpp_gold_bpb_3shot",
+    "mt_mbpp_java_gold_bpb_3shot",
+    "mt_mbpp_rust_gold_bpb_3shot",
+    # Sanity check for MCQA ability
+    "copycolors_10way_fast",
+]
+
+# NOTE: preserved for backwards compatibility, OLMO2_DEV_1B_TASKS is the same as
+# FULL_TASKS_SMALL_COMPUTE but with "copycolors_10way" instead of "copycolors_10way_fast"
+OLMO2_DEV_1B_TASKS = [
```
tyler-romero, line 138: This is soooo similar to `FULL_TASKS_SMALL_COMPUTE`, with just this one tweak. Worth preserving both for back compatibility?

undfined: Feels like we should just pick one. I'm not sure why we have non-fast version here but seems like FULL_TASKS_SMALL_COMPUTE should be the canonical.

tyler-romero: Ok cool, I'll get rid of this one then

PR #380, opened 2025-09-19
Improve distributed error handling

epwalsh:
I've seen a few scenarios recently where an error occurs before training begins, either during loading a checkpoint or in the `pre_train()` method of a callback, and instead of failing immediately the process will hang for the duration of the distributed timeout. It turns out that's because we were setting up our training scripts like this:

```python
try:
    train()
finally:
    teardown_distributed_environment()
```

So we'd end up trying to teardown the distributed process group from one rank (the first rank to raise an exception) while the others are doing something else, leading to the hang. A fix for this is to only call the teardown function when the training loop completes successfully:

```python
train()
teardown_distributed_environment()
```

lI tested this out with a callback that raises an exception on `pre_train()` and this seems to work well. The job ended within a few seconds after the error as opposed to 30 mins (the NCCL timeout).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - A configuration error is now raised if you pass in a URL for the trainer or dataset's working directory.
   Previously the URL would just get mangled into a local path, leading to unexpected behavior.
 - Fixed an issue where the `ConsoleLoggerCallback` would attempt to log before the first step.
+- Only call `teardown_distributed_environment()` when training ends cleanly to avoid a hang for the duration of the distributed backend's timeout when there's an error from one rank.
 
 ### Changed
 


src/olmo_core/train/trainer.py
```diff
@@ -659,29 +659,29 @@ def fit(self):
...
-
         # Install SIGTERM + SIGINT handlers.
         og_sigterm_handler = signal.signal(signal.SIGTERM, self._handle_os_signal)
         og_sigint_handler = signal.signal(signal.SIGINT, self._handle_os_signal)
 
-        # Do a dry-run for compiling and catch OOMs.
-        self._dry_run_batch()
-
         try:
+            for callback in self._iter_callbacks():
```
epwalsh, line 667: Moved this stuff down here so that `callback.on_error()` will be called if any of these steps fail.

tyler-romero (approved):
Nice!

PR #383, opened 2025-09-22
fix bug with masking out instances

epwalsh:


epwalsh:
@dirkgr came up with a better approach

PR #384, opened 2025-09-22
Add attention backend abstraction and integrate transformer engine's attention

epwalsh:
# Overview

This PR adds a new abstraction called `AttentionBackend` that encapsulates an SDPA kernel, making it easier to add new attention kernels. It also adds a TransformerEngine attention backend, which wraps TE's [DotProductAttention](https://docs.nvidia.com/deeplearning/transformer-engine/user-guide/api/pytorch.html#transformer_engine.pytorch.DotProductAttention). This gives us a nice speed-up even over flash-attn, both with and without context parallelism.

The attention backed is now configured with the `backend` argument to the `Attention` and `AttentionConfig` classes. Valid values are currently "torch", "flash", and "te". The `use_flash` argument is deprecated in favor of `backend`. And if neither is specified, we'll choose a compatible backend based on other configuration options.

So to use the TransformerEngine attention backend, just add the override `--model.block.attention.backend=te`.

# For reviewers

So far the commit history is clean and can be reviewed one-by-one, if that makes it easier.
- 4e622610fa38515ca3588c8c832b65eabf4f94e0 refactors the attention module to use the new `AttentionBackend` abstraction.
- 25256f488c34f523476b42dd262572bee9cfa279 adds TE to our Docker builds.
- 73195807511d43cb04c9abab0b942552d3394dfa adds a TE attention backend.

### DIFF
--- a/src/olmo_core/launch/beaker.py
+++ b/src/olmo_core/launch/beaker.py
@@ -96,12 +96,12 @@ class OLMoCoreBeakerImage(StrEnum):
     The stable image with CUDA pinned to 12.8.
     """
 
-    tch280_cu128 = "olmo-core-tch280cu128-2025-09-18"
+    tch280_cu128 = "olmo-core-tch280cu128-2025-09-19"
     """
     Built with torch 2.8.0 and CUDA 12.8.
     """
 
-    tch280_cu129 = "olmo-core-tch280cu129-2025-09-18"
+    tch280_cu129 = "olmo-core-tch280cu129-2025-09-23"
     """
     Built with torch 2.8.0 and CUDA 12.9.
     """


src/olmo_core/nn/attention/__init__.py
```diff
@@ -259,6 +329,337 @@ def build(
             ) from e
 
 
+class AttentionBackend(nn.Module):
```
2015aroras: nit: Can you put some of the new backend objects in a separate file called, say, `backend.py`?

src/olmo_core/nn/attention/__init__.py
```diff
@@ -259,6 +329,337 @@ def build(
...
+    def apply_cp(
+        self,
+        cp_mesh: DeviceMesh,
+        load_balancer: RingAttentionLoadBalancerType,
+        head_stride: int = 1,
+    ):
+        """
+        Apply context parallelism, if supported by the backend.
+        """
+        self.cp_pg = cp_mesh.get_group()
```
2015aroras: Maybe the default should be `NotImplementedError` or the error from `TorchAttentionBackend`. I imagine supporting CP should be optional for new backends.

src/olmo_core/nn/attention/__init__.py
```diff
@@ -903,17 +1239,24 @@ def __init__(
         clip_qkv: Optional[float] = None,
         dropout: float = 0.0,
         dtype: torch.dtype = torch.float32,
+        backend: Optional[AttentionBackendName] = None,
         init_device: str = "cpu",
         cache: Optional[BufferCache] = None,
     ):
         super().__init__()
 
+        if backend is not None and backend != AttentionBackendName.flash:
```
2015aroras: thought: we could have the backend say that it supports fused qkv (via a new method `supports_fused_kv()`) instead.

src/olmo_core/nn/attention/__init__.py
```diff
@@ -135,6 +143,67 @@ class AttentionType(StrEnum):
...
+        if self == AttentionBackendName.flash:
+            return has_flash_attn()
+        else:
+            return True
+
+    def supports_swa(self) -> bool:
+        if self in (AttentionBackendName.flash,):
+            return True
+        else:
+            return False
```
tyler-romero: A similar "supports kvcache" helper might be useful (flash only)

src/olmo_core/nn/attention/__init__.py
```diff
@@ -259,6 +329,337 @@ def build(
...
+        if not has_te_attn():
+            raise RuntimeError("Transformer Engine attention is not available")
+        assert TEDotProductAttention is not None
+        self.te_attn = TEDotProductAttention(
+            self.n_heads,
+            self.head_dim,
+            num_gqa_groups=self.n_kv_heads,
+            attention_dropout=self.dropout_p,
+            attn_mask_type="causal",
+            window_size=(self.window_size[0], 0),  # be explicit about causal mask
```
tyler-romero: So TE does support SWA? The rest of the code makes it seem like it does not

src/olmo_core/nn/attention/__init__.py
2015aroras: You could similarly add a `supports_context_parallelism()`.

2015aroras (commented):
Partial review for now

src/olmo_core/nn/attention/__init__.py
epwalsh: Done: 22831d39

src/olmo_core/nn/attention/__init__.py
epwalsh: Done: b5afa586

src/olmo_core/nn/attention/__init__.py
```diff
@@ -359,18 +364,31 @@ def __init__(
...
+        if window_size is not None:
             if window_size <= 0:
                 raise OLMoConfigurationError(f"'window_size' must be positive (got {window_size})")
-            # Flash attn window is [i - window_size[0], i + window_size[1]] inclusive
-            self.window_size = (window_size - 1, 0)
-        else:
-            self.window_size = (-1, -1)
+
+            if backend is None:
+                backend = AttentionBackendName.flash
```
tyler-romero, line 388: when we're auto-selecting a backend, can we log which one is chosen?

src/olmo_core/nn/attention/__init__.py
epwalsh: Updated in: 73867e4b

Changed these methods to `assert_supported()` or `assert_supports_cp()` so that each backend can provide more information when a dependency is missing.

src/Dockerfile
```diff
@@ -61,6 +61,9 @@ RUN pip install --no-build-isolation --no-cache-dir "${GROUPED_GEMM_VERSION}"
 ARG FLASH_ATTN_VERSION=2.8.2
 RUN pip install --no-build-isolation --no-cache-dir flash-attn==${FLASH_ATTN_VERSION}
 
+# Install transformer-engine.
+RUN pip install --no-build-isolation --no-cache-dir transformer_engine[pytorch]
```
tyler-romero: should we pin to a specific version? Most recent is 2.6: https://github.com/NVIDIA/TransformerEngine/releases/tag/v2.6

src/olmo_core/nn/attention/__init__.py
epwalsh: It's logged below on line 407

src/olmo_core/nn/attention/__init__.py
epwalsh: It does support SWA.

> The rest of the code makes it seem like it does not

How so?

src/olmo_core/nn/attention/__init__.py
epwalsh: Done: 6270d62b

src/Dockerfile
epwalsh: Done: 9729065e

src/test/nn/attention_test.py
```diff
@@ -40,6 +42,49 @@
...
 
+@pytest.mark.parametrize(
+    "window_size",
+    [
+        pytest.param((-1, -1), id="full"),
+        pytest.param((8, 8), id="SWA"),
+    ],
+)
+@pytest.mark.parametrize("n_kv_heads", [None])
+@pytest.mark.parametrize("n_heads", [8, 4])
```
2015aroras: Did you mean to do this to test GQA?
```suggestion
@pytest.mark.parametrize("n_kv_heads", [None, 4])
@pytest.mark.parametrize("n_heads", [8])
```

epwalsh: yeup, good catch!

PR #385, opened 2025-09-23
bnb style adamw implemenation (better numerical properties)

tyler-romero:


src/olmo_core/optim/adamw.py
```diff
@@ -36,9 +37,17 @@ def adamw_step(
...
-
-    denom = (exp_avg_sq.sqrt() / bias_correction2.sqrt()).add_(eps)
+    if bnb_style:
+        # https://github.com/bitsandbytes-foundation/bitsandbytes/blob/e8170363e0c94db8e5ef6a150fc7e1f9fc858602/bitsandbytes/backends/default/ops.py#L428
+        # https://x.com/Tim_Dettmers/status/1969133748382765407
+        # mathematically identical, but numerically more stable.
+        # avoids eps having an outsized effect when bc2_sqrt is tiny (early in training).
+        bc2_sqrt = bias_correction2.sqrt()
+        step_size = lr * bc2_sqrt / bias_correction1
+        denom = exp_avg_sq.sqrt() + (bc2_sqrt * eps)
```
epwalsh, line 47: Should this be:
```suggestion
        denom = (exp_avg_sq.sqrt() + (bc2_sqrt * eps)) / bc2_sqrt
```

tyler-romero: No actually, the update equation is rearranged as if both the numerator and denominator have been multiplied by `bc2_sqrt`. Which is why there is now a `bc2_sqrt` being multiplied by eps.

tyler-romero: Image

PR #381, opened 2025-09-19
Np Dataset Config Refactor

tyler-romero:
todo: think a bit deeper about backwards compat

PR #386, opened 2025-09-23
refactor internal experiment configuration assembler

tyler-romero:
Currently, our internal experiment config assembly process does not easily allow for modifying the underlying dataset / data mix. There was a recent workaround added for this (https://github.com/allenai/OLMo-core/commit/801de4be6091403267709e1cb99d958ec634b955) but its awkward to need to override *all* common components in order to construct a custom dataset.

This PR makes several changes to the internal API:
1.  `main(...)` now takes only a function that converts CLI args to an `ExperimentConfig`. This means that it is possible to fully define a custom `ExperimentConfig` and still leverage the common `launch`, `train`, `dry_run`, etc CLI interface.
2. `build_config()` has been pulled out of `main` and is now the default way to construct an `ExperimentConfig`. It has been refactored to support custom `dataset` and `dataloader` configuration.
3. `CommonComponents` is now more focused and holds only things that are actually "common" in some sense. in the old factorization, some components like the dataloader don't really make sense.
4. `max_sequence_length` no longer is redirected through the dataset config and out the other side. This allows decoupling the construction of the dataset from construction of the model/trainer. Now all of these override functions  can just take `CommonComponents` for a more unified API.

Corresponding scripts have been updated w/o changing actual configuration settings/functionality.


tyler-romero:
Looking for feedback on this refactor so I've only applied the changes to one downstream consumer of this api: `src/scripts/train/OLMo2-1B.py`.

If it looks good I'll apply it to the other ~20 files.

tyler-romero:
cc: @dirkgr

Similar to https://github.com/allenai/OLMo-core/pull/381, we should probably avoid merging this into the 32B branch for now.

PR #387, opened 2025-09-23
Fix unsupported configuration type in numpy_dataset

tyler-romero:
Fix an issue caused by trying to use `PathOrStr` in an OmegaConf config:
```
E         During handling of the above exception, another exception occurred:
E         
E         Traceback (most recent call last):
E           File "/Users/tylerr/workspace/OLMo-core/src/scripts/train/OLMo2-1B.py", line 106, in <module>
E             main(
E           File "/Users/tylerr/workspace/OLMo-core/src/olmo_core/internal/experiment.py", line 392, in main
E             config = build_config(
E           File "/Users/tylerr/workspace/OLMo-core/src/olmo_core/internal/experiment.py", line 299, in build_config
E             config = config.merge(overrides)
E           File "/Users/tylerr/workspace/OLMo-core/src/olmo_core/config.py", line 198, in merge
E             raise OLMoConfigurationError(str(e))
E         olmo_core.exceptions.OLMoConfigurationError: Unsupported value type: 'Union[pathlib.Path, os.PathLike, str]'
E             full_key: label_mask_paths
E             object_type=None
```

Add small test to help catch this type of issue in the future

### DIFF
--- a/src/olmo_core/internal/common.py
+++ b/src/olmo_core/internal/common.py
@@ -72,7 +72,6 @@ def _to_beaker_env_secret(
 
 
 def get_root_dir(cluster: str) -> str:
-    root_dir: str = "weka://oe-training-default/ai2-llm"
     if cluster in [
         "ai2/test-h100",
         "ai2/jupiter",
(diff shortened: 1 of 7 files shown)


src/test/scripts/train/OLMo2-1B_test.py
tyler-romero, line 1: This doesn't *exactly* follow the directory structure in the same way as other tests. To do that we would need to nest every test file one directory deeper. Thoughts?

epwalsh: I don't think we need to test these scripts. If anything we should have a unit test on the dataset config class.

tyler-romero: The actual part that failed is the override merging part. There are unit tests on the dataset config class but they dont interact with omegaconf

epwalsh: But you could test that with `DatasetConfig.merge(...)`

tyler-romero: Yeah thats fair enough. I'll add that test for this specific issue.

Why dont you think we should run a basic test for these scripts? It would catch a broader set of issues.

epwalsh: I guess I'm not against it in general, I just think most of the issues that would catch should already be caught by linters, mypy, or unit tests

PR #388, opened 2025-09-24
Add sliding window support for torch attention backend

2015aroras:
This PR adds sliding window support for the torch attention backend. Sliding window SDPA support will decrease increase validation failures in OLMo Core to HF conversion.

I ran a 1B and its 57K TPS with flash attention vs 37K with torch SDPA, so perf is bad.

As a bonus, the added SDPA test shows that our use of intra-document masking with the TE backend is incorrect.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed an issue where the `ConsoleLoggerCallback` would attempt to log before the first step.
 - Only call `teardown_distributed_environment()` when training ends cleanly to avoid a hang for the duration of the distributed backend's timeout when there's an error from one rank.
 - Fixed tensor parallelism issue with torch 2.8.
+- More fixes for Beaker cluster names.
 - `Callback.post_train()` will still be called even if the run is canceled before the dry-run batch.
 - `GarbageCollectorCallback` will restore `gc` settings even when `Trainer.fit()` exits on an error.
 
--- a/src/olmo_core/internal/common.py
+++ b/src/olmo_core/internal/common.py
@@ -19,6 +19,10 @@ from olmo_core.utils import generate_uuid
 
 log = logging.getLogger(__name__)
 
+GOOGLE_CLUSTERS = [
+    "ai2/augusta",
+]
+
 
 @lru_cache()
 def get_beaker_client() -> Optional[Beaker]:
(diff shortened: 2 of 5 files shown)


src/olmo_core/nn/attention/__init__.py
```diff
@@ -412,6 +412,7 @@ def __init__(
             scale=softmax_scale,
             dropout_p=dropout,
             window_size=window_size_tuple,
+            cache=cache,
```
epwalsh, line 415: Also should have this in the `FusedAttention` class

src/test/nn/attention_test.py
```diff
@@ -168,6 +169,129 @@ def test_attention(
...
+        pytest.skip("flash-attn requires a low precision dtype")
+    if (
+        backend_name in (AttentionBackendName.flash, AttentionBackendName.te)
+        and device.type == "cpu"
+    ):
+        pytest.skip(f"{backend_name} backend requires GPU")
+    if backend_name == AttentionBackendName.torch and intra_doc_masking:
+        pytest.skip("intra-document masking is not supported by torch backend")
+    if backend_name == AttentionBackendName.te and intra_doc_masking:
+        pytest.xfail("intra-document masking is currently broken in te backend")
```
epwalsh: Does this stop the test here (by raising an exception) or let it continue and fail on its own?

epwalsh (commented):
Overall looks good, just minor comments

src/test/nn/attention_test.py
2015aroras: Stops the test here.
https://docs.pytest.org/en/stable/reference/reference.html#pytest.xfail

src/olmo_core/nn/attention/__init__.py
2015aroras: https://github.com/allenai/OLMo-core/pull/388/commits/aa7278647270f90a2c694098454de0bd4d943c37 Done

src/test/nn/attention_test.py
epwalsh: In that case we should refactor so that the test can run. I usually do that by adding the xfail mark to the right parameter, but it's a combination of parameters here you could either (1) refactor the parametrized decorators to include both `backend_name` and `intra_doc_masking` in one decorator, or (2) catch the actual error that occurs below and raise the xfail in that case.

2015aroras: https://github.com/allenai/OLMo-core/pull/388/commits/bbe933d4ad101ae9598df4af4e7999986e30f11d I picked option 2. Easier to implement and easier to remove once the problem is fixed.

PR #393, opened 2025-09-24
Remove old beaker refs

2015aroras:
Some references to the old beaker cluster names were removed, this PR removes the rest. Also it makes `get_root_dir` error out on an invalid cluster, since this burnt me today.


tyler-romero (approved):
LGTM - there are always more!

PR #395, opened 2025-09-25
MPS bug fixes

2015aroras:
This PR addresses 2 issues I came across when running MPS locally and having tests fail.
- `get_default_device` tries to set the default device to MPS in distributed contexts, but MPS is not supported by any distributed backends.
- `move_to_device` is non-blocking when moving from CPU to MPS, but this leads to invalid results sometimes (no guarantees, unlike CUDA).


epwalsh (approved):
LGTM!

PR #404, opened 2025-10-01
Cookbook migration part 2 - long context config

tyler-romero:


tyler-romero:
Next step is to make the public/external versions. Should be straightforward.

src/olmo_core/nn/transformer/config.py
```diff
@@ -1132,3 +1132,39 @@ def ngpt_like(
...
+                i: add_rope_scaling(new_config.block.copy())
+                for i in range(new_config.n_layers)
+                if not (
+                    new_config.block.attention.sliding_window
+                    and new_config.block.attention.sliding_window.should_use_swa(
+                        i, new_config.n_layers
+                    )
+                )
+            }
+        return new_config
```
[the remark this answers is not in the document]
epwalsh: > ...which can incorrectly apply scaling to unintended blocks due to shared references

It's a deep copy so it should be okay.

> Additionally, when `full_attn_layers_only=True`, it unconditionally overwrites any existing `block_overrides`, discarding prior custom block configurations.

That's actually a good point. Could at least error out if block overrides is already set.

src/olmo_core/nn/transformer/config.py
```diff
@@ -1132,3 +1132,39 @@ def ngpt_like(
...
+            def add_rope_scaling(block: TransformerBlockConfig) -> TransformerBlockConfig:
+                rope_config = block.attention.rope
+                if rope_config is None:
+                    raise ValueError("Cannot apply RoPE scaling to a layer without RoPE.")
+                rope_config.scaling = rope_scaling
+                block.attention.rope = rope_config
+                return block
+
+            # Add rope scaling only to layers that do not use sliding window attention.
+            new_config.block_overrides = {
```
epwalsh: My other comment got hidden but I think we should error out if block overrides is already set.

soldni: that wouldn't be bad. maybe we can have error to indicate how one would need to override in case they really want to? 

there are corner cases where, for experimentation purposes, you want to change to see what happens. but it should be a high friction change, so maybe not through here, but CLI overrides.

src/scripts/train/OLMo3/OLMo3-7B-long-context.py
```diff
@@ -0,0 +1,131 @@
...
+    )
+
+    tokenizer_config = TokenizerConfig.dolma2()
+    model_config = TransformerConfig.olmo3_7B(
+        vocab_size=tokenizer_config.padded_vocab_size()
+    ).with_rope_scaling(
+        YaRNRoPEScalingConfig(factor=8, beta_fast=32, beta_slow=1, old_context_len=8192)
+    )
+
+    max_tokens = 50_000_000_000  # 50B
```
soldni: nit: maybe move this to top alongside the other hardcoded hparams.

src/scripts/train/OLMo3/OLMo3-7B-long-context.py
```diff
@@ -0,0 +1,131 @@
...
+        num_nodes=32,
+        nccl_debug=True,
+        # override priority from the CLI eg `--launch.priority=high`
+    )
+
+    tokenizer_config = TokenizerConfig.dolma2()
+    model_config = TransformerConfig.olmo3_7B(
+        vocab_size=tokenizer_config.padded_vocab_size()
+    ).with_rope_scaling(
+        YaRNRoPEScalingConfig(factor=8, beta_fast=32, beta_slow=1, old_context_len=8192)
```
soldni, line 51: The `old_context_len=8192` made me think of something: a pattern we have in cookbook is, in case we are restoring a run, is to load old config from checkpoint dir and pick existing parameters from there. 

Maybe not specifically for this script, but it would be cool to have an example that shows how to do that. Or maybe a util somewhere in `olmo_core.internal.cookbook` to make that easier. 

just a thought, maybe for future PR.

src/scripts/train/OLMo3/OLMo3-7B-long-context.py
```diff
@@ -0,0 +1,131 @@
...
+    )
+
+    max_tokens = 50_000_000_000  # 50B
+    max_duration = Duration.tokens(max_tokens)
+
+    train_module_config: TransformerTrainModuleConfig = cookbook.configure_train_module(
+        max_sequence_length=SEQ_LENGTH,
+        rank_microbatch_size=SEQ_LENGTH,
+        learning_rate=0.00020712352850360292,
```
soldni: same as above, I would move to hardcoded experiment params at the top.

src/scripts/train/OLMo3/OLMo3-7B-long-context.py
```diff
@@ -0,0 +1,131 @@
...
+            join_path(
+                root_dir, "preprocessed/tylerr/lc-reshard-final/v0.6/allenai/dolma2-tokenizer/*.npy"
+            )
+        ),
+        tokenizer=tokenizer_config,
+        work_dir=work_dir,
+        sequence_length=SEQ_LENGTH,
+        generate_doc_lengths=True,  # enables intra-document masking
+        source_group_size=8,
+        source_permutation_seed=123,
```
soldni, line 76: Is there a reason why this seed is different from other seed?

src/scripts/train/OLMo3/OLMo3-7B-long-context.py
```diff
@@ -0,0 +1,131 @@
...
+
+    data_loader_config = NumpyDataLoaderConfig(
+        global_batch_size=GLOBAL_BATCH_SIZE, seed=SEED, num_workers=4
+    )
+
+    trainer_config = cookbook.configure_trainer(
+        load_path=str(
+            join_path(
+                root_dir,
+                "checkpoints/allysone/anneal-round5-100B-olmo25_7b-anneal-6T-decon-sparkle-motion-8730626c/step47684",
```
soldni, line 87: move to top

soldni (approved):

src/scripts/train/OLMo3/OLMo3-7B-long-context.py
tyler-romero: Just how it was in cookbook for the actual LC experiment, no fundamental reason

src/olmo_core/nn/transformer/config.py
tyler-romero: This is just a convenience method, it would be easy to work around if you want to do something really custom.

PR #412, opened 2025-10-04
Add Dolma 3 sample, and a way to alias data mixes

dirkgr:


src/olmo_core/data/mixes/__init__.py
```diff
@@ -40,6 +40,26 @@ class DataMix(DataMixBase):
     v3_small_ppl_validation = "v3-small-ppl-validation"
     OLMo_mix_0625 = "OLMo-mix-0625"
     OLMo_mix_0925 = "OLMo-mix-0925"
+    OLMo_mix_0625_150Bsample = "OLMo-mix-0625-150Bsample"
```
tyler-romero, line 43: Alternatively could add more values to the enum that map to the same string, eg:
```
OLMo_mix_0625_150Bsample = "OLMo-mix-0625-150Bsample"
dolma3-0925-150B-mix = "OLMo-mix-0625-150Bsample"
```

dirkgr: You can't have a member named "dolma3-0925-150B-mix" because of the minus signs.

src/olmo_core/data/mixes/__init__.py
```diff
@@ -40,6 +40,26 @@ class DataMix(DataMixBase):
...
+
+    @classmethod
+    def _missing_(cls, value: object) -> "DataMix | None":
```
epwalsh, line 46: TIL about this!

PR #415, opened 2025-10-08
olmo3 conversion

tyler-romero:


src/examples/huggingface/convert_checkpoint_to_hf.py
```diff
@@ -77,45 +78,61 @@ def convert_checkpoint_to_hf(
...
-            backend = AttentionBackendName.flash_2
-
-        assert backend in (
-            AttentionBackendName.flash_2,
-            AttentionBackendName.flash_3,
-        ), "flash_2 or flash_3 backend is expected for fused attention"
-
-        try:
-            backend.assert_supported()
+    block_entries: list[tuple[str, TransformerBlockConfig]] = [("base block", model_config.block)]
```
2015aroras, line 85: Could you add these changes to `src/examples/huggingface/convert_checkpoint_from_hf.py` too? Should hopefully be almost copy-paste

src/olmo_core/nn/hf/config.py
```diff
@@ -1,8 +1,9 @@
-from transformers import Olmo2Config, PretrainedConfig
+from transformers import Olmo2Config, Olmo3Config, PretrainedConfig
```
2015aroras: Could you import and use `Olmo3Config` like it's done for `FlexOlmoConfig`? Then we don't need to constraint the transformers version.

```
try:
    from transformers import Olmo3Config  # type: ignore
except ImportError:
    Olmo3Config = None
```

src/examples/huggingface/convert_checkpoint_from_hf.py
```diff
@@ -50,6 +51,7 @@ def _get_transformer_config(model_arch: str, vocab_size: int) -> TransformerConf
         "olmo2_7b": TransformerConfig.olmo2_7B,
         "olmo2_13b": TransformerConfig.olmo2_13B,
         "olmo2_32b": TransformerConfig.olmo2_32B,
+        "olmo3_7b": TransformerConfig.olmo3_7B,
```
2015aroras, line 55: Could you add the 190M too? No harm in doing it afaik

src/test/examples/huggingface/convert_checkpoint_from_hf_test.py
```diff
@@ -7,55 +7,74 @@
...
+        "num_attention_heads": 12,
+        "rope_theta": 500_000,
+        "rms_norm_eps": 1e-6,
+    }
+
+    if model_family == "olmo2":
+        hf_config = Olmo2Config(**common_config)
+    elif model_family == "olmo3":
+        pytest.skip("The installed transformers version does not support Olmo3")
+        hf_config = Olmo3Config(**common_config)
```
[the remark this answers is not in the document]
2015aroras: +1 to this. Do you need an if conditional here?

PR #417, opened 2025-10-09
Flex Attention support with sinks

aman-17:
- Added flex_attn integration in Attention class
- Updated sdpa method to handle flex attention with BlockMask
- Integrated FlexAttention for GPT-OSS type Attention Sinks


src/olmo_core/nn/attention/__init__.py
```diff
@@ -162,6 +166,8 @@ class AttentionConfig(Config):
     qk_norm: Optional[LayerNormConfig] = None
     dropout: Optional[float] = None
     use_flash: Optional[bool] = None
+    use_flex: Optional[bool] = None
```
2015aroras, line 169: Adding `use_*` is now a deprecated pattern for adding attention implementations. Instead you should create a new attention backend (see https://github.com/allenai/OLMo-core/blob/main/src/olmo_core/nn/attention/backend.py)

PR #418, opened 2025-10-10
Script to dump training tokens

dirkgr:


src/olmo_core/nn/attention/__init__.py
```diff
@@ -397,7 +397,7 @@ def __init__(
             if window_size <= 0:
                 raise OLMoConfigurationError(f"'window_size' must be positive (got {window_size})")
 
-            if backend is None:
+            if backend is None and torch.cuda.is_available():
```
2015aroras: May as well check if we have flash attention 2 available.

```suggestion
            if backend is None and has_flash_attn_2():
```

src/scripts/dump_training_batch.py
```diff
@@ -0,0 +1,288 @@
...
+        return False
+
+    for i, (actual, expected) in enumerate(zip(data_paths, mix_paths)):
+        if normalize_path(actual) != normalize_path(expected):
+            log.error(f"Path mismatch at index {i}:\n  Actual:   {actual}\n  Expected: {expected}")
+            return False
+    return True
+
+
+def NumpyFSLDataLoader_getitem_monkeypatch(self: NumpyFSLDataLoader, index: int) -> Dict[str, Any]:
```
epwalsh: Instead of monkey-patching this here, how about adding a new method like this to `NumpyFSLDataLoader`, like `.threaded_get()` or something? That way we don't have to worry about recreating the global indices logic here.

dirkgr: Can we do this more elegantly? Can we detect in `__getitem__` that we don't have a process group, and in that case go threaded? So that we don't have to keep the rest of the logic twice.

epwalsh: Yeup I like that idea. We know when we're in a worker when the `NumpyFSLDataLoader` gets wrapped by this class: https://github.com/allenai/OLMo-core/blob/73543e9e60aa52a03086f1c06cd3926e492f4251/src/olmo_core/data/data_loader.py#L981-L983

So that wrapper class could set a field on the `NumpyFSLDataLoader` to tell it that it's in a worker, and otherwise we could use the threaded get.

dirkgr: Do we need a field? We have `self.num_workers` right in `NumpyFSLDataLoader`.

dirkgr: It also has `self.num_threads` 🤔

dirkgr:
@epwalsh , can you check this solution?

src/olmo_core/data/data_loader.py
```diff
@@ -673,7 +674,14 @@ def __getitem__(self, index: int) -> Dict[str, Any]:
             indices = indices[:, self.dp_rank :: self.dp_world_size]
 
         # Get instances for the batch.
-        instances = [self._get_dataset_item(int(idx)) for idx in indices[index]]
+        map_fn: Callable
+        if self.num_threads is not None and self.num_threads > 1:
```
epwalsh: I forgot we can just check `self.worker_info`:
```suggestion
        if self.worker_info is None and self.num_threads is not None and self.num_threads > 1:
```

Without this additional check on `worker_info` we might end up "doubling" the thread usage while training. See below in `_IterableDatasetWrapper.__iter__()`.

PR #424, opened 2025-10-16
handle nan losses/gnorms in the `SkipStepOptimizer`, add option to skip ranges of steps

epwalsh:


dirkgr:
We did this in a different way.

PR #425, opened 2025-10-16
Add option to skip ranges of steps in the trainer

epwalsh:


src/olmo_core/train/trainer.py
```diff
@@ -1325,16 +1332,26 @@ def _fit_epoch(self):
...
+                for callback in self._iter_callbacks():
+                    callback.pre_optim_step()
 
-            self.train_module.optim_step()
-            self.train_module.zero_grads()
+                self.train_module.optim_step()
+                self.train_module.zero_grads()
 
             for callback in self._iter_callbacks():
                 callback.post_train_batch()
```
2015aroras, line 1357: Feels kinda odd to not skip this when we're skipping the batch, but since checkpointing happens here I guess we don't have a choice.

PR #428, opened 2025-10-21
clean up checkpointing code to minimize distributed communication

epwalsh:
We're consistently running into failures while checkpointing on Augusta (this is synchronous checkpointing since GLOO has been an issue), and the point of failure is often a `barrier` right before we actually write the checkpoint. We're still investigating _why_ we're failing at a barrier, but we hope that reducing the number of barriers (and distributed communication in general) will reduce the probability of failure.

So this PR:
- Reduces the number of `barrier` calls prior to writing the checkpoint.
- Reduces the amount of communication required for subsequent trainer checkpoints by caching the global save plan.

### DIFF
--- a/src/olmo_core/distributed/checkpoint/__init__.py
+++ b/src/olmo_core/distributed/checkpoint/__init__.py
@@ -77,6 +77,7 @@ def save_state_dict(
     save_overwrite: bool = False,
     thread_count: Optional[int] = None,
     throttle_uploads: bool = False,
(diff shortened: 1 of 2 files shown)


dirkgr (approved):
Do we ever not want to cache the plan?

epwalsh:
> Do we ever not want to cache the plan?

I guess if you were saving tons of different tiny state dicts for some reason.

PR #431, opened 2025-10-23
Send a Slack notification when a Beaker job appears to be stuck

epwalsh:
> [!NOTE]
> Adds step inactivity detection while following Beaker jobs with Slack warnings (and optional hard stop), plus wires a default soft-timeout from internal launcher.
> 
> - **Beaker launch/follow (`src/olmo_core/launch/beaker.py`)**:
>   - Add `step_timeout` (hard stop) and `step_soft_timeout` (Slack warning) options to `BeakerLaunchConfig.launch()` and `follow_experiment()`.
>   - Detect training/eval steps via `_STEP_REGEX`; track last step time to flag inactivity or stop runs.
>   - Send new Slack "inactive" notification via `_send_slack_notification_for_event()`.
>   - Refactor log streaming to a background thread + `Queue` for non-blocking monitoring.
> - **Internal launcher (`src/olmo_core/internal/experiment.py`)**:
>   - Pass a default `step_soft_timeout=10*60` in `launch()` (hard `step_timeout` commented for optional use).
> - **Changelog**:
>   - Note Slack notification when a Beaker job appears stuck.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2c1c1dbba01228f6a212b9c93d806c99cb3e914a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/olmo_core/launch/beaker.py
+++ b/src/olmo_core/launch/beaker.py
@@ -751,7 +751,7 @@ def follow_experiment(
             cur_time = time.monotonic()
 
             # If (a) we've detected training steps already or (b) the run has been up for over 30 min,
-            # then we warn if we haven't detected new steps with the past `step_soft_timeout` seconds.
+            # then we warn if we haven't detected new steps within the past `step_soft_timeout` seconds.
             # But we only send a warning at most once per hour.
             if (
                 slack_webhook_url is not None
@@ -766,7 +766,7 @@ def follow_experiment(
                 last_inactivity_warning = cur_time
 
             # If (a) we've detected training steps already or (b) the run has been up for over 60 min,
-            # then we kill the job if we haven't detected new steps with the past `step_timeout` seconds.
+            # then we kill the job if we haven't detected new steps within the past `step_timeout` seconds.
             if (
                 step_timeout is not None
                 and (first_step_detected or (cur_time - start_time) > max(step_timeout, 3600))


dirkgr (approved):
I think this won't work. Or a least, it won't work the way we think it will. When we have our hangs, one rank stops printing things completely, but not usually rank0. The other ones print escalating warnings. So we would never see any silence in the logs (though eventually the escalating warnings come so far apart that it would trigger after all).

Still better than nothing, I guess, but how about this? We put in some regexes, and if any of those regexes match, it prints to the channel? This way we can detect the NCCL warnings quickly.

epwalsh:
Hmm or just decrease to 5min? Looks like that would catch things early. But if there's one thing I would grep for in the logs, I think it would be the messages like this: `NCCL WARN num_scheduled_reqs:1128710, num_completed_reqs:1128702`. What do you think?

dirkgr:
There is another message that actually says that it's been waiting for 5 minutes. We should match that one.

In the failures we're seeing, there seem to be two types of messages actually. There is one type where it prints a warning every 5 minutes until failure. And there is another where it doubles the amount of time it waits a few times, until it says "DEADLINE EXCEEDED" and then it just sits there. I'm not sure the structure of the two messages is different though. Either way, these are the messages that express their wait times in nanoseconds.

epwalsh:
Or we raise the alarm when we haven't seen our typical step log messages after 5/10 minutes.

```
2025-10-23T20:14:43.988Z [2025-10-23 20:14:43] INFO     [olmo_core.train.callbacks.console_logger:60, rank=0] [step=485999/707039,epoch=1]
```

I think that would be more robust.

dirkgr:
I like that. How long does checkpointing + evals take? Take that number, + 50%, and that's our timeout?

epwalsh:
Checkpointing takes ~4 minutes, evals take ~1 minute. Let's say 10 minute timeout to wait. But since it takes longer than that for the run to start, I say we postpone that timeout until either the first step is detected or an hour has gone by, whichever happens first.

dirkgr:
👍🏻

dirkgr:
Actually, since this just warns, and doesn't kill the job, make the initial timeout 30 minutes? Then it can say "Startup taking longer than 30 minutes! Alarm!"

dirkgr:
But once we get this dialed in, I absolutely want to kill the job with it.

epwalsh:
Done: 5ef543e0

epwalsh:
So I added a separate option for a hard timeout which would kill the job. I have it disabled right now, but we could also just set it to a relatively long time, like 2 hours.

PR #435, opened 2025-10-29
Allow manual bypass of fingerprint mismatch when switching datasets

baileykuehl:
Because some of the 7B pretraining mix files will not be available on R2 (namely, olmOCR pdfs), it is expected that users will get a `FileNotFoundError` when training with OLMo-mix-0625 from the official training script. We want to allow them to continue training with the OLMo-mix-0925 dataset, but this currently will fail due to a fingerprint mismatch. I want to:

1. Add an override flag to allow users to bypass the fingerprint mismatch
2. Add messages to alert users that this is an option when they come across `FileNotFoundError`

---

> [!NOTE]
> Introduce a configurable bypass for dataset fingerprint mismatches and add targeted guidance when OLMo-mix-0625 files are missing.
> 
> - **Data loading**:
>   - Add `ignore_fingerprint_mismatch` to `NumpyDataLoaderBase` (ctor, `wrap_numpy_dataset`) and `NumpyDataLoaderConfig`; propagate through builder.
>   - Update `load_state_dict()` to optionally bypass dataset fingerprint mismatches and log warnings.
> - **Dataset/mix path resolution**:
>   - Improve `FileNotFoundError` messages in `NumpyDatasetConfig._expand_globs()` and `SourceMixtureConfig.resolved_paths` to guide users off OLMo-mix-0625, suggesting OLMo-mix-0925 and the fingerprint override flag.
> - **Tests**:
>   - Add `test_fingerprint_override` covering both error and success paths with `ignore_fingerprint_mismatch`.
> - **Changelog**:
>   - Document new flag and enhanced error guidance for OLMo-mix-0625.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 63fbef1abb341d0a7fb8715c6bfc03394b689995. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/olmo_core/data/data_loader.py
+++ b/src/olmo_core/data/data_loader.py
@@ -442,7 +442,7 @@ class NumpyDataLoaderBase(TextDataLoaderBase):
                 raise RuntimeError(
                     "Dataset fingerprint does not match the fingerprint in the checkpoint, "
                     "set ignore_fingerprint_mismatch=True to ignore this error. "
-                    "Note that this will change the data order!"
+                    "This will probably result in a different data order!"
                 )
             log.warning(
                 "Fingerprint mismatch ignored since ignore_fingerprint_mismatch=True. "
--- a/src/olmo_core/data/numpy_dataset.py
+++ b/src/olmo_core/data/numpy_dataset.py
@@ -2399,7 +2399,7 @@ class NumpyDatasetConfig(Config, ABC):
                         "If you are resuming training from a checkpoint that used mix-0625, you will need to "
                         "switch to a newer mix such as OLMo-mix-0925. To continue training with a different "
                         "dataset mix, set 'ignore_fingerprint_mismatch=True' in your NumpyDataLoaderConfig "
-                        "to bypass the fingerprint mismatch error. Note that this will change the data order!"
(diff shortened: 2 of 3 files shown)


src/olmo_core/data/data_loader.py
```diff
@@ -434,8 +438,15 @@ def load_state_dict(self, state_dict: Dict[str, Any]):
...
-                "Restoring state from a different dataset is not supported! (fingerprint doesn't match)"
+            if not self.ignore_fingerprint_mismatch:
+                raise RuntimeError(
+                    "Dataset fingerprint does not match the fingerprint in the checkpoint, "
+                    "set ignore_fingerprint_mismatch=True to ignore this error. "
+                    "Note that this will change the data order!"
+                )
+            log.warning(
+                "Fingerprint mismatch ignored since ignore_fingerprint_mismatch=True. "
+                "Warning: The data order will be different!"
```
epwalsh: I guess it's not guaranteed the order will be different, so I'd suggest:
```suggestion
                "This will probably result in a different data order!"
```

baileykuehl: Fixed! I also updated the other warnings to be consistent with that wording

PR #438, opened 2025-10-30
Add gradient dumping to GAPMonitorCallback

farhatkevin:
## Summary

- Consolidates the standalone `GradientDumperCallback` into `GAPMonitorCallback` as an optional feature controlled by `dump_gradients=True`.
- Adds `dump_gradients_*` configuration fields (start_step, end_step, step_interval, save_first_n) to control when and how gradients are saved.
- Supports two modes: full distributed checkpoint saving, or preview mode (`dump_gradients_save_first_n`) that gathers and saves only the first N elements per parameter to rank 0.
- Adds `persist_working_subdir` to `Trainer` for uploading working directory subdirectories to the save folder.
- Deletes the standalone `GradientDumperCallback`, its example directory, training script, and tests.

### Design note for reviewers

Currently, `dump_gradients=True` requires `enabled=True` to work — gradient dumping is gated behind the same `enabled` flag that controls the full GAP monitoring stack (forward/backward hooks, per-tensor stats computation). This means you can't dump gradients without also paying the cost of monitoring.

We kept it this way to stay consistent with how other callbacks in the codebase use `enabled` as a master switch. However, an alternative would be to decouple these: rename `enabled` to something like `monitor` and let `dump_gradients` work independently. This would allow a `monitor=False, dump_gradients=True` configuration for jobs that only need gradient snapshots without the per-step stats overhead.

Would appreciate input on whether decoupling would be useful based on how this callback is expected to be used.


src/examples/gradient_dumping/README.md
```diff
@@ -0,0 +1,264 @@
...
+gradients = torch.load("/tmp/reconstructed_gradients/step2_all_gradients.pt")
+
+# Load metadata
+with open("/tmp/reconstructed_gradients/step2_metadata.json") as f:
+    metadata = json.load(f)
+
+# Print gradient norms
+for name, norm in sorted(metadata["parameter_norms"].items(), 
+                         key=lambda x: x[1], reverse=True)[:10]:
+    print(f"{name}: {norm:.6e}")
```
dirkgr: The gradient dumper also computes norms?

dirkgr (changes requested):
Haven't gone through the whole thing yet, but two quick comments:
* Can you save the output in safetensors format? Much faster.
* Can you add an option to only keep the first n values of each tensor? We're thinking of the situation where we want to use this to estimate gradient noise. For that, it's much more useful to have a few values and many steps, than many values for few steps.

src/examples/gradient_dumping/README.md
farhatkevin: this happens only in the reconstruction script (which is run if we save separate shards of each parameter) that will happen offline and not during gradient dumping. I thought this may be helpful for a pulse check but can remove if we don't want it.

farhatkevin:
Gather full gradient to rank 0 and save only the first N elements along the shard dimension. This creates a "preview" by slicing along the shard dim (usually dim 0). We can either choose to save first N in which this gathering will occur, or save all gradients in separate shards and later use the reconstruction script to recreate the full tensor, depending on whether `save_first_n` is passed in to the gradient dumper.

src/olmo_core/train/callbacks/gradient_dumper.py
```diff
@@ -0,0 +1,109 @@
...
+
+        step_dir = output_dir / f"step{self.step}"
+        step_dir.mkdir(exist_ok=True, parents=True)
+
+        assert hasattr(self.trainer.train_module, "model")
+        model = getattr(self.trainer.train_module, "model")
+
+        if self.save_first_n is None:
+            # Save sharded gradients using distributed checkpoint
+            sharded_dir = step_dir / "sharded"
```
epwalsh: Technically these gradients might not be sharded, like if we're training with DDP instead of FSDP. So maybe rename this dir to just `step_dir / "gradients"` or something.

src/olmo_core/train/callbacks/gradient_dumper.py
```diff
@@ -0,0 +1,109 @@
...
+
+            for name, p in model.named_parameters():
+                if p.grad is None:
+                    continue
+
+                # get_full_tensor handles DTensor all-gather automatically
+                full_grad = get_full_tensor(p.grad.detach())
+
+                # Only rank 0 saves
+                if not is_distributed() or get_rank() == 0:
```
epwalsh: `get_rank()` will return 0 when not distributed
```suggestion
                if get_rank() == 0:
```

src/olmo_core/train/callbacks/gradient_dumper.py
```diff
@@ -0,0 +1,109 @@
...
+            # Save first N elements of full gradients, gather all gradients and save from rank 0
+            sampled_dir = step_dir / "sampled"
+            sampled_dir.mkdir(exist_ok=True, parents=True)
+
+            for name, p in model.named_parameters():
+                if p.grad is None:
+                    continue
+
+                # get_full_tensor handles DTensor all-gather automatically
+                full_grad = get_full_tensor(p.grad.detach())
```
epwalsh: It can help save memory to manually `del full_grad` at the bottom of this block. You should also run garbage collection at the end of this method: `olmo_core.utils.gc_cuda()`.

epwalsh (commented):
Just a comment minor suggestions on the callback itself. @farhatkevin and I also discussed putting this functionality into the `GAPMonitorCallback` (https://github.com/allenai/OLMo-core/pull/446), which I think is a good idea. We could also generalize it to save activations.

src/olmo_core/train/callbacks/gradient_dumper.py
farhatkevin: added, thanks!

src/olmo_core/train/callbacks/gradient_dumper.py
farhatkevin: done

src/olmo_core/train/callbacks/gradient_dumper.py
farhatkevin: without `save_first_n` will now save to `step_dir/full_gradients`
with `save_first_n` will now save to `step_dir/sampled_gradients`

src/olmo_core/train/callbacks/gradient_dumper.py
```diff
@@ -0,0 +1,114 @@
...
+                del full_grad
+
+            log.info(f"Saved sampled gradients for step {self.step} to '{sampled_gradients_dir}'")
+
+        # Persist directory to save_folder if different from work_dir
+        # In distributed mode, only rank 0 uploads to avoid conflicts
+        rel_step_dir = step_dir.relative_to(self.trainer.work_dir)
+        target_dir = join_path(self.trainer.save_folder, rel_step_dir)
+        if str(step_dir) != str(target_dir):
+            if not is_distributed() or get_rank() == 0:
```
epwalsh:
```suggestion
            if get_rank() == 0:
```

src/examples/gradient_dumping/README.md
farhatkevin: since we no longer save shards like we did before we do not need a reconstruction script at all.

src/olmo_core/train/callbacks/gradient_dumper.py
```diff
@@ -0,0 +1,114 @@
...
+                log.info(f"Uploading gradients for step {self.step} to '{target_dir}'...")
+                copy_dir(
+                    step_dir, target_dir, save_overwrite=self.trainer.save_overwrite, quiet=True
+                )
+                log.info(f"Gradients for step {self.step} uploaded to '{target_dir}'")
```
epwalsh: We have a similar method on the trainer for individual files (`Trainer.persist_working_file()`). It would be nice to make this into a method as well, like `Trainer.persist_working_subdir()`.

src/olmo_core/train/trainer.py
```diff
@@ -1056,6 +1064,28 @@ def persist_working_file(self, name: PathOrStr) -> PathOrStr:
...
+        :raises FileExistsError: If any files in the directory already exist in the save folder and :data:`save_overwrite`
+            is ``False``.
+        """
+        if Path(name).is_relative_to(self.work_dir):
+            name = Path(name).relative_to(self.work_dir)
+        source = join_path(self.work_dir, name)
+        target = join_path(self.save_folder, name)
+        if source != target:
+            copy_dir(source, target, save_overwrite=self.save_overwrite)
+        elif dir_is_empty(source):
```
farhatkevin, line 1166: `dir_is_empty()` returns True when the directory is empty or does not exist, which matches our requirement since the directory should exist and not be empty if `source == target`. I considered creating a `dir_exists()` function to align better with `persist_working_file()`'s implementation, but 1) this function should raise `FileNotFoundError` in both cases (if the directory is missing or exist but is empty), and 2) If i were to create a standalone `dir_exists()` function in `io.py`, I think it should return True when the dir exists but is empty, which we don't want here.

src/olmo_core/train/callbacks/gradient_dumper.py
farhatkevin: agree, added `persist_working_subdir().`

src/olmo_core/train/trainer.py
epwalsh: I agree this handles both cases but it would be nice to get a more informative error message, depending on the case. See my suggestion in the other comment.

src/olmo_core/train/trainer.py
```diff
@@ -1056,6 +1064,28 @@ def persist_working_file(self, name: PathOrStr) -> PathOrStr:
...
+            raise FileNotFoundError(source)
```
epwalsh: Since the working directory is guaranteed to be a local directory, we can use `pathlib.Path.is_dir()`:
```suggestion
        elif not Path(source).is_dir():
            raise FileNotFoundError(f"'{source}' is not a directory")
        elif dir_is_empty(source):
            raise FileNotFoundError(f"source directory '{source}' is empty")
```

src/olmo_core/train/callbacks/gradient_dumper.py
```diff
@@ -0,0 +1,111 @@
...
+
+            if get_rank() == 0:
+                log.info(
+                    f"Saved sampled gradients for step {self.step} to '{sampled_gradients_dir}'"
+                )
+
+        if get_rank() == 0:
+            rel_step_dir = step_dir.relative_to(self.trainer.work_dir)
+            target_dir = self.trainer.persist_working_subdir(rel_step_dir)
+            log.info(f"Gradients for step {self.step} uploaded to '{target_dir}'")
```
epwalsh: The save folder could be local.
```suggestion
            log.info(f"Gradients for step {self.step} saved to '{target_dir}'")
```

src/olmo_core/train/trainer.py
farhatkevin: Could the error message "{source} is not a directory" be clarified for the case where `source` is a path that doesn't exist? Should we first check for whether it exists and return a different error message if it does not exist? Or do you think that "{source} is not a directory" is sufficient?

```suggestion
        elif not Path(source).exists():
            raise FileNotFoundError(f"Source '{source}' does not exist")
        elif not Path(source).is_dir():
            raise FileNotFoundError(f"Source '{source}' exists but is not a directory")
        elif dir_is_empty(source):
            raise FileNotFoundError(f"Source directory '{source}' is empty")
```

epwalsh (commented):
> We kept it this way to stay consistent with how other callbacks in the codebase use enabled as a master switch. However, an alternative would be to decouple these: rename enabled to something like monitor and let dump_gradients work independently. This would allow a monitor=False, dump_gradients=True configuration for jobs that only need gradient snapshots without the per-step stats overhead.

I like having separate flags, but also a main `enabled` flag that would turn both off. That would keep backwards compatibility as well.

src/olmo_core/train/callbacks/gap_monitor.py
```diff
@@ -1,45 +1,83 @@
...
+    def __post_init__(self):
+        if self.dump_gradients_step_interval <= 0:
+            raise ValueError(
+                f"dump_gradients_step_interval must be positive, got {self.dump_gradients_step_interval}"
+            )
+        if self.dump_gradients_save_first_n is not None and self.dump_gradients_save_first_n <= 0:
+            raise ValueError(
+                f"dump_gradients_save_first_n must be positive, got {self.dump_gradients_save_first_n}"
+            )
+        if not self.enabled and self.dump_gradients:
```
farhatkevin: Added validation for the new flag hierarchy. Two things to note:                                                                                                               
                                                                                                                                                                                 
  1. When `enabled=False`, we only warn if `dump_gradients=True ` since that's explicitly set by the user. We don't warn about `monitor` because it defaults to `True` — otherwise every  
  `GAPMonitorCallback(enabled=False)` would trigger a warning. Should we have this check at all, or is it expected that the user knows they must have `enabled = True`?                                                                                                          
  2. `enabled=True` with both `monitor=False` and `dump_gradients=False` is a hard error since the callback would do nothing.

  Let me know if you'd prefer different behavior for either case.

epwalsh: One small change here...

I would make `monitor` and `dump_gradients` optional bools. So you'd have:

```python
enabled: bool = True
monitor: Optional[bool] = None
dump_gradients: Optional[bool] = None
```

- Hard error if `enabled` is false and either of the other two flags are true.
- Hard error if `enabled` is true and both other flags are false.

farhatkevin: That makes sense, will do. A few questions on the resolution behavior:    
                                                                                           
  1. **How important is backwards compatibility here?** If we go fully explicit (`None`    
  resolves to `False`), existing `GAPMonitorCallback(enabled=True)` configs would need to  
  become `GAPMonitorCallback(enabled=True, monitor=True)`. Is that fine?

  2. **Should we explicitly resolve `None` to `False`?** Leaving it as `None` technically
  works since `None` is falsy, but usually better to in case later we check `if self.monitor is False`.

  3. **Two options for resolving `None`:**
     - **Strict**: `None` → `False`. User must always be explicit, which breaks existing configs.
     - **Backwards compatible**: When only `enabled=True` is set, default `monitor=True` to keep existing behavior. Otherwise just use the flags that are passed in as is. 
     
Which one should we go for here? I like the strict option better to be more explicit, but not sure how important it is to keep exact same behavior as already exists with `enabled=True`.

epwalsh: Keeping backwards compatibility should be the main concern. You could change the `None` values to the sensible default `bool` values in the `__post_init__` method. E.g.:

```python
def __post_init__(self):
    if self.enabled:
        self.monitor = self.monitor if self.monitor is not None else True
        self.dump_gradients = self.dump_gradients if self.dump_gradients is not None else False
```

farhatkevin: done, thanks!

src/olmo_core/train/callbacks/gap_monitor.py
```diff
@@ -171,6 +246,79 @@ def record_tensor_stats(
...
+            return
+
+        if (self.step - self.dump_gradients_start_step) % self.dump_gradients_step_interval != 0:
+            return
+
+        output_dir = self.trainer.work_dir / "gradients"
+        output_dir.mkdir(exist_ok=True, parents=True)
+
+        step_dir = output_dir / f"step{self.step}"
+        step_dir.mkdir(exist_ok=True, parents=True)
```
epwalsh: Minor simplification:
```suggestion
        step_dir = self.trainer.work_dir / "gradients" / f"step{self.step}"
        step_dir.mkdir(exist_ok=True, parents=True)
```

src/scripts/inspect_tensor.py
```diff
@@ -0,0 +1,117 @@
...
+except ImportError:
+    print("Error: safetensors library not installed.")
+    print("Install it with: pip install safetensors")
+    sys.exit(1)
+
+
+def inspect_safetensors(file_path, show_values=False):
+    """Load and inspect a safetensors file."""
+
+    file_path = Path(file_path)
```
epwalsh: Better yet could wrap this with `cached-path` (`from cached_path import cached_path`) to handle URLs.
```suggestion
    file_path = cached_path(file_path)
```

epwalsh (approved):
LGTM! Just two minor suggestions

src/olmo_core/train/callbacks/gap_monitor.py
```diff
@@ -171,6 +246,76 @@ def record_tensor_stats(
...
+                    if get_rank() == 0:
+                        full_grad = full_grad.cpu()
+
+                        dim_size = full_grad.shape[0]
+                        actual_n = min(self.dump_gradients_save_first_n, dim_size)
+                        if actual_n < self.dump_gradients_save_first_n:
+                            log.warning(
+                                f"Parameter '{name}': dump_gradients_save_first_n={self.dump_gradients_save_first_n} exceeds "
+                                f"dimension size {dim_size}, capping to {actual_n}"
+                            )
```
dirkgr: I don't think we need this warning.

src/olmo_core/train/callbacks/gap_monitor.py
```diff
@@ -171,6 +246,76 @@ def record_tensor_stats(
...
+
+                        sliced_grad = full_grad.narrow(0, 0, actual_n)
```
dirkgr: It should narrow across all dimensions, not just the first. If we have a tensor `(d_model, vocab_size)`, that's still huge if it's `(10, vocab_size)`.

src/olmo_core/train/callbacks/gap_monitor.py
```diff
@@ -171,6 +246,76 @@ def record_tensor_stats(
...
+                        sliced_filename = f"{name}_first{actual_n}.safetensors"
```
dirkgr: It also contains `firstN` if it's not sampled?

src/olmo_core/train/trainer.py
dirkgr: What about this comment?

src/scripts/inspect_tensor.py
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env python3
+"""
+Simple script to load and inspect safetensors files.
+Shows tensor names, shapes, dtypes, and other metadata.
+Can use some work for more detailed inspection!
```
dirkgr: No sandbagging. The script is what it is. All code could be better.

src/scripts/inspect_tensor.py
```diff
@@ -0,0 +1,118 @@
...
+import sys
+
+from cached_path import cached_path
+
+try:
+    from safetensors import safe_open
+except ImportError:
+    print("Error: safetensors library not installed.")
+    print("Install it with: pip install safetensors")
+    sys.exit(1)
```
dirkgr: Why is this guarded like this? Just let it fail like any other missing import fails?

src/scripts/inspect_tensor.py
```diff
@@ -0,0 +1,118 @@
...
+                    print(f"Dtype:  {dtype}")
+                    print(f"Params: {num_params:,}")
+
+                print("-" * 80)
+
+            print(f"\n✅ Total parameters: {total_params:,}")
+            print(f"   ({total_params / 1e6:.2f}M parameters)")
+
+    except Exception as e:
+        print(f"Error loading file: {e}")
```
dirkgr: Don't catch all exceptions. Just let it fail.

This one is particularly bad, because it turns errors into successes. When you run this script, and it can't read the file, it will still return with a 0 error code.

dirkgr (dismissed):

AkshitaB (approved):

PR #439, opened 2025-10-30
OlmoCore native chat interface

tyler-romero:
Its useful to have an OlmoCore native chat interface for vibe testing and for debugging the VLLM implementation.

---

> [!NOTE]
> Adds an interactive `olmo_core.generate.chat` CLI and enables choosing the attention backend when loading generation modules; tweaks default device selection.
> 
> - **Added**
>   - `olmo_core.generate.chat`: interactive CLI for chatting with OLMo-core checkpoints.
>     - Tokenizer auto-load from checkpoint, Jinja chat templates, system prompt, sampling params, KV cache toggle, dtype/attention backend options, and Rich UI.
> - **Generation** (`TransformerGenerationModule`)
>   - `from_checkpoint(..., attention_backend=...)` to override attention backend; clears deprecated `use_flash`.
>   - Import `AttentionBackendName`; removes strict CUDA/H100 assumptions.
> - **Utils**
>   - `get_default_device()`: prefer CUDA if `torch.cuda.is_available()` (drops `is_initialized()` check).
> - **Changelog**
>   - Notes new chat module.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 809c867bf38cc6a5d50f83557a88d7c08de1830c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/olmo_core/distributed/utils.py
+++ b/src/olmo_core/distributed/utils.py
@@ -66,7 +66,7 @@ def init_distributed(
             # NOTE: For single-node training we still need all of these settings and we also
             # need host networking enabled so that the ethernet interface names don't change.
             set_env_var("NCCL_CROSS_NIC", "0")
-            set_env_var("NCCL_ALGO", "Ring,Tree")
+            #  set_env_var("NCCL_ALGO", "Ring,Tree")
             set_env_var("NCCL_PROTO", "Simple,LL128")
             set_env_var("NCCL_MIN_NCHANNELS", "4")
             set_env_var("NCCL_P2P_NET_CHUNKSIZE", "524288")
@@ -83,7 +83,7 @@ def init_distributed(
             set_env_var(
                 "NCCL_FASTRAK_PLUGIN_ACCEPT_TIMEOUT_MS", str(int(timeout.total_seconds() * 1000))
             )
-            set_env_var("NCCL_NVLS_ENABLE", "0")
+            #  set_env_var("NCCL_NVLS_ENABLE", "0")
             set_env_var("NCCL_USE_SNAP", "1")
             set_env_var("NCCL_FASTRAK_USE_LLCM", "1")
             set_env_var("NCCL_FASTRAK_LLCM_DEVICE_DIRECTORY", "/dev/aperture_devices")


src/olmo_core/generate/chat.py
```diff
@@ -0,0 +1,489 @@
...
+        "--do-sample",
+        action="store_true",
+        help="Use sampling (default: True). Set --no-do-sample for greedy decoding.",
+    )
+    parser.add_argument(
+        "--no-do-sample",
+        dest="do_sample",
+        action="store_false",
+        help="Disable sampling (use greedy decoding)",
+    )
```
epwalsh, line 250: Maybe this instead?

```suggestion
    parser.add_argument(
        "--do-sample",
        action=argparse.BooleanOptionalAction,
        default=True,  # could also be `None`
        help="Use sampling (default: True). Set --no-do-sample for greedy decoding.",
    )
```

epwalsh (commented):
Is there a small, reasonably good model I could test this with on a MacBook?

tyler-romero:
> Is there a small, reasonably good model I could test this with on a MacBook?

I'm not informed of the quality of any models smaller than 7B. You could run something like:
```
uv run python -m olmo_core.generate.chat /weka/oe-training-default/ai2-llm/checkpoints/allysone/anneal-round5-100B-olmo25_7b-anneal-6T-decon-sparkle-motion-8730626c/step47684 --max-new-tokens 1024 --temperature 1.0 --top-p 0.7 --show-special-tokens --no-use-cache
```

On mac you need to disable kv cache b/c using the kv cache requires flash attn

src/olmo_core/generate/chat.py
```diff
@@ -0,0 +1,445 @@
...
+        default="WARNING",
+        choices=["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"],
+        help="Logging verbosity level (default: INFO)",
+    )
+
+    args = parser.parse_args()
+    logging.basicConfig(
+        level=getattr(logging, args.verbosity),
+        format="%(asctime)s - %(name)s - %(levelname)s - %(message)s",
+    )
```
epwalsh: Any reason not to use `olmo_core.utils.prepare_cli_environment()` instead?

epwalsh (approved):
LGTM, just one more minor suggestion.

src/olmo_core/generate/chat.py
```diff
@@ -0,0 +1,445 @@
...
+
+    device = torch.device("cuda") if torch.cuda.is_available() else torch.device("cpu")
```
dirkgr: Can we have support for mps? It mostly works out of the box.

dirkgr:
Dangerous question: How hard would it be to expose an endpoint?

tyler-romero:
> How hard would it be to expose an endpoint?

Not too hard, but if we were to do that I would want to make it openai compatible, which is quite different from this PR. Happy to do that at some point in the future.

src/olmo_core/generate/chat.py
tyler-romero: Mps support added and instructions for running on mac added to the top (need to work around the flash-attn dependency).

Also prob want to point to an unsharded checkpoint to avoid eating too much of your macs disk space.

PR #452, opened 2025-11-07
Instance mask correction to avoid nan loss

tyler-romero:
When training with a fixed batch size on a bunch of devices its possible to hit a nan loss if all of the instances that land on one device are masked. We're more likely to hit this when training on many devices vs training on fewer devices because fewer instances land on one device.

---

> [!NOTE]
> Include masked-instance tokens in loss divisor to avoid NaNs/low loss when all instances on a rank are masked; tweak related metrics and update changelog.
> 
> - **Train module (`src/olmo_core/train/train_module/transformer/train_module.py`)**:
>   - Adjust loss token count by adding tokens from masked instances to `batch_num_tokens_for_loss`, ensuring fair per-rank loss reduction and preventing NaNs when instances are fully masked.
>   - Metrics: record masked labels/instances as proportions (not percentages) and always log instance masking when present.
> - **Changelog**:
>   - Add "Fix potential NaN loss that can occur when using instance masking."
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 29dabb3ba9da8dbcc83cc8babef8dca5041b08e0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/olmo_core/train/train_module/transformer/train_module.py
+++ b/src/olmo_core/train/train_module/transformer/train_module.py
@@ -346,6 +346,7 @@ class TransformerTrainModule(TrainModule):
 
         # Calculate how many tokens will be used in the loss.
         batch_num_tokens = batch["labels"].numel()
+        batch_num_tokens_per_instance = batch["labels"].shape[1]
         batch_num_tokens_for_loss = move_to_device(
             (batch["labels"] != self.label_ignore_index).sum(), self.device
         )
@@ -364,10 +365,13 @@ class TransformerTrainModule(TrainModule):
                 (~instance_mask).float().mean(),
                 ReduceType.mean,
             )
-            # WARN: When we mask out all local instances with the instance filter, we add a small epsilon
-            # to the loss divisor to avoid division by zero while minimizing the impact on loss calculation.
-            all_local_instances_masked = (~instance_mask).all()
-            batch_num_tokens_for_loss += all_local_instances_masked * 1e-5
+
+            # WARN: When we mask out instances with the instance filter, we count those tokens
+            # for the loss anyways. They will count as tokens with a zero loss. This means we
+            # get an artificially *low* loss for these batches. But it is really hard (and slow)
+            # to do this properly in a distributed setup. We add back in the full number of tokens
+            # for the loss so that each rank contributes to the loss calculation fairly.
+            batch_num_tokens_for_loss += (~instance_mask).sum() * batch_num_tokens_per_instance
 
         # Batch losses to record.
         ce_batch_loss = move_to_device(torch.tensor(0.0), self.device)


src/olmo_core/train/train_module/transformer/train_module.py
```diff
@@ -344,22 +344,31 @@ def train_batch(self, batch: Dict[str, Any], dry_run: bool = False):
...
+            # Record percentage of masked instances.
+            if (instance_mask := batch.get("instance_mask")) is not None:
+                self.record_metric(
+                    "train/masked instances (%)", (~instance_mask).float().mean(), ReduceType.mean
+                )
+                # WARN: When we mask out instances with the instance filter, we count those tokens
+                # for the loss anyways. They will count as tokens with a zero loss. This means we
+                # get an artificially *low* loss for these batches. But it is really hard (and slow)
+                # to do this properly in a distributed setup.
+                batch_num_tokens_for_loss += (~instance_mask).sum() * batch_num_tokens_per_instance
```
tyler-romero: @dirkgr thoughts on just adding 1 token per masked instance instead of `batch_num_tokens_per_instance` to avoid making the loss more artificially low than it needs to be?

epwalsh: Or a small epsilon. Doesn't have to be an integer.

tyler-romero: Done

src/olmo_core/train/train_module/transformer/train_module.py
```diff
@@ -344,22 +344,31 @@ def train_batch(self, batch: Dict[str, Any], dry_run: bool = False):
...
+            if (instance_mask := batch.get("instance_mask")) is not None:
+                self.record_metric(
+                    "train/masked instances (%)",  # just a proportion, not a percentage
+                    (~instance_mask).float().mean(),
+                    ReduceType.mean,
+                )
+                # WARN: When we mask out all local instances with the instance filter, we add a small epsilon
+                # to the loss divisor to avoid division by zero while minimizing the impact on loss calculation.
+                all_local_instances_masked = (~instance_mask).all()
+                batch_num_tokens_for_loss += all_local_instances_masked * 1e-5
```
epwalsh: Probably should have this in the dry-run as well?

src/olmo_core/train/train_module/transformer/train_module.py
tyler-romero: Discussed with dirk offline: https://allenai.slack.com/archives/C05K6TF5W3Z/p1762559683603419?thread_ts=1762556299.098969&cid=C05K6TF5W3Z

PR #470, opened 2025-11-20
2B config

dirkgr:


epwalsh (commented):
Everything looks good now other than learning rate.

src/scripts/train/OLMo3/OLMo3-2B-ty.py
```diff
@@ -0,0 +1,187 @@
...
+        .with_callback(
+            "checkpointer",
+            CheckpointerCallback(
+                save_interval=1000,
+                ephemeral_save_interval=None,
+                save_async=False,
+            ),
+        )
+        .with_callback(
+            "comet",
```
baileykuehl, line 137: Do we still need this even though we don't have comet anymore? Won't it always be enabled=False?

tyler-romero: Yeah I agree we can drop it in configs going forwards

PR #479, opened 2025-11-26
Fix MFU Calculation for SWA and GQA (Issue #475)

### LINKED ISSUE #475: MFU calculation is a little wrong with SWA and GQA

ada-ggf25:
# Fix MFU Calculation for SWA and GQA (Issue #475)

Fixes #475

## Problem

`SpeedMonitorCallback` derives MFU from `num_flops_per_token()`, but the helper assumed dense attention with `n_heads == n_kv_heads` and that every layer attended over the full sequence. When Sliding Window Attention (SWA), Grouped Query Attention (GQA), or block overrides were enabled, the FLOPS estimate (and therefore MFU) was inflated because:

- SWA windows were ignored, so FLOPS used the full sequence length instead of the per-layer window.
- GQA still used `n_heads` instead of `n_kv_heads`.
- Block overrides that changed `n_heads` never recalculated the head dimension `q`.

## Solution

1. **Accurate FLOPS calculation**
   - `src/olmo_core/nn/transformer/config.py`: `TransformerConfig.num_flops_per_token()` now iterates per layer, recomputes `q = d_model // 
n_heads`, uses each layer’s `n_kv_heads`, and caps SWA window sizes at the actual sequence length.
   - `src/olmo_core/nn/transformer/model.py`: `Transformer.num_flops_per_token()` mirrors the same logic and stores the block configs so 
overrides with custom `n_heads` are honoured.
2. **Tests**
   - Added/expanded coverage in `src/test/nn/transformer/model_test.py` for:
     - GQA-only, SWA-only, and combined SWA+GQA scenarios.
     - SWA patterns with alternating window sizes.
     - Block overrides that change `n_heads`.
     - SWA window sizes larger than the actual sequence length.
3. **Docs**
   - `CHANGELOG.md` documents that MFU accounts for SWA (with window capping) and per-layer head counts.

## Files Changed

- `src/olmo_core/nn/transformer/config.py`
- `src/olmo_core/nn/transformer/model.py`
- `src/test/nn/transformer/model_test.py`
- `CHANGELOG.md`

## Testing

```bash
make style
make docs
make build
make checks
```

All FLOPS-related tests pass and the repo passes the style/docs/build/checks targets, confirming MFU now reflects the true FLOPS for SWA and GQA configurations and the tree is clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fixed MFU calculation to account for sliding window attention (SWA) and grouped-query attention (GQA). The `num_flops_per_token()` method in `TransformerConfig` and `Transformer` now correctly computes FLOPS per layer, using window sizes for SWA and `n_kv_heads` for GQA instead of assuming full attention and equal query/key-value heads.
+- Fixed MFU calculation to account for sliding window attention (SWA) and grouped-query attention (GQA). The `num_flops_per_token()` method in `TransformerConfig` and `Transformer` now correctly computes FLOPS per layer, using window sizes for SWA (capped at the actual sequence length) and `n_kv_heads` for GQA instead of assuming full attention and equal query/key-value heads. Head dimensions are recalculated per layer so block overrides with custom `n_heads` values are handled correctly.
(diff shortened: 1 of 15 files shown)


ada-ggf25:
Need to Fix what the Cursor bugbot point out.

PR #480, opened 2025-11-26
Fix MFU Calculation for SWA and GQA (Issue #475)

### LINKED ISSUE #475: MFU calculation is a little wrong with SWA and GQA

ada-ggf25:
# Fix MFU Calculation for SWA and GQA (Issue #475)

Fixes #475

## Problem

`SpeedMonitorCallback` derives MFU from `num_flops_per_token()`, but the helper assumed dense attention with `n_heads == n_kv_heads` and that every layer attended over the full sequence. When Sliding Window Attention (SWA), Grouped Query Attention (GQA), or block overrides were enabled, the FLOPS estimate (and therefore MFU) was inflated because:

- SWA windows were ignored, so FLOPS used the full sequence length instead of the per-layer window.
- Block overrides that changed `n_heads` never recalculated the head dimension `q`.

## Solution

1. **Accurate FLOPS calculation**
   - `src/olmo_core/nn/transformer/config.py`: `TransformerConfig.num_flops_per_token()` now iterates per layer, recomputes `q = d_model // 
n_heads`, caps SWA window sizes at the actual sequence length, and respects per-layer GQA configuration while treating FLOPS as dense 
attention (proportional to `n_heads`).
   - `src/olmo_core/nn/transformer/model.py`: `Transformer.num_flops_per_token()` mirrors the same logic and stores the block configs so 
overrides with custom `n_heads` are honoured.
2. **Tests**
   - Added/expanded coverage in `src/test/nn/transformer/model_test.py` for:
     - GQA-only, SWA-only, and combined SWA+GQA scenarios.
     - SWA patterns with alternating window sizes.
     - Block overrides that change `n_heads`.
     - SWA window sizes larger than the actual sequence length.
3. **Docs**
   - `CHANGELOG.md` documents that MFU accounts for SWA (with window capping) and per-layer head counts.

## Files Changed

- `src/olmo_core/nn/transformer/config.py`
- `src/olmo_core/nn/transformer/model.py`
- `src/test/nn/transformer/model_test.py`
- `CHANGELOG.md`

## Testing

```bash
make style
make docs
make build
make checks
```

All FLOPS-related tests pass and the repo passes the style/docs/build/checks targets, confirming MFU now reflects the true FLOPS for SWA and GQA configurations and the tree is clean.

---

> [!NOTE]
> Corrects FLOPS/MFU estimation by accounting for sliding-window attention, per-layer overrides, and GQA, and adds targeted tests.
> 
> - **Core (FLOPS/MFU computation)**:
>   - Update `TransformerConfig.num_flops_per_token()` and `Transformer.num_flops_per_token()` to sum per-layer attention FLOPS using effective SWA window sizes (capped by `seq_len`) and respect block overrides; treat GQA FLOPS as dense (proportional to `n_heads`).
>   - Add `num_flops_per_token()` to `AttentionConfig` and `Attention` and declare it in `AttentionBase` to keep config/module estimates aligned.
>   - Cache per-layer block configs in `Transformer` for accurate FLOPS.
> - **Tests**:
>   - Add tests in `src/test/nn/transformer/model_test.py` covering GQA-only, SWA-only, SWA+GQA, SWA patterns, block overrides with different `n_heads`, and window>sequence capping.
> - **Docs**:
>   - Update `CHANGELOG.md` to note corrected MFU accounting for SWA/GQA.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 83ba4dda71b165416d583b4199ae951080cdd6f5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/olmo_core/nn/attention/__init__.py
+++ b/src/olmo_core/nn/attention/__init__.py
@@ -297,6 +297,16 @@ class AttentionBase(nn.Module):
     """
 
     @abstractmethod
+    def num_flops_per_token(self, seq_len: int) -> int:
+        """
+        Approximate the attention FLOPs per token for this module.
+
+        Implementations should mirror the heuristic used by their corresponding
+        :class:`AttentionConfig` so that config-level and module-level FLOPs
+        estimates stay in sync.
+        """
+
+    @abstractmethod
     def apply_tp(
         self,
         tp_mesh: DeviceMesh,
@@ -669,6 +679,19 @@ class Attention(AttentionBase):
             head_dim=self.head_dim,
             device=self.w_k.weight.device,
         )
+
+    def num_flops_per_token(self, seq_len: int) -> int:
+        """
+        Approximate the attention FLOPs per token for this module.
+
+        This mirrors :meth:`AttentionConfig.num_flops_per_token`, using the
+        PaLM-style heuristic based on the number of heads, head dimension, and
+        effective sequence length.
+        """
+        # The factor 12 matches the existing heuristic:
+        # 2 matmuls (QK^T, attn @ V) * 2 (forward + backward) * 2 (mult + add)
+        # * ~1.5 overhead.
+        return 12 * self.n_heads * self.head_dim * seq_len
 
 
 @beta_feature


src/olmo_core/nn/transformer/config.py
```diff
@@ -452,20 +452,49 @@ def num_active_non_embedding_params(self) -> int:
     def num_flops_per_token(self, seq_len: int) -> int:
         """
         Get the approximate number of flops per token.
+
+        Accounts for sliding window attention (SWA) and grouped-query attention (GQA).
+        For SWA, uses the window size for each layer instead of the full sequence length.
+        For GQA, uses per-layer configuration but treats FLOPS as dense attention (same as n_heads).
         """
-        n, h, q, t = (
```
tyler-romero, line 456: Thanks for your suggestion! We agree that this MFU heuristic breaks down under a variety of scenarios:
1) SWA / Sparse Attn / Linear Attn Variants
2) GQA
3) Alternate transformer blocks like those with gating

Ultimately to upgrade from this basic heuristic (proposed in [PALM](https://arxiv.org/pdf/2204.02311#page=65.09)) we want a solution that draws from the transformer's underlying building blocks - eg ones where attention mechanisms self-report the number of flops they require to blocks, which self-report the number of flops they require to the overall Transformer.

tyler-romero: Was thinking about this more and I think that another issue in our existing MFU computation that we would want to address is that the attention flops approximation in PaLM breaks down for long seq lengths (because it should have some t^2 factor) - this is also something we want to fix - we're doing some work currently that really cares about the accuracy of these flop calculations. For SWA, the correct number of flops should be something like (t * t_sw).

ada-ggf25: Thank you for the review and feedback! I appreciate you taking the time to look at this.

You're absolutely right about the attention FLOPS approximation. The current implementation uses the PaLM heuristic which is linear in sequence length (`12 * n_heads * q * t`), but for SWA the correct formula should account for the quadratic nature of attention: each of the `t` tokens attends to `t_sw` tokens in the window, so it should be `t * t_sw` rather than just using the window size linearly.

I'll update this PR to fix the immediate mathematical issue (using `t * t_sw` for SWA) as an incremental improvement.

I also understand you're working towards a more comprehensive architectural solution where attention mechanisms and blocks self-report their FLOPS, which would be a much cleaner and more extensible approach for handling SWA, sparse attention variants, linear attention, GQA, and alternate transformer blocks.

I'd love to contribute to that architectural refactor in any way you see fit if you'd like an extra pair of hands. It sounds like a great direction for making the FLOPS calculations more accurate and maintainable.

ada-ggf25:
I've created a follow-up PR that implements the `t * t_sw` formula you suggested for SWA FLOPS calculation. The new PR addresses your feedback by accounting for the quadratic nature of attention where each of `t` tokens attends to `t_sw` tokens in the window.

**Changes:**
- Updated `num_flops_per_token()` to use `seq_len * window_size` for SWA (when `window_size < seq_len`)
- Uses `seq_len^2` when `window_size >= seq_len` (equivalent to dense attention if it were quadratic)
- Updated all related tests to reflect the new formula
- Updated CHANGELOG documentation

This is an incremental improvement addressing the immediate mathematical issue.

The new PR #482. I hope it helps!

ada-ggf25:
Considering the detailed feedback here and on the follow‑up PR [#482](https://github.com/allenai/OLMo-core/pull/482) – especially the clarification that the original per‑token FLOPs behaviour was correct and that the real missing piece was having the attention modules self‑report their FLOPs.

As a follow‑up to that discussion, I have updated this PR’s branch to:

- **Restore the original per‑token FLOPs maths**  
  - I removed the `t * t_sw` change from [#482](https://github.com/allenai/OLMo-core/pull/482), so both `TransformerConfig.num_flops_per_token` and `Transformer.num_flops_per_token` now use the same per‑token PaLM‑style heuristic as before, just applied per layer with SWA window capping and GQA handled as dense (proportional to `n_heads`).

- **Add a module‑level FLOPs API for attention**  
  - Introduced `AttentionBase.num_flops_per_token(self, seq_len: int)` and implemented it for the standard `Attention` module using the same factor‑12 PaLM heuristic that `AttentionConfig.num_flops_per_token` uses.  
  - This keeps config‑level and module‑level estimates aligned while giving backends a clear hook to override FLOPs estimation if needed.

- **Prefer the attention module as the source of truth in `Transformer`**  
  - Updated `Transformer.num_flops_per_token` to, for each layer, compute the same `effective_seq_len` as before (respecting SWA) and then:
    - Call `block.attention.num_flops_per_token(effective_seq_len)` when that method exists, or  
    - Fall back to `block_config.attention.num_flops_per_token(d_model=self.d_model, seq_len=effective_seq_len)` otherwise.  
  - `TransformerConfig.num_flops_per_token` continues to use the config‑level helper, so the configuration API remains unchanged.

Please let me know if you would like the FLOPs helper or its docstrings to be structured differently, or if you would prefer any of this to live only on the config side rather than also on the module. I am happy to adjust the shape of the API to better match your longer‑term plans for MFU/FLOPs accounting.

tyler-romero:
Thanks! Before I give another review can you fix the formatting? Running `make style` in the root of the repo should do it. That way theres fewer extraneous lines of code to dig through.

ada-ggf25:
> Thanks! Before I give another review can you fix the formatting? Running `make style` in the root of the repo should do it. That way theres fewer extraneous lines of code to dig through.

Already did:

1. ./OLMo-core
❯ make style
======== formatting with isort... ========
Skipped 5 files
======== formatting with black... ========
All done! ✨ 🍰 ✨
314 files left unchanged.

2. ./OLMo-core
❯ make checks
======== running isort... ========
Skipped 5 files
======== running black... ========
All done! ✨ 🍰 ✨
314 files would be left unchanged.
======== running ruff... =========
All checks passed!
======== running mypy... =========
Success: no issues found in 313 source files

Are there still anything out of format, maybe in early commits?  The PR includes commits from earlier work. If you're seeing formatting issues in the GitHub diff, it might be from those earlier commits. I'm sorry for the trouble.

tyler-romero:
Hmmm, thanks for clarifying - I wonder if you have some rules that are overriding our workspace rules. Its ok I can format on my side tomorrow and then take a look.

ada-ggf25:
> Hmmm, thanks for clarifying - I wonder if you have some rules that are overriding our workspace rules. Its ok I can format on my side tomorrow and then take a look.

Weird, I tried deleting the virtual environment I was using, recreating it from scratch, and running `make style`, and I got the same output. I’m not sure what’s going on, but thanks for the heads up. I’m going to take a closer look at this.

tyler-romero:
Ended up making the flop counting corrections in this PR: https://github.com/allenai/OLMo-core/pull/511

Thanks again

ada-ggf25:
> Ended up making the flop counting corrections in this PR: #511
> 
> Thanks again

I'm glad to contribute! Thank you for the guidance. I'm going to close this PR now then.

PR #487, opened 2025-11-30
Fix type annotation issue in NumpyInterleavedFSLDataset (Issue #486)

### LINKED ISSUE #486: Fix type annotation issue in NumpyInterleavedFSLDataset
# Fix type annotation issue in NumpyInterleavedFSLDataset

## Description

The `NumpyInterleavedFSLDataset` class has missing type annotations for two instance attributes, causing mypy type checking errors.

## Problem

When running `make type-check`, mypy reports the following error:

```
src/olmo_core/data/numpy_dataset.py:1416: error: Incompatible types in assignment (expression has type "Any | int", variable has type "None")  [assignment]
```

## Root Cause

In `NumpyInterleavedFSLDataset.__init__()`, two attributes are initialised to `None` without explicit type annotations:

```python
self._num_interleaving_exempt_instances = None
self._num_interleavable_instances = None
```

mypy infers these as literal `None` types rather than `Optional[int]`. Later in the `__len__()` method (line 1416), these attributes are 
assigned integer values, causing a type mismatch error.

**Note**: This appears to be an oversight rather than a deliberate design choice, as:
- The parent class `NumpyPaddedFSLDataset` uses `_num_instances: Optional[int] = None` (line 729)
- All other similar lazy-initialized attributes throughout the codebase use `Optional[int]` type annotations
- The original commit (5cc16a2) that added this class also omitted annotations while the parent class already had them
- There are no comments or documentation explaining why these attributes should differ from the established pattern

## Expected Behaviour

The attributes should be properly typed as `Optional[int]` to match the pattern used throughout the codebase and allow mypy to correctly type-check assignments.

## Steps to Reproduce

1. Run type checking:
   ```bash
   make type-check
   ```
2. Observe the error of `src/olmo_core/data/numpy_dataset.py`

## Proposed Solution

Add explicit type annotations to match the pattern used elsewhere:

```python
self._num_interleaving_exempt_instances: Optional[int] = None
self._num_interleavable_instances: Optional[int] = None
```

This is consistent with how `_num_instances` is typed in the parent class and other similar attributes throughout the codebase.

## Environment

- **Python Version**: 3.10+
- **mypy Version**: >=1.0,<1.4 (as specified in `pyproject.toml`)
- **File**: `src/olmo_core/data/numpy_dataset.py`
- **Class**: `NumpyInterleavedFSLDataset`

ada-ggf25:
# Fix type annotation issue in NumpyInterleavedFSLDataset (Issue #486)

## Summary

This PR fixes a mypy type checking error in `NumpyInterleavedFSLDataset` by adding explicit `Optional[int]` type annotations to two ins attributes that were missing type hints.

Fixes #486

## Changes

- Added `Optional[int]` type annotations to `_num_interleaving_exempt_instances` and `_num_interleavable_instances` in `NumpyInterleavedFSLDataset.__init__()`
- Updated `CHANGELOG.md` to document the fix

## Problem

When running `make type-check`, mypy reported:

```
src/olmo_core/data/numpy_dataset.py:1416: error: Incompatible types in assignment (expression has type "Any | int", variable has type "None")  [assignment]
```

The attributes were initialised to `None` without type annotations, causing mypy to infer them as literal `None` types. Later assignmen integer values in `__len__()` resulted in a type mismatch.

## Solution

Added explicit type annotations to match the pattern used throughout the codebase:

```python
self._num_interleaving_exempt_instances: Optional[int] = None
self._num_interleavable_instances: Optional[int] = None
```

This is consistent with how similar attributes are typed in the parent class (`NumpyPaddedFSLDataset`) and other dataset classes.

## Testing

- Verified the fix resolves the mypy error at line 1416
- Confirmed the change is consistent with patterns used elsewhere in the codebase
- No linter errors introduced
- Type checking passes for the modified file (remaining errors are unrelated missing type stubs for external libraries)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fixed `AttentionConfig.num_params()` overcounting QK norm parameters when using GQA/MQA with `use_head_qk_norm=False`.
 - Fixed the peak learning rate in `src/scripts/train/OLMo3/OLMo3-32B-midtraining.py` to the correct one.
+- Fixed type annotation issue in `NumpyInterleavedFSLDataset` where `_num_interleaving_exempt_instances` and `_num_interleavable_instances` were missing `Optional[int]` type hints, causing mypy type errors.
 
 
 ### Changed


epwalsh (approved):
Thanks @ada-ggf25, LGTM!

PR #504, opened 2025-12-16
Avoid torch recompile when intra-doc masking enabled

### LINKED ISSUE #500: [Performance] Frequent recompilation triggered by dynamic shapes (cu_doc_lens) and varying scalars (max_seq_len) in block_kwargs
I encountered a performance issue related to `torch.compile` where the model triggers frequent recompilations during training.

This happens inside the transformer block loop where `block_kwargs` are passed to the block. 
Specifically, `block_kwargs` contains arguments that vary in shape or value, preventing the compiler from caching the graph effectively.

In the following code segment:

https://github.com/allenai/OLMo-core/blob/89cef3df1e8c0beff479919dd657aaa0a46e87e2/src/olmo_core/nn/transformer/model.py#L510

There are two specific arguments inside block_kwargs causing the cache miss and recompilation:

`cu_doc_lens` (Dynamic Shape Tensor):
The tensor `cu_doc_lens` has a dynamic shape depending on the batch structure or sequence lengths. 
Since it is passed via `block_kwargs` and not explicitly marked as dynamic using `torch._dynamo.mark_dynamic` , the compiler treats each new shape as a new graph signature.

`max_seq_len` (Varying Scalar/Int):
The argument `max_seq_len` is passed as a standard Python int. Its value changes depending on the input. By default, `torch.compile` specializes on scalar values. If `max_seq_len` changes (e.g., from 128 to 129), it triggers a full recompilation unless it is marked as a symbolic variable or wrapped in a Tensor.

We need to ensure that arguments inside block_kwargs are properly handled before passing them to the compiler.

I'm happy to open a Pull Request to fix this. Please let me know if this solution aligns with the project.

epwalsh:
Fixes #500. 

- Run before the fix with extra logging enabled: https://beaker.org/ex/01KCM28516ZJKW5P02G0GHAXNW
- Run with the fix with the same logging enabled: https://beaker.org/ex/01KCM4N3G1VK9F64FXMSCJGPMC

See https://docs.pytorch.org/docs/main/user_guide/torch_compiler/compile/dynamic_shapes_core_concepts.html.

---

> [!NOTE]
> Marks `cu_doc_lens` and `max_doc_len` as dynamic to avoid torch recompiles with intra-document masking and enables `TORCH_LOGS` in Beaker runs.
> 
> - **Transformer / compile**:
>   - Mark `cu_doc_lens` dynamic via `mark_dynamic(..., dim=0)` to stabilize compiled graphs.
>   - Treat `max_doc_len` as a dynamic source in `torch.compiler.config`.
> - **Beaker launcher**:
>   - Add default env var `TORCH_LOGS=recompiles,graph_breaks`.
> - **Changelog**:
>   - Note fix to avoid torch dynamo recompiles when intra-document masking is enabled.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 31cf5224b6320b510b9bbd5283a24924201f7994. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed the peak learning rate in `src/scripts/train/OLMo3/OLMo3-32B-midtraining.py` to the correct one.
 - Fixed type annotation issue in `NumpyInterleavedFSLDataset` where `_num_interleaving_exempt_instances` and `_num_interleavable_instances` were missing `Optional[int]` type hints, causing mypy type errors.
 - Fixed bug in GPUMonitorCallback where it was using a Wandb reserved keyword, causing data to be unable to be visualized in the Wandb dashboard.
+- Avoid torch dynamo recompiles when intra-document masking enabled by marking `cu_doc_lens` and `max_doc_len` dynamic.
 
 ### Changed
 


src/olmo_core/launch/beaker.py
```diff
@@ -343,6 +343,7 @@ def default_env_vars(self) -> List[Tuple[str, str]]:
             ("NUM_NODES", str(self.num_nodes)),
             ("OLMO_CORE_VERSION", VERSION),
             ("FORCE_COLOR", "1"),  # for 'rich' because Beaker supports ANSI colors in logs
+            ("TORCH_LOGS", "recompiles,graph_breaks"),
```
epwalsh, line 346: There doesn't seem to be any performance penalty to having these logs enabled and it will help us catch future issues early.

PR #509, opened 2025-12-20
No-global-rope (GNoPE)

tyler-romero:
RoPE config option to disable on global attn layers.


tyler-romero:
I dont love the config factorization here... We have separate SWA and RoPE configs and this option depends on both.

I could move it up into the Attention block but that also seems weird since the rest of our RoPE config is in one place.

PR #510, opened 2025-12-20
MXFP8 Linear Support

tyler-romero:
Support for MXFP8 training on blackwell via torchao.

Successful small-scale run on B200s:
https://beaker.allen.ai/orgs/ai2/workspaces/oe-t-ladder/work/01KCYM8R89KDNNVZ4BKKTR1JYG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added support for gated attention.
 - Added support for no-global-rope ("GNoPE").
 - Added support for MXFP8 Linear layers via torchao.
+- Added support for tracking total flops.
 
 ### Fixed
 
@@ -28,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed bug in GPUMonitorCallback where it was using a Wandb reserved keyword, causing data to be unable to be visualized in the Wandb dashboard.
 - Fixed the ConsoleLoggerCallback filtering to support the new prefix (gpu_memory) for GPUMonitorCallback.
 - Avoid torch dynamo recompiles when intra-document masking enabled by marking `cu_doc_lens` and `max_doc_len` dynamic.
+- Flops tracking for ParallelMLP and SWA layers.
 
 ### Changed
 
--- a/src/olmo_core/nn/attention/__init__.py
+++ b/src/olmo_core/nn/attention/__init__.py
@@ -328,6 +328,10 @@ class AttentionBase(nn.Module):
     ):
         raise NotImplementedError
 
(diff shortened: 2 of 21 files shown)


src/olmo_core/float8/__init__.py
```diff
@@ -25,17 +27,34 @@ class Float8Config(Config):
 
     :param ao: A torchao ``Float8Linear`` linear configuration.
     :param ao_recipe: Alternatively you can specify a recipe name from torchao.
+    :param ao_mx: A torchao ``MXLinearConfig`` configuration for MX formats (MXFP8/MXFP4).
     :param enabled: If ``False`` this will be a no-op.
     """
 
     ao: Optional[AOFloat8LinearConfig] = None
     ao_recipe: Optional[AOFloat8LinearRecipe] = None
+    ao_mx: Optional[AOMXLinearConfig] = None
```
tyler-romero, line 36: could split this out into its own `MXConfig` class, might be cleaner

src/olmo_core/float8/utils.py
tyler-romero, line 1: unused utilities

pyproject.toml
```diff
@@ -59,7 +59,7 @@ dev = [
 beaker = ["beaker-py>=1.37.0,<2.0", "GitPython>=3.0,<4.0", "google-cloud-compute"]
 comet = ["comet_ml"]
 eval = ["ai2-olmo-eval==0.8.5"]
-torchao = ["torchao"]
+torchao = ["torchao==0.15.0"]
```
epwalsh, line 62: Should probably pin this in our Docker file as well: https://github.com/allenai/OLMo-core/blob/6bc5302f62523aa55a6a8bb596bdda7974316f64/src/Dockerfile#L49

Or just remove `torchao` from that line in the Dockerfile and let this line handle it: https://github.com/allenai/OLMo-core/blob/6bc5302f62523aa55a6a8bb596bdda7974316f64/src/Dockerfile#L93

tyler-romero: https://github.com/allenai/OLMo-core/pull/510/commits/fa8a69169587775208ee6cd296a719fbd5b63ade

src/olmo_core/float8/__init__.py
```diff
@@ -6,14 +6,16 @@
 from dataclasses import dataclass
 from typing import Optional, Set
 
-import torch
 import torch.nn as nn
+from torchao.utils import torch
```
safelix, line 10: This breaks the optional dependency on torchao: without having torchao installed a simple `from olmo_core.train import TrainerConfig` does not work anymore. Maybe the import could be moved into the `quantize_filter_fn()`?

epwalsh: Good catch, fixed: b6987f8b

PR #511, opened 2025-12-22
Bottom-up flops estimation & total flops tracking

tyler-romero:
Organized flops tracking so each module self-reports its flop count. This just makes it easier to support correct flop counts across different models. Anticipating wanting flops for layers like gated deltanet soon.

Corrects FLOPs/token estimation for:
1. Sliding-window attention
2. MoE

Removes incorrect estimation for pipeline parallel models.

Speedmonitor will now log the total flops for the run (similar to total tokens), should be VSL-training aware.

Example run: https://beaker.org/ex/01KD45JTZZD7C5CMXV0RJQDKSK
```
    gpu_memory/GPU active mem (%)=63.31
    gpu_memory/GPU active mem (GiB)=50.14
    gpu_memory/GPU reserved mem (%)=79.17
    gpu_memory/GPU reserved mem (GiB)=62.69
    optim/LR (group 0)=0.0005
    optim/LR (group 1)=0.0005
    optim/step skipped=0.0
    optim/total grad norm=0.5816
    throughput/device/BPS=3.939
    throughput/device/BPS (actual avg)=2.990
    throughput/device/MFU=44.57
    throughput/device/MFU (actual avg)=33.82
    throughput/device/TPS=258,163
    throughput/device/TPS (actual avg)=195,922
    throughput/device/flopsPS=441.0T
    throughput/device/flopsPS (actual avg)=334.7T
    throughput/total flops=89.56P
    throughput/total tokens=52,428,800
    train/CE loss=6.445
    train/Z loss=0.0013
    throughput/device/data loading (%)=0.0701
    throughput/device/data loading (s)=0.0002
    train/PPL=629.6
 ```

### DIFF
--- a/src/olmo_core/nn/moe/moe.py
+++ b/src/olmo_core/nn/moe/moe.py
@@ -325,17 +325,12 @@ class MoEBase(nn.Module):
         )
 
     def num_flops_per_token(self, seq_len: int) -> int:
-        # Router
-        flops = 6 * sum(p.numel() for p in self.router.parameters())
-
-        # Shared MLP
-        if self.shared_mlp is not None:
-            flops += self.shared_mlp.num_flops_per_token(seq_len)
-
-        # Experts
-        flops += self.experts.num_flops_per_token(seq_len)
-
-        return flops
+        router_flops = 6 * sum(p.numel() for p in self.router.parameters())
+        shared_mlp_flops = (
+            self.shared_mlp.num_flops_per_token(seq_len) if self.shared_mlp is not None else 0
+        )
+        expert_flops = self.experts.num_flops_per_token(seq_len)
+        return router_flops + shared_mlp_flops + expert_flops
 
 
 class MoE(MoEBase):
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/olmo_core/train/train_module/transformer/pipeline_train_module.py [hunk 1]
```diff
@@ -596,8 +602,20 @@ def capture_losses(
 
         return ce_batch_loss, z_batch_loss
 
-    def num_flops_per_token(self, seq_len: int) -> int:
-        return self.model_parts[0].num_flops_per_token(seq_len)
+    @lru_cache
```
epwalsh, line 605: The cache here is redundant since `warn_once` also caches.

tyler-romero:
> good reason to ignore layer norm FLOPs

Elementwise ops really are quite small compared to the matmuls, they dont really move the needle and are more costly to correctly account for. We're also omitting activation fns, gating operations, residual adds, CE loss computation, etc so this would not be the only small source of flop bias.

PR #518, opened 2025-12-23
Add scripts for slurm env

epwalsh:


src/scripts/lambda/README.md
```diff
@@ -0,0 +1,64 @@
...
+```bash
+# Set your unique username (ideally use your Ai2/Beaker username).
+username=petew
+
+# Clone repo to unique directory.
+cd "/data/ai2/$username"
+git clone https://github.com/allenai/OLMo-core.git
+cd OLMo-core
+
+# Create unique virtual environment.
```
tyler-romero, line 42: I miss docker already haha

tyler-romero: If running multiple jobs in parallel w/ incompatible dependencies we'll need to modify entrypoint to use separate venvs

epwalsh: Yea.. I figured we probably won't run into that 🤞 But we'll cross that bridge when we come to it.

PR #519, opened 2025-12-24
update slurm scripts

epwalsh:


src/scripts/lambda/README.md
tyler-romero: is beaker token needed or just WANDB?

epwalsh: Typo there. You need W&B API key and `USERNAME` set to your Ai2/Beaker username.

PR #522, opened 2025-12-24
Add temp & mem healthchecks and allow comments in cordoned nodelist

tyler-romero:
Make it easier to track when and why we cordoned certain nodes.

Comment lines start with `#`.

### DIFF
--- a/src/scripts/lambda/healthchecks.sh
+++ b/src/scripts/lambda/healthchecks.sh
@@ -1,7 +1,11 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-# Temperature-based node healthchecks for Lambda/SLURM jobs.
+# Node healthchecks for Lambda/SLURM jobs.
+#
+# Checks:
+# - GPU temperature
+# - GPU idle (process + memory)
 #
 # This script is intended to be run on *every node* before training starts.
 # It will exit non-zero on failure to force the job to fail fast.


CHANGELOG.md
```diff
@@ -43,6 +43,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - All `olmo_core.nn` module configs now inherit from a common base class, `ModuleConfig`.
 - Big changes to `olmo_core.model_ladder` API.
 - Add ngram instance filter to olmo3_ladder.
+- Allow comments in cordoned nodelist for lambda slurm launcher.
```
epwalsh: Changelog is only meant to track changes since the last release, so I would omit this.
```suggestion
```

src/scripts/lambda/README.md
```diff
@@ -35,10 +35,20 @@ cd OLMo-core
 uv venv --python=3.12 /data/ai2/uv/OLMo-core-$USERNAME
 source "/data/ai2/uv/OLMo-core-${USERNAME}/bin/activate"
 uv pip install --torch-backend=cu129 numpy torch==2.8.0 torchvision torchaudio torchao==0.15.0
+uv pip install flash-attn --no-build-isolation
```
epwalsh, line 38: This took like 5 hours for me lol, I probably should have ran it on a compute node.

src/scripts/lambda/entrypoint.sh
```diff
@@ -62,6 +62,10 @@ node_0_only echo "Activating Python virtual environment..."
 source "$VENV_DIR/bin/activate"
 node_0_only uv pip freeze
 
+# Run per-node healthchecks (fails fast on bad nodes).
+node_0_only echo "Running per-node healthchecks..."
+bash src/scripts/lambda/healthchecks.sh
```
epwalsh: If this is meant to stop the job when it fails you'll need:
```suggestion
bash src/scripts/lambda/healthchecks.sh || exit 1
```

epwalsh: Or better yet:
```suggestion
if ! ./src/scripts/lambda/healthchecks.sh; then
    echo "Healthcheck failed on node '$(hostname)'. Consider adding it to the cordoned list by running:"
    echo ""
    echo "  echo $(hostname) >> /data/ai2/cordoned-nodes.txt"
    echo ""
    exit 1
fi
```

Need to `chmod +x` this file though.

src/scripts/lambda/README.md
tyler-romero: Oh really? I must have gotten the cached version after you installed it.

epwalsh: Ahh yea that's probably it

epwalsh (approved):

PR #523, opened 2025-12-24
Consolidate bash helper functions

epwalsh:
And match the logging format with our format from Python.

image

### DIFF
--- a/src/scripts/lambda/launch.sh
+++ b/src/scripts/lambda/launch.sh
@@ -36,7 +36,7 @@ fi
 # done
 
 SBATCH_ARGS=(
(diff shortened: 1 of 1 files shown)


src/scripts/lambda/launch.sh
tyler-romero:
```suggestion
    --export="WANDB_API_KEY,USERNAME,HOME"
```
While you're here...

Pytorch uses HOME to determine where to cache kernels and it seems HOME is expected to be set in `entrypoint.sh` but is not actually passed through.

tyler-romero (approved):

PR #527, opened 2025-12-30
GAP Monitor Improvements

tyler-romero:
Fix bug where GAP monitor would crash if a local tensor shard had 0 elements (when taking max()).

Reduce memory consumption by decorating GAP with `@no_grad`

Allow recording GAP metrics every N steps.

### DIFF
--- a/src/olmo_core/internal/ladder.py
+++ b/src/olmo_core/internal/ladder.py
@@ -84,7 +84,7 @@ def parse_args(
         parser.add_argument(
             "--cluster",
             type=str,
-            choices=["ai2/augusta", "ai2/jupiter", "ai2/titan"],
+            choices=["ai2/augusta", "ai2/jupiter", "ai2/titan", "lambda"],
             default="ai2/augusta",
             help="The Beaker cluster to launch each run on.",
         )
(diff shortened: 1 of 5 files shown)


src/olmo_core/train/callbacks/gap_monitor.py
```diff
@@ -150,10 +155,19 @@ def record_tensor_stats(
...
-            max_ = get_local_tensor(tensor).abs().max()
             var, mean = var_mean(tensor)
+            local_tensor = get_local_tensor(tensor)
+            if local_tensor.numel() > 0:
+                local_max = local_tensor.abs().max()
+            else:
+                # Use 0.0 as sentinel value for empty tensors in max reduction.
+                # Since we're taking abs(), all actual values are >= 0, so 0.0
+                # won't affect the max reduction when other processes have non-empty tensors.
+                local_max = torch.tensor(0.0, device=tensor.device, dtype=tensor.dtype)
```
epwalsh: This way would avoid the host-device sync:
```suggestion
                local_max = torch.zeros([], device=tensor.device, dtype=tensor.dtype)
```

epwalsh (changes requested):
LGTM, just one minor comment

epwalsh (approved):

PR #528, opened 2025-12-30
Support for Muon and Dion Optimizers

tyler-romero:
Uses implementations of Muon and Dion from https://github.com/microsoft/dion, which are great because they (1) support custom triton kernels for the symmul in NS iteration (2) support DDP, FSDP, HSDP, and any of the above with TP and (3) support both "standard" muon scaling and kimi/moonlight scaling.


src/olmo_core/optim/__init__.py
```diff
@@ -1,7 +1,15 @@
...
+from .config import (
+    INITIAL_LR_FIELD,
+    LR_FIELD,
+    MatrixAwareOptimConfig,
+    OptimConfig,
+    OptimGroupOverride,
+)
+from .dion import DionConfig
 from .lion import Lion, LionConfig, SkipStepLion, SkipStepLionConfig
+from .muon import MuonConfig, NorMuonConfig
```
dirkgr, line 12: The way this is structured, we need to have all dependencies present even when we're not using them. This could be particularly bad because I forsee a situation where it's difficult to get some of them to build, and then it blocks us even though we don't care about that particular functionality. Any chance we can avoid that?

src/olmo_core/optim/dion.py
```diff
@@ -0,0 +1,159 @@
...
+            meshes["inner_shard_mesh"] = get_tp_mesh(world_mesh)
+
+        log.info(f"Dion parallelism_config: {meshes}")
+        return meshes
+
+    def create_optimizer(self, model: torch.nn.Module, strict: bool = True, **kwargs):
+        """
+        Create the optimizer.
+        """
+        torch._dynamo.config.recompile_limit = 16
```
epwalsh: What's up with this?

src/olmo_core/optim/muon.py
```diff
@@ -0,0 +1,224 @@
...
+            parallelism_config["distributed_mesh"] = get_dp_model_mesh(world_mesh)
+
+        log.info(f"Muon parallelism_config: {parallelism_config}")
+        return parallelism_config
+
+    def create_optimizer(self, model: torch.nn.Module, strict: bool = True, **kwargs):
+        """
+        Create the optimizer.
+        """
+        torch._dynamo.config.recompile_limit = 16
```
epwalsh: Same question here

pyproject.toml
```diff
@@ -58,6 +58,7 @@ dev = [
 ]
 beaker = ["beaker-py>=1.37.0,<2.0", "GitPython>=3.0,<4.0", "google-cloud-compute"]
 comet = ["comet_ml"]
+dion = ["dion @ git+https://github.com/microsoft/dion.git ; sys_platform == 'linux'"]
```
epwalsh: Should add this to the "all" extra below.

src/olmo_core/optim/dion.py
tyler-romero: When torch.compile() triggers recompilation too many times, it throws an exception rather than proceed with too many recompilations. Typically on the second recompilation it attempts to compile a dynamic version of the op, unless `dynamic=False` is marked, like here: https://github.com/microsoft/dion/blob/b17a7699f1398b9cba69d760a52c79bc0ee0e420/dion/newton_schulz_triton.py#L306

https://docs.pytorch.org/docs/stable/compile/programming_model.recompilation.html

So essentially on the first step we are recompiling the newton_schultz function for each different grad matrix shape in our model. The default is 8, so I bumped it to 16.

pyproject.toml
tyler-romero: [`aa3a09f` (#528)](https://github.com/allenai/OLMo-core/pull/528/commits/aa3a09f9afd2e80fd166a97c45083d28e101912e)

src/olmo_core/optim/config.py
```diff
@@ -222,3 +225,159 @@ def reset_fixed_fields(opt: torch.optim.Optimizer):
...
+    @classmethod
+    def optimizer(cls) -> Type[Opt]:
+        raise NotImplementedError
+
+    def categorize_parameters(self, model: nn.Module) -> Dict[str, List[str]]:
+        assert isinstance(model, Transformer)
+        embed_params = [f"embeddings.{n}" for n, p in model.embeddings.named_parameters()]
+        matrix_params = [f"blocks.{n}" for n, p in model.blocks.named_parameters() if p.ndim >= 2]
+        vector_params = [f"blocks.{n}" for n, p in model.blocks.named_parameters() if p.ndim < 2]
+        vector_params += [f"lm_head.{n}" for n, p in model.lm_head.named_parameters() if p.ndim < 2]
```
dirkgr, line 248: I'm assuming there are none of these?

src/olmo_core/optim/config.py
```diff
@@ -222,3 +225,159 @@ def reset_fixed_fields(opt: torch.optim.Optimizer):
...
+        lm_head_params = [
+            f"lm_head.{n}" for n, p in model.lm_head.named_parameters() if p.ndim >= 2
+        ]
+
+        # Assert all parameters are categorized
+        all_model_params = {n for n, p in model.named_parameters() if p.requires_grad}
+        categorized_params = set(embed_params + matrix_params + vector_params + lm_head_params)
+        uncategorized = all_model_params - categorized_params
+        assert not uncategorized, f"Uncategorized parameters: {uncategorized}"
```
dirkgr, line 264: There is no issue if a param is in multiple categories?

src/olmo_core/optim/config.py
```diff
@@ -222,3 +225,159 @@ def reset_fixed_fields(opt: torch.optim.Optimizer):
...
+            match any parameter.
+        """
+
+        kwargs = self.as_dict(exclude_private_fields=True)
+        group_overrides = kwargs.pop("group_overrides")
+        if group_overrides is not None:
+            log.warning(f"group_overrides is set but will be ignored for {self.__class__.__name__}")
+        compile_opt = kwargs.pop("compile")
+        if compile_opt:
+            log.warning(f"compile is set to True but will be ignored for {self.__class__.__name__}")
```
dirkgr, line 347: We never compile the optimizer?

src/olmo_core/optim/dion.py
dirkgr: Can you explain that in a comment? It's a little sketchy that we're setting a global thing so deep in here.

src/scripts/train/OLMo3/OLMo3-7B-midtraining.py
```diff
@@ -25,7 +25,7 @@ def build_experiment_config(cli_context: CliContext) -> ExperimentConfig:
     )
     root_dir = get_root_dir(cli_context.cluster)
     work_dir = get_work_dir(root_dir)
-    save_dir = f"{root_dir}/checkpoints/{run_name_with_ts}"
+    save_dir = f"{root_dir}/checkpoints/{cli_context.run_name}"
```
dirkgr, line 28: Orthogonal fix?

src/olmo_core/optim/dion.py
epwalsh: Also should probably do something like 
```suggestion
        torch._dynamo.config.recompile_limit = max(torch._dynamo.config.recompile_limit, 16)
```

src/olmo_core/optim/config.py
tyler-romero: Our lmhead contains a layernorm, which can optionally contain parameters

src/olmo_core/optim/config.py
tyler-romero: Good call, added check.

src/olmo_core/optim/config.py
tyler-romero: We dont allow control over it for these ones. They all do mandatory/scoped torch.compile calls internally. Or they use custom triton kernels.

src/scripts/train/OLMo3/OLMo3-7B-midtraining.py
tyler-romero: Yes, known bug that I saw and wanted to make sure I fixed this time

src/scripts/train/OLMo3/OLMo-3-190M.py
```diff
@@ -0,0 +1,189 @@
+from datetime import datetime
```
tyler-romero, line 1: Added this to make it easier to hack on things

src/olmo_core/optim/config.py
tyler-romero: [`cf71f75` (#528)](https://github.com/allenai/OLMo-core/pull/528/commits/cf71f758c817390552f7be63fe5167c060a46e6b)

src/olmo_core/optim/dion.py
tyler-romero: [`cf71f75` (#528)](https://github.com/allenai/OLMo-core/pull/528/commits/cf71f758c817390552f7be63fe5167c060a46e6b)

src/olmo_core/optim/__init__.py
tyler-romero: Dion is already imported lazily everywhere it is used: https://github.com/allenai/OLMo-core/blob/cf71f758c817390552f7be63fe5167c060a46e6b/src/olmo_core/optim/muon.py#L23

So its possible to import the DionConfig without importing the Dion package. Only when `opt.build()` is called is dion imported.

dirkgr: 👍🏻

src/olmo_core/optim/config.py
dirkgr: Ah, yes.

PR #532, opened 2026-01-04
Ulysses CP Support

tyler-romero:
Ulysses CP support. Ulysses-style CP is convenient because it typically does *not* require modifications to the underlying attention mechanism. Instead, differentiable all-to-all collectives are inserted around the attention mechanism to transition from CP-sharding to Head-sharding and back. 

image

This means we can support CP with more attention backends, such as flash_3. Additionally, TE already has native support for deepseed ulysses in addition to ring. One of the primary motivations for this PR is to lay the groundwork to support ulysses-style CP for flash-linear-attention layers like GatedDeltaNet.

Ulysses and Ring CP are not mutually exclusive! Although not implemented here, in the future modifications could be made to apply both at once.

## Perf
Ulysses is typically a bit faster than the megatron/llama3-style all-gather CP. I havent found a comparison to ring-flash-attn yet, which better overlaps compute and comms.

image

### DIFF
--- a/src/test/nn/transformer/model_test.py
+++ b/src/test/nn/transformer/model_test.py
@@ -1,5 +1,6 @@
 import logging
 from dataclasses import replace
+from test.nn.attention_test import BF16_ATOL, BF16_RTOL
 from typing import Optional, cast
 
 import pytest
@@ -54,7 +55,6 @@ from olmo_core.testing import (
     run_distributed_test,
 )
 from olmo_core.utils import get_default_device, seed_all
-from test.nn.attention_test import BF16_ATOL, BF16_RTOL
 
 log = logging.getLogger(__name__)
 


tyler-romero:
TODO:
1. Need to implement ulysses load balancer - even splitting
2. Need to pass through full (unsharded) cu_seq_lens
3. Size mismatch on output of flash_2 backend

src/olmo_core/distributed/autograd.py
```diff
@@ -0,0 +1,72 @@
...
+
+    The backward pass performs a reverse all-to-all with swapped split sizes to
+    correctly propagate gradients.
+    """
+
+    @staticmethod
+    def forward(
+        ctx,
+        group: torch.distributed.ProcessGroup,
+        iinput: torch.Tensor,
```
dirkgr: Why the `ii`?

src/olmo_core/distributed/autograd.py
```diff
@@ -0,0 +1,72 @@
+import torch
+
+from olmo_core.distributed.utils import get_world_size
+
+
+class _AllToAll(torch.autograd.Function):
```
dirkgr: I'm surprised this is not part of torch already.

src/olmo_core/nn/attention/__init__.py
```diff
@@ -323,15 +327,11 @@ def apply_tp(
     def apply_cp(
         self,
         cp_mesh: DeviceMesh,
-        load_balancer: RingAttentionLoadBalancerType,
-        head_stride: int = 1,
+        ring: Optional[RingContextParallelStyle] = None,
+        uly: Optional[UlyssesContextParallelStyle] = None,
```
dirkgr, line 333: Can you combine them?

src/olmo_core/nn/attention/backend.py
```diff
@@ -298,6 +328,14 @@ def forward(
                 f"'{self.__class__.__name__}' doesn't support intra-document masking"
             )
 
+        if self.cp_enabled and self.uly is not None:
+            assert self.cp_pg is not None
+            # Transform from context-parallel to head-parallel partitioning
+            # [B, T/CP, H, D] -> [B, T, H/CP, D]
+            q = all_to_all_cp2hp(q, self.cp_pg)
+            k = all_to_all_cp2hp(k, self.cp_pg)
+            v = all_to_all_cp2hp(v, self.cp_pg)
```
dirkgr, line 337: This is a little unfortunate with the structure we have here, but in principle we can get some overlap here by doing the k projection while q is being distributed.

src/olmo_core/train/train_module/transformer/config.py
```diff
@@ -178,32 +180,43 @@ class TransformerContextParallelConfig(ContextParallelConfig):
...
-    The stride of the head dimension to process for each iteration of ring attention. A value of 1
-    means each iteration will process one k and one v head. A value of 2 will process two k and two
-    v heads, etc. A larger stride will reduce the number of communication ops.
-    """
+    def __post_init__(self):
+        if self.ring is not None and self.uly is not None:
+            raise NotImplementedError(
+                "Only one of ring or ulysses can be specified. While not technically "
+                "mutually exclusive, a combined context parallel style is not yet supported."
+            )
```
dirkgr, line 193: That's my answer right there.

src/olmo_core/distributed/autograd.py
tyler-romero: `input` is a python keyword

src/olmo_core/nn/attention/backend.py
tyler-romero: Yeah this is not optimized at the moment. I'd be curious to run a profiler to see what is happening here. I wouldnt be surprised if that is already happenin.

src/olmo_core/distributed/autograd.py
tyler-romero: Good intuition, it actually is. Simplified things.

src/olmo_core/distributed/parallel/context_parallel.py
```diff
@@ -13,3 +21,105 @@ class ContextParallelConfig(Config):
...
+
+    exchanged = all_to_all_single(cp_group, input_split)
+
+    # [CP, B, T/CP, H/CP, D] -> [B, T/CP, CP, H/CP, D] -> [B, T/CP, H, D]
+    exchanged = exchanged.view(world_size, B, t_out, h_in, d_in).permute(1, 2, 0, 3, 4)
+    exchanged = exchanged.reshape(B, t_out, h_in * world_size, d_in)
+    return exchanged
+
+
+class UlyssesContextParallelStyle(Config):
```
epwalsh: Configs should be dataclasses

```suggestion
@dataclass
class UlyssesContextParallelStyle(Config):
```

src/olmo_core/distributed/parallel/context_parallel.py
```diff
@@ -13,3 +21,105 @@ class ContextParallelConfig(Config):
...
+    """
+    Configuration for Ulysses-style context parallelism.
+    """
+
+    @property
+    def load_balancer(self) -> "RingAttentionLoadBalancerType":
```
epwalsh: Why not make this a field instead of a property to be consistent with `RingContextParallelStyle`?

src/olmo_core/nn/attention/backend.py
```diff
@@ -483,26 +533,64 @@ def forward(
...
-                max_seqlen_k=max_doc_len_k,
-                heads_k_stride=self.cp_head_stride,
-                local_k_slice=local_k_slice,
-                dropout_p=self.dropout_p,
-                causal=True,
-                softmax_scale=self.scale,
-                window_size=self.window_size,
-            )
+            if self.ring is not None:
+                assert self.cp_pg is not None and self.cp_load_balancer is not None
```
epwalsh: As far as I can tell `self.cp_load_balancer` is not being set anymore so this will error.

src/olmo_core/distributed/parallel/context_parallel.py
tyler-romero: The thought here is that this shouldn't be modifiable for ulysses. It requires this particular load balancer, which handles cu_doc_lengths differently than the others.

PR #533, opened 2026-01-05
Add Qwen3 model support (0.6B-32B)

finbarrtimbers:
## Summary

- Add support for Qwen3 dense models (0.6B, 1.7B, 4B, 8B, 14B, 32B) to OLMo-core
- Add explicit `head_dim` parameter to `AttentionConfig` and `Attention` class to support Qwen3's non-standard architecture where `head_dim=128` differs from `d_model/n_heads`
- Add HuggingFace weight conversion mappings for Qwen3
- Add training script for Qwen3-8B
- Add comprehensive tests including HuggingFace output comparison

### Architecture Details

Qwen3 uses a Llama-like architecture with key differences:
- All models use explicit `head_dim=128` (not derived from d_model/n_heads)
- Per-head QK normalization (`use_head_qk_norm=True`)
- RoPE theta of 1,000,000
- RMS norm epsilon of 1e-6

| Model | d_model | layers | heads | kv_heads | head_dim | intermediate_size |
|-------|---------|--------|-------|----------|----------|-------------------|
| 0.6B  | 1024    | 28     | 16    | 8        | 128      | 3072              |
| 1.7B  | 2048    | 28     | 16    | 8        | 128      | 6144              |
| 4B    | 2560    | 36     | 32    | 8        | 128      | 9728              |
| 8B    | 4096    | 36     | 32    | 8        | 128      | 12288             |
| 14B   | 5120    | 48     | 40    | 8        | 128      | 17408             |
| 32B   | 5120    | 64     | 40    | 8        | 128      | 25600             |

## Test plan

- [x] Local tests pass (`pytest -v src/test/nn/hf/qwen3_test.py`)
- [x] Beaker GPU tests pass (4/4 tests including HF comparison)
- [x] `test_qwen3_matches_huggingface` verifies output matches HuggingFace Qwen/Qwen3-0.6B (rtol=1e-4, atol=1e-4)
- [x] Ruff checks pass

### DIFF
--- a/src/olmo_core/nn/transformer/config.py
+++ b/src/olmo_core/nn/transformer/config.py
@@ -1166,7 +1166,9 @@ class TransformerConfig(ModelConfig):
             layer_norm_eps=1e-6,
             qk_norm=kwargs.pop("qk_norm", True),
             use_head_qk_norm=kwargs.pop("use_head_qk_norm", True),
-            feed_forward=FeedForwardConfig(hidden_size=3072, bias=False),
+            feed_forward=FeedForwardConfig(
+                hidden_size=3072, bias=False, dtype=kwargs.get("dtype", DType.float32)
+            ),
             **kwargs,
         )
 
@@ -1183,7 +1185,9 @@ class TransformerConfig(ModelConfig):
             layer_norm_eps=1e-6,
             qk_norm=kwargs.pop("qk_norm", True),
             use_head_qk_norm=kwargs.pop("use_head_qk_norm", True),
-            feed_forward=FeedForwardConfig(hidden_size=6144, bias=False),
+            feed_forward=FeedForwardConfig(
+                hidden_size=6144, bias=False, dtype=kwargs.get("dtype", DType.float32)
+            ),
             **kwargs,
         )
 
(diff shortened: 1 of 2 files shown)


src/olmo_core/nn/transformer/config.py
```diff
@@ -1153,6 +1153,108 @@ def llama3_405B(
             **kwargs,
         )
 
+    @classmethod
+    def qwen3_0_6B(cls, vocab_size: int = 151936, **kwargs) -> "TransformerConfig":
```
epwalsh: Is there a reason to provide a default vocab here size? We don't with the other classmethod constructors to avoid accidentally using the wrong tokenizer.

src/test/nn/hf/qwen3_test.py
```diff
@@ -0,0 +1,118 @@
...
+def test_qwen3_forward_pass():
+    config = TransformerConfig.qwen3_0_6B(vocab_size=1024, n_layers=2)
+    model = config.build(init_device="cpu")
+    model.eval()
+
+    input_ids = torch.randint(0, 1024, (2, 16))
+    with torch.no_grad():
+        logits = model(input_ids)
+
+    assert logits.shape == (2, 16, 1024)
```
epwalsh: This test seems redundant with what's below, and it's not testing anything in `nn.hf` anyway, so I suggest removing it.

epwalsh (commented):
Overall looks great, just a couple minor suggestions!

src/olmo_core/nn/transformer/config.py
finbarrtimbers: Ah, great point! Removed.

src/test/nn/hf/qwen3_test.py
finbarrtimbers: Good point. Done.

epwalsh (approved):

PR #534, opened 2026-01-05
Add Gemma 3 support with HF weight conversion

finbarrtimbers:
## Summary
- Add `gemma3_like` factory method for creating Gemma 3-style model configurations
- Add explicit `head_dim` support to `AttentionConfig` and `Attention` class (Gemma 3 uses head_dim=256 independent of d_model/n_heads)
- Add `embed_scale` parameter for embedding scaling (sqrt(d_model))
- Use `PeriNormTransformerBlock` for Gemma 3's pre+post norm architecture (4 norms per layer)
- Add HF weight mappings for `gemma3_text` model type
- Add norm weight transformation (+1) for Gemma 3's zero-initialized RMSNorm convention
- Add GeGLU activation support (GELU with gating)
- Add tests for loading real HF Gemma 3 models (google/gemma-3-270m, google/gemma-3-1b-pt)

## Key Architectural Differences Handled
1. **Embedding scaling**: HF Gemma 3 scales embeddings by `sqrt(hidden_size)`
2. **Peri-norm block structure**: Pre-norm + post-norm for both attention and FF (4 norms per layer)
3. **RMSNorm initialization**: HF uses zero-init with `(1 + weight)`, OLMo uses ones-init with `weight`
4. **Explicit head_dim**: Gemma 3 uses head_dim=256 regardless of d_model, allowing "fat attention"

## Test plan
- [x] Existing HF checkpoint tests pass
- [x] `test_load_hf_gemma3_270m_pretrained` passes with max_diff=0.0
- [x] `test_load_hf_gemma3_1b_pretrained` passes with max_diff=0.0
- [x] All 8 checkpoint tests pass

Successful run of the golden tests: [Beaker](https://beaker.allen.ai/orgs/ai2/workspaces/OLMo-core/work/01KF1P48P0GZFHMYC37AKH9Y8N?)

### DIFF
--- a/src/scripts/beaker/launch_test.py
+++ b/src/scripts/beaker/launch_test.py
@@ -31,7 +31,6 @@ def build_config(command: List[str], overrides: List[str]) -> BeakerLaunchConfig
         num_nodes=1,
         num_gpus=2,
         shared_filesystem=True,
-        #  host_networking=False,
         torchrun=False,
         env_secrets=[
             BeakerEnvSecret(name="HF_TOKEN", secret="HF_TOKEN"),


src/olmo_core/nn/attention/__init__.py
```diff
@@ -175,6 +175,7 @@ class AttentionConfig(ModuleConfig):
     """
     n_heads: int = 16
     n_kv_heads: Optional[int] = None
+    head_dim: Optional[int] = None
```
tyler-romero, line 184: head_dim should be implicit from `d_model` and `n_heads`. Or am I missing something? Whats the purpose of setting head_dim to something other than `d_model // n_heads`?

src/olmo_core/nn/hf/convert.py
```diff
@@ -337,6 +348,33 @@ def get_converter_from_hf(model_type: str | None = None) -> StateConverter:
     return _get_converter_from_hf(model_type=model_type)
 
 
+def _apply_gemma3_norm_transform(state: Dict[str, Any]) -> Dict[str, Any]:
```
tyler-romero, line 357: Gemma and their norms...

They also use a different casting location, how are we  looking on that front?
https://github.com/linkedin/Liger-Kernel/blob/main/src/liger_kernel/ops/rms_norm.py#L596

src/olmo_core/nn/transformer/config.py
```diff
@@ -1276,6 +1341,105 @@ def ngpt_like(
             **kwargs,
         )
 
+    @classmethod
+    def gemma3_like(
```
tyler-romero, line 1588: 👍

src/olmo_core/nn/feed_forward.py
```diff
@@ -90,6 +94,7 @@ def build(
             if self.name == FeedForwardType.default:
                 return FeedForward(**kwargs)
             elif self.name == FeedForwardType.normalized:
+                kwargs.pop("activation", None)
```
tyler-romero: What sort of activation does NormalizedFeedForward use? could we take an activation arg and just raise an error if its not the right one? Rather than silently being not what is specififed.

src/olmo_core/nn/feed_forward.py
```diff
@@ -48,6 +48,10 @@ class FeedForwardConfig(ModuleConfig):
     """
     bias: Optional[bool] = None
     dtype: Optional[DType] = None
+    activation: str = "silu"
```
tyler-romero: We love enums here haha can we have an enum to make sure the user is specifying either silu or gelu_tanh and not making a typo. This means we can do fewer `else: raise ValueError()` clauses.

src/olmo_core/nn/feed_forward.py
```diff
@@ -112,10 +117,12 @@ def __init__(
         bias: bool = True,
         dtype: torch.dtype = torch.float32,
         init_device: str = "cpu",
+        activation: str = "silu",
```
tyler-romero: same comment about piping through an enum

src/olmo_core/nn/feed_forward.py
```diff
@@ -126,6 +133,8 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
 
         :param x: The input of shape ``(*, d_model)``.
         """
+        if self.activation == "gelu_tanh":
+            return self.w2(F.gelu(self.w1(x), approximate="tanh") * self.w3(x))
         return self.w2(F.silu(self.w1(x)) * self.w3(x))
```
tyler-romero: We could probably even have the activation be a function, so this reduces to:
```python
return self.w2(self.activation_fn(self.w1(x)) * self.w3(x))
```

src/test/nn/hf/checkpoint_test.py
```diff
@@ -1,12 +1,28 @@
...
+    Olmo2Config,
+)
 
 from olmo_core.nn.hf.checkpoint import load_hf_model, save_hf_model
 from olmo_core.nn.transformer.config import TransformerConfig
 
+log = logging.getLogger(__name__)
+
+HF_TOKEN = os.environ.get("HF_TOKEN")
+requires_hf_token = pytest.mark.skipif(
```
tyler-romero: Can we push this into `olmo_core/testing/utils.py`?

src/olmo_core/nn/feed_forward.py
tyler-romero: One pattern we like in particular are these configs that build themselves, like: https://github.com/allenai/OLMo-core/blob/64ebcf5babc1e6b6d8a692908181ed8144063add/src/olmo_core/nn/attention/backend.py#L77

Sort of dependency-injection like.

src/olmo_core/nn/attention/__init__.py
finbarrtimbers: Well, for whatever reason, this isn't the case for Gemma 3 and Qwen 3! They have them defined explicitly. 

E.g. [Gemma 3 270M](https://huggingface.co/google/gemma-3-270m/blob/main/config.json), which has `"head_dim": 256,` `"hidden_size": 640`, and `"num_attention_heads": 4`.

src/test/nn/hf/checkpoint_test.py
finbarrtimbers: Great suggestion! Done.

src/olmo_core/nn/feed_forward.py
finbarrtimbers: Ah, very nice. Done.

src/olmo_core/nn/hf/convert.py
finbarrtimbers: I believe that the current behaviour of the `RMSNorm` class (when `full_precision` is `True`) matches Gemma exactly, iiuc. 

That makes me think that our Llama implementation isn't correct, though... But I haven't looked into that extensively.

src/olmo_core/nn/feed_forward.py
finbarrtimbers: Perfect! Done.

src/olmo_core/nn/feed_forward.py
finbarrtimbers: Fixed!

finbarrtimbers:
I think we're good to go here, I think the failing tests are not due to this PR?

src/scripts/beaker/run_gpu_pytest.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+set -eo pipefail
+
+TEST_PATH="${1:-src/test/nn/hf/golden_tests.py}"
```
tyler-romero: nit: can we make this just like running pytest on beaker? So all arguments are passed to `pytest ...` on the final line.

src/scripts/beaker/launch_test.py
```diff
@@ -31,7 +31,6 @@ def build_config(command: List[str], overrides: List[str]) -> BeakerLaunchConfig
         num_nodes=1,
         num_gpus=2,
         shared_filesystem=True,
-        #  host_networking=False,
```
tyler-romero, line 34: Why did this get commented out?

finbarrtimbers: Came from @epwalsh: https://github.com/allenai/OLMo-core/pull/159

src/olmo_core/nn/feed_forward.py
finbarrtimbers: Done!

src/scripts/beaker/run_gpu_pytest.sh
finbarrtimbers: Great idea! Done.

PR #536, opened 2026-01-07
Add 60M and 1M param model configs

tyler-romero:
Per request from @mayeechen @soldni @kyleclo 

Note that the 1M is almost all lm_head params. And the sizes for all models are predicated on the dolma2 tokenizer vocab size.

```
Using vocab_size=100,352

Config                   Params      Non-Embedding         LM Head
------------------------------------------------------------------
olmo2_1M              2,447,820          1,243,596       1,204,224
olmo2_30M            54,792,448         29,102,336      25,690,112
olmo2_60M            95,957,376         57,422,208      38,535,168
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added support for tracking total flops.
 - Added support for Muon and Dion optimizers.
 - Added support for Ulysses-style context parallelism.
+- Added 60M and 1M model sizes.
 
 ### Fixed
 


src/olmo_core/nn/transformer/config.py
```diff
@@ -458,6 +458,21 @@ def num_active_non_embedding_params(self) -> int:
         """
         return self.num_active_params - self.d_model * self.vocab_size
 
+    @classmethod
+    def olmo2_1M(cls, vocab_size: int, **kwargs) -> "TransformerConfig":
+        return cls.llama_like(
+            d_model=12,
```
epwalsh, line 464: this is ludicrously small 😆

tyler-romero: Yeah the paper we're getting `1M` from is off by 12x on the size of their "1M" model haha
image

Just the lm_head would have 12M params

PR #537, opened 2026-01-07
Fix flaky and broken transformer model tests

tyler-romero:
Flaky test due to low precision and randomness: test_context_parallel_transformer_ulysses

Broken tests due to OOM: test_gemma3_builder_configs, test_qwen3_builder_configs

### DIFF
--- a/src/test/nn/transformer/model_test.py
+++ b/src/test/nn/transformer/model_test.py
@@ -1,5 +1,6 @@
 import logging
 from dataclasses import replace
+from test.nn.attention_test import BF16_ATOL, BF16_RTOL
 from typing import Optional, cast
 
 import pytest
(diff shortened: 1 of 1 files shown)


src/test/nn/transformer/model_test.py
```diff
@@ -592,8 +597,8 @@ def test_gemma3_builder_configs(config_builder, expected_d_model, expected_n_lay
     assert config.block.attention.rope is not None
     assert config.block.attention.rope.theta == 10_000
 
-    model = config.build(init_device="cpu")
-    model.init_weights(device=torch.device("cpu"))
+    # Use meta device to avoid allocating large amounts of memory for big models.
+    model = config.build(init_device="meta")
```
tyler-romero, line 601: These were hitting an OOM on the GH test runner

finbarrtimbers: Ty for fixing!

PR #538, opened 2026-01-08
Add a stepped WSDS variant, integrated into ladder

epwalsh:
For example:
image

Compared to the default:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added support for Muon and Dion optimizers.
 - Added support for Ulysses-style context parallelism.
 - Added 60M and 1M model sizes.
+- `SpeedMonitorCallback` will log Chinchilla multiple number of tokens during training with a `TransformerTrainModule`.
 
 ### Fixed
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,7 +56,7 @@ dev = [
     "twine>=1.11.0",
     "wheel",
 ]
-beaker = ["beaker-py>=2.5.3,<3.0", "beaker-gantry>=3.4.0,<4.0", "GitPython>=3.0,<4.0", "google-cloud-compute"]
+beaker = ["beaker-py>=2.5.4,<3.0", "beaker-gantry>=3.4.3,<4.0", "GitPython>=3.0,<4.0", "google-cloud-compute"]
 comet = ["comet_ml"]
 dion = ["dion @ git+https://github.com/microsoft/dion.git@b17a7699f1398b9cba69d760a52c79bc0ee0e420 ; sys_platform == 'linux'"]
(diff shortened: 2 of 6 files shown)


dirkgr:
Why is it constant at the first step?

src/olmo_core/model_ladder/wsds_chinchilla_run_configurator.py
```diff
@@ -37,6 +37,12 @@ class WSDSChinchillaRunConfigurator(RunConfigurator):
     """The number of tokens per parameter to use for Chinchilla calculations."""
     lr_multiplier: float = 1.0
     """A multiplier to apply to the learning rate calculated from Chinchilla scaling laws."""
+    stepped_schedule: bool = False
+    """
+    If ``True``, use a stepped schedule for the peak learning rate instead of a constant one,
+    where the peak learning rate will be scaled down by ``1 / sqrt(D)`` during each stage, where
+    ``D`` is the target chinchilla multiple of the stage.
```
dirkgr, line 44: We're doing this in terms of chinchillas, not tokens? Let me think through this. The LR we use for a stage is `lr * (sqrt(#tokens_in_first_stage) / sqrt(#cumulative_tokens_in_this_stage))`? Then it doesn't matter. If it's `lr / sqrt(#cumulative_tokens_in_this_stage)`, then it does matter.

src/olmo_core/model_ladder/wsds_chinchilla_run_configurator.py
```diff
@@ -99,8 +105,12 @@ def configure_chinchilla_periods(self, num_params: int) -> tuple[int, list[float
     def configure_lr_scheduler(self, num_params: int, batch_size: int) -> Scheduler:
         warmup, chinchilla_periods = self.configure_chinchilla_periods(num_params)
         period_lengths = []
+        period_lr_multipliers = [] if self.stepped_schedule else None
         for pidx, c in enumerate(chinchilla_periods):
             period = self._chinchilla_duration(num_params, batch_size, c).value
+            if self.stepped_schedule:
+                assert period_lr_multipliers is not None
+                period_lr_multipliers.append(1 / math.sqrt(max(c, 1.0)))
```
dirkgr: Looks like it does matter that it's chinchillas, not tokens. I'm not sure which one is right. I think in the correct formulation of this, it should not matter what the unit is.

dirkgr (changes requested):
I think there is a problem with this, but I'm not 100% sure.

epwalsh:
> Why is it constant at the first step?

@dirkgr the first decay is at 0.5xC. I choose to keep the LR constant until 1.0xC. Alternatively we could have a higher LR up to 0.5xC.

dirkgr:
I think we should have the higher LR up to 0.5xC.

epwalsh:
Comparison posted here: https://wandb.ai/ai2-llm/olmo3-baseline-ladder/reports/Stepped-Ladder-Overlay--VmlldzoxNTYxNzEyOQ

PR #540, opened 2026-01-09
update to Beaker-py v2, use gantry API internally

epwalsh:


src/olmo_core/launch/beaker.py
```diff
@@ -914,13 +760,24 @@ def _parse_args():
...
+        type=str,
+        nargs="*",
+        help="""Cluster tag constraints (multiple allowed).""",
+    )
     parser.add_argument(
         "--priority",
-        choices=[p.value for p in Priority],
-        default=Priority.normal,
+        choices=["low", "normal", "high", "urgent"],
+        default="low",
```
tyler-romero, line 830: Why different default priority here than above and in ladder.py?

src/scripts/beaker/launch_test.py
```diff
@@ -46,10 +46,10 @@ def build_config(command: List[str], overrides: List[str]) -> BeakerLaunchConfig
     sep_index = sys.argv.index("--")
     overrides = sys.argv[1:sep_index]
     entrypoint = sys.argv[sep_index + 1]
-    command = sys.argv[sep_index + 2 :]
+    command = sys.argv[sep_index + 1 :]
```
tyler-romero, line 49: Is this correct? is entrypoint above unused now?

src/olmo_core/launch/beaker.py
epwalsh: If you run this with all the defaults (no cluster/hostname constraints) then Beaker won't allow it at normal priority, it requires low.

src/scripts/beaker/launch_test.py
epwalsh: yes, forgot to delete it thought

PR #543, opened 2026-01-12
Add 60M and 100M param ladder rungs

tyler-romero:
Cheap way to get more data to fit accurate scaling laws. Plus lets us get a sense for if our 190M rung is well optimized. In It looks like the 0.5xC rung of the 190M ladder is a bit off.
image


epwalsh (approved):
LGTM!

PR #549, opened 2026-01-14
improve beaker client caching

epwalsh:
# Overview

Consolidates Beaker client management into a single, thread-safe caching mechanism via `get_beaker_client()`. Previously, `BeakerLaunchConfig` and `BeakerCallback` each created and managed their own clients. Now they share
  a thread-local cache that reuses clients for the same workspace within a context.

# All Changes
  • Smarter client caching in `get_beaker_client()` (reuses clients by workspace with fallback to last-used workspace).
  • Made the beaker client cache thread-local for safety.
  • Bumped beaker-py to ≥2.5.4 (includes a bug fix for handling updates to the default workspace) and beaker-gantry to ≥3.4.2 (so we can reuse the same client when calling `Recipe.launch()`).
  • Added tests for the new caching behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added support for Muon and Dion optimizers.
 - Added support for Ulysses-style context parallelism.
 - Added 60M and 1M model sizes.
+- `SpeedMonitorCallback` will log Chinchilla multiple number of tokens during training with a `TransformerTrainModule`.
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


src/olmo_core/launch/beaker.py
```diff
@@ -468,125 +475,127 @@ def launch(
...
-        if torchrun is None:
-            if self.num_gpus > 1 or (self.num_gpus >= 1 and self.num_nodes > 1):
-                torchrun = True
-            else:
-                torchrun = False
-
-        if self.git.is_dirty and not self.allow_dirty:
-            raise RuntimeError(
-                "You have uncommitted changes! Set 'allow_dirty=True' in your launch config to force."
+        with get_beaker_client(workspace=self.workspace) as beaker:
```
epwalsh, line 478: Only changes below are just indentation.

PR #553, opened 2026-01-16
Add model ladders for recently run ablations

tyler-romero:
Checking in a bunch of recently run ladders. These can be removed later if theres too much clutter but for now I think they're useful.

### DIFF
--- a/src/test/nn/transformer/model_test.py
+++ b/src/test/nn/transformer/model_test.py
@@ -284,6 +284,7 @@ def run_context_parallel_transformer_ring(checkpoint_dir, outputs_path, architec
 @pytest.mark.parametrize("architecture", ["olmo2"])
 @pytest.mark.skip("known precision issues with ring-flash-attn")
 def test_context_parallel_transformer_ring(architecture: str, tmp_path):
+    seed_all(0)
     device = torch.device("cuda")
     config = get_transformer_config(architecture, dtype=torch.bfloat16)
     config.block.attention.use_flash = True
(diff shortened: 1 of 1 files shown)


src/olmo_core/launch/beaker.py
```diff
@@ -514,7 +514,6 @@ def launch(
                     start_timeout=launch_timeout,
                     inactive_timeout=step_timeout,
                     inactive_soft_timeout=step_soft_timeout,
-                    client=beaker,
```
tyler-romero, line 517: unrelated fix

src/olmo_core/model_ladder/base.py
```diff
@@ -672,7 +672,5 @@ def _get_in_loop_eval_tasks(self) -> list[str]:
     def _get_mix_base_dir(self) -> str:
         if self.dir.startswith("/weka/"):
             return "/weka/oe-training-default/ai2-llm"
-        elif self.dir.startswith("/data/ai2/"):  # lambda cluster
-            return "/data/caia-mltrain/data/"
```
dirkgr, line 676: We still need this? Is this coming back ever?

dirkgr (changes requested):
Other than this one thing, I'm very happy to have this.

src/olmo_core/launch/beaker.py
epwalsh: Wait, this should be here, but you might need to upgrade your gantry version.

src/olmo_core/model_ladder/base.py
tyler-romero: I kind of doubt it? If it does come back, it prob wont be this path either

src/olmo_core/launch/beaker.py
tyler-romero: Ah yeah that fixed it

PR #554, opened 2026-01-16
Add utilities for basic fitting of chinchilla parametric scaling laws

tyler-romero:
Fitting of scaling laws follows the original chinchilla methodology + known helpful improvements closely.

Important points:
1. It seems many scaling law papers dont do a full grid search over initialization params, which can lead to suboptimal fits.
2. Since chinchilla scaling laws are intended to fit the best-achievable loss for a given `N, D` or `C`, using `overestimate_penalty=10` can be helpful to compensate against undertuned ladder rungs. Empirically I have found that using this leads to better predictions of held out points when doing a rollout over increasing `N`.
3. Bootstrapping is useful to get a feel for the plausible variance in predictions based on minor differences in underlying data but I would not consider it robust yet.


src/olmo_core/model_ladder/analysis/eval.py
```diff
@@ -0,0 +1,524 @@
...
+        :param min_groups_train: Minimum number of unique groups in training set.
+        :param **fit_kwargs: Additional arguments passed to ``fit_fn``.
+        :returns: ScalingLawRollout instance with computed splits.
+        """
+        N_arr = np.asarray(N)
+        D_arr = np.asarray(D)
+        L_arr = np.asarray(L)
+        weights_arr = np.asarray(weights) if weights is not None else None
+
+        if len(N_arr) != len(D_arr) != len(L_arr):
```
epwalsh: This is equivalent to:
```
        if len(N_arr) != len(D_arr) and len(D_arr) != len(L_arr):
```

But I think it should be:
```suggestion
        if not (len(N_arr) == len(D_arr) == len(L_arr)):
```

Otherwise you could have a situation like this:
```
a = 1
b = 2
c = 2
print(a != b != c)  # False
```

src/olmo_core/model_ladder/analysis/scaling_laws.py
```diff
@@ -0,0 +1,432 @@
...
+            D=D,
+            L=L,
+            weights=weights_clean,
+            overestimate_penalty=overestimate_penalty,
+        )
+
+        results: list[Tuple[float, ChinchillaParams]] = []
+        if parallel:
+            n_workers = os.cpu_count() or 1
+            ctx = multiprocessing.get_context("fork")
```
epwalsh: Does this work on Mac?

src/test/model_ladder/analysis/plotting_test.py
```diff
@@ -0,0 +1,104 @@
+"""Basic smoke tests for plotting functions."""
+
+import numpy as np
+import pytest
+
+from olmo_core.model_ladder.analysis.eval import RolloutSplit, ScalingLawRollout
+from olmo_core.model_ladder.analysis.scaling_laws import ChinchillaParams
+
+# Check if plotly is available
```
tyler-romero:
```suggestion
```

src/test/model_ladder/analysis/scaling_laws_test.py
```diff
@@ -0,0 +1,166 @@
...
+    D = np.array([1e9])
+    # Use tiny values for A and B
+    loss = chinchilla_parametric_scaling_law(N, D, 1.0, 1e-20, 0.5, 1e-20, 0.3)
+    assert np.all(np.isfinite(loss))
+    # Loss should be approximately E when A and B are tiny
+    assert abs(loss[0] - 1.0) < 1e-10
+
+
+def test_chinchilla_parametric_fit_synthetic_data():
+    """Test fitting on synthetic data."""
```
tyler-romero:
```suggestion
```

src/test/model_ladder/analysis/evals_test.py
```diff
@@ -0,0 +1,281 @@
...
+    assert len(eval_by_cutoff.split_evaluations) == 2
+
+    # Constant weight (should be same as uniform)
+    eval_const = evaluate_rollout(rollout, split_weights_fn=lambda s: 42.0)
+    eval_uniform = evaluate_rollout(rollout)
+    assert np.isclose(eval_const.overall_mean_ppl_error, eval_uniform.overall_mean_ppl_error)
+
+
+def test_scaling_law_rollout_fit():
+    """Test ScalingLawRollout.fit creates rollout splits."""
```
tyler-romero:
```suggestion
```

src/olmo_core/model_ladder/analysis/scaling_laws.py
tyler-romero: It was working but you're right its sketchy. I got rid of it to just use the platform-specific defaults.

tyler-romero: The important thing is that it is a process pool instead of a threadpool to get around the GIL.

PR #558, opened 2026-01-21
Add model merging callback with integration test

baileykuehl:
Added a model merging callback that averages model weights over the last N steps of training and saves the result as a separate `stepXXXX-merged` directory and evaluates with metrics saved as `eval/merged/lm/`
                                                                                                                                                                                                       
Key features:   
  - Disabled by default (enabled=False) because degree of performance impact at scale is currently unknown (measuring this impact is out of scope on this PR, but I will follow-up with experiments to test / optimize this)                                                                                                  
  - Does NOT save accumulator state (no state_dict/load_state_dict). Instead, will load from most recent checkpoint prior to the merge and blocks ephemeral checkpoints during the merge window to prevent shortened merge windows
      -  Adds a flag to the trainer to block ephemeral checkpoints during merge windows
  - Integration test added that trains a small model for 5 steps (averages the last 3 steps) and verifies that the merged model is correctly saved and loaded (including in distributed / sharded cases)


Example of recent runs with this callback:
  - 60M (merge 500 steps): https://beaker.org/orgs/ai2/workspaces/oe-t-ladder/work/01KHMEFS0ZF2Q15RYE3N101M7H
  - 1B (merge 500 steps): https://beaker.org/orgs/ai2/workspaces/oe-t-ladder/work/01KHMFRGEMHSJ3Y7V7YH0TM328   

Passing integration test:
- https://beaker.org/orgs/ai2/workspaces/OLMo-core/work/01KHNBKD3A6HZ9E0785VNBP9B4

### DIFF
--- a/src/olmo_core/train/callbacks/model_merger.py
+++ b/src/olmo_core/train/callbacks/model_merger.py
@@ -276,9 +276,7 @@ class ModelMergeCallback(Callback):
         for name, param in params_dict.items():
             if name in averaged_state:
                 local_param = get_local_tensor(param.data)
-                local_param.copy_(
-                    averaged_state[name].to(local_param.device, local_param.dtype)
-                )
+                local_param.copy_(averaged_state[name].to(local_param.device, local_param.dtype))
 
         barrier()
 


.github/workflows/main.yml
```diff
@@ -76,6 +76,11 @@ jobs:
               pytest -v --color=yes --durations=3 \
                 src/test/scripts/
 
+          - name: Test small model
```
epwalsh: Since we might add more tests here in the future let's just name this "Integration tests"
```suggestion
          - name: Integration tests
```

.github/workflows/main.yml
```diff
@@ -193,6 +198,13 @@ jobs:
             run: |
               pytest -v --color=yes --durations=3 -m gpu \
                 src/test/optim*
+
+          - name: Test small model FSDP (GPU)
```
epwalsh: Same thing here
```suggestion
          - name: Integration tests (GPU)
```

src/test/train/callbacks/model_merger_test.py
epwalsh, line 1: Nit: this should go in `src/test/train/callbacks/model_merger_test.py`

src/test/train/callbacks/model_merger_test.py
```diff
@@ -0,0 +1,96 @@
...
+
+import logging
+
+import pytest
+
+from olmo_core.exceptions import OLMoConfigurationError
+from olmo_core.train.callbacks import ModelMergeCallback
+
+
+class TestModelMergeCallbackValidation:
```
epwalsh: Nit: we typically just use test functions, not classes. I don't see any benefit to using a class here since `self` is never accessed.

.gitignore
```diff
@@ -60,3 +60,6 @@ site/
 /scratch/
 core
 /dataset-cache/
+venv/
+*.md
+test_fingerprint*.py
```
epwalsh: Was this intentional?
```suggestion
```

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `ModelMergeCallback` for averaging model weights over the last N steps before a merge step.
+- Added integration test for training a small model (`src/integration_tests/test_train_small_model.py`).
```
epwalsh: This can be excluded from the changelog since it's not part of the API, i.e. not something a user would need to know about.
```suggestion
```

src/olmo_core/train/callbacks/model_merger.py
```diff
@@ -0,0 +1,555 @@
...
+            )
+            # Ensure all ranks have restored before continuing training
+            barrier()
+
+    def state_dict(self) -> Dict[str, Any]:
+        """
+        Save callback state for checkpointing.
+
+        NOTE: Each rank saves its own shard of the accumulator.
+        This increases total checkpoint size by ~1x model sizeca.
```
epwalsh: "sizeca"? Typo?

src/olmo_core/train/callbacks/model_merger.py
```diff
@@ -0,0 +1,555 @@
...
+        This might cause issues with scalability. Could consider:
+            - Not checkpointing accumulator (would lose ability to resume mid-window)
+            - Making accumulator checkpointing configurable
+        """
+        state = {
+            "n_accumulated": self._n_accumulated,
```
epwalsh: The way the trainer saves this state is not ideal for model state, especially when it's distributed/sharded. So we'll probably need to rethink this.

Here's one alternative:

Say we're merging the last 100 steps. We enforce saving a full training checkpoint right when start to accumulate, i.e. 100 steps before we want to evaluate the merged model, and we don't checkpoint again until after the merged model is evaluated. That way we wouldn't need to save the state of this call back because it would always be recomputed.

src/olmo_core/train/callbacks/model_merger.py
```diff
@@ -0,0 +1,555 @@
...
+
+        NOTE(large-scale): For larger models, calling state_dict() and moving to CPU at every
+        step in the merge window may cause significant throughput overhead.
+        Might want to consider sparse sampling (accumulate every Nth step instead of every step)
+        or doing this one time (moving to CPU only at the end)
+        """
+        sd_options = dist_cp_sd.StateDictOptions(full_state_dict=False, cpu_offload=True)
+        model_state = dist_cp_sd.get_model_state_dict(
+            self.trainer.train_module.model, options=sd_options
+        )
```
epwalsh: Using `dist_cp*` functions comes with additional overhead that I think we can safely avoid by just collecting model parameters directly, e.g. something like `model_state = {k: get_local_tensor(p.data.detach()) for k, p in model.named_parameters()}`.

There's still the overhead of copying those params to CPU... One option is to keep them on GPU if we can afford it. But for larger models we're typically memory-starved, so that's doesn't scale. Alternatively we could still copy to CPU but try to hide the overhead of the copy by doing it asynchronously.

For example,
1. Trigger an async copy as early as possible, like in `pre_load_batch()`. You can do an async copy with `olmo_core.utils.move_to_device(tensor, torch.device("cpu"), non_blocking=True)`. Now you can't do anything else with the CPU-tensor until you've triggered a host-device sync, so for now you wait.
2. Right before the optimizer step you force a host-device sync (call `torch.cuda.synchronize()`) to ensure the copy to CPU has finished before params are modified.
3. Then accumulate those weights.

This means the accumulated weights are a step behind (since they're updated pre-optim step), but I think we can adjust the bookkeeping accordingly.

.gitignore
baileykuehl: Ah whoops, no it wasn't. Thank you!

src/olmo_core/train/callbacks/model_merger.py
baileykuehl: Yep good catch. Thanks!

src/test/train/callbacks/model_merger_test.py
baileykuehl: Makes sense! Will fix this, thanks!

src/olmo_core/train/callbacks/model_merger.py
baileykuehl: That makes a lot of sense! So the trade-off of this would just be that if we get interrupted during the merge window, we would lose up to `merge_last_n_steps` - 1 steps of work, at most? Which seems reasonable while we're using fairly low values

baileykuehl: cc @dirkgr - I think you had some thoughts on this approach

dirkgr: Yeah, I worry about the scenario where we want to save ephemeral checkpoints all the time, say, every 200 steps, because the cluster is unstable, but we also want to merge more than 200 steps.

We don't really know how many steps we'll want to merge yet. I'm assuming the logic of saving checkpoints from this callback is just really bad? If so, we should do some experiments to see if we ever want to merge more than 200 steps. If that's a ridiculous number, maybe this is not worth fixing.

src/olmo_core/train/callbacks/model_merger.py
dirkgr: How about doing it in a different CUDA stream?

src/olmo_core/train/callbacks/model_merger.py
baileykuehl: Could an alternative to checkpointing at the start of the merge window be to allow ephemeral checkpoints during merge windows and just accept a shortened merge (i.e. resuming from the last ephemeral checkpoint in the merge window and start accumulating from there) on resume if we get interrupted? That would still avoid saving the accumulated state in the checkpoint like @epwalsh raised a concern about.

For experiments, what I've seen so far is that merging ~500 steps does almost as well as post anneal (though this was on a small model), so I don't think 200 is a crazy number, though I can run more experiments (I could try a 1B and/or 7B?).

epwalsh: Here's an alternative: just block ephemeral checkpoints during the accumulation window. So we could still have ephemeral checkpoints every 200 steps the rest of the time, just not during the merge window. That seems reasonable to me.

src/olmo_core/train/callbacks/model_merger.py
epwalsh: We'd have to be careful. You're copying data in one stream that will eventually (in the optimizer step) be modified in-place on a different stream. We'd have to call `.wait_stream()` before doing the optim step I think.

epwalsh: https://docs.pytorch.org/docs/stable/notes/cuda.html#cuda-streams

baileykuehl: Sounds good- I'll take a look into both options. I think to keep this moving, I'm going to keep this out of scope for the current PR and after we get the current version merged, I'll follow-up with a separate PR to address this.

src/olmo_core/train/callbacks/model_merger.py
baileykuehl: @dirkgr does that work for you?

dirkgr: If merging 200 steps does fine, then I'm OK with it. If merging 200 steps is catastrophically worse than 500 steps, we'll need a better option.

We'll have to look out for the off-by-one error. Merging every 200 steps and ephemeral checkpointing every 200 steps should work together.

For us to use this regularly, it has to reach a usability threshold. It has to be an MVP.

src/olmo_core/train/callbacks/evaluator_callback.py
```diff
@@ -113,7 +113,13 @@ def post_step(self):
         ):
             self._perform_eval()
 
-    def _perform_eval(self):
+    def _perform_eval(self, prefix: str = "eval"):
```
epwalsh: Since we're now calling this method from another callback, I think we should make it public.

src/olmo_core/train/trainer.py
```diff
@@ -242,6 +242,13 @@ class Trainer:
     training throughput.
     """
 
+    block_ephemeral_checkpoints: bool = False
```
epwalsh: One problem with this approach is that we could have multiple callbacks mutating this one flag, and that could lead to confusing bugs. Alternatively we could do something like this:


```python
class Trainer:
    ...

    _blocking_ephemeral_checkpoints: set[str] = dataclasses.field(repr=False, init=False, default_factory=set)
    """Callbacks that are blocking ephemeral checkpoints."""

    @property
    def block_ephemeral_checkpoints(self) -> bool:
        return len(self._blocking_ephemeral_checkpoints) > 0

    def get_callback_name(self, callback: Callback) -> str:
        for name, cb in self.callbacks.items():
            if cb is callback:
                return name
        raise ValueError("callback not registered with trainer!")
```

```python
class Callback:
    def block_ephemeral_checkpoints(self):
        name = self.trainer.get_callback_name(self)
        self.trainer._blocking_ephemeral_checkpoints.add(name)

    def unblock_ephemeral_checkpoints(self):
        name = self.trainer.get_callback_name(self)
        if name in self.trainer._blocking_ephemeral_checkpoints:
            self.trainer._blocking_ephemeral_checkpoints.pop(name)
```

src/olmo_core/train/callbacks/evaluator_callback.py
baileykuehl: Ah makes sense, fixed!

src/olmo_core/train/trainer.py
baileykuehl: Good call, thanks! Fixed this

src/olmo_core/train/trainer.py
```diff
@@ -242,6 +243,19 @@ class Trainer:
...
+
+    @property
+    def block_ephemeral_checkpoints(self) -> bool:
+        return len(self._blocking_ephemeral_checkpoints) > 0
+
+    def get_callback_name(self, callback: Callback) -> str:
+        for name, cb in self.callbacks.items():
+            if cb is callback:
+                return name
+        raise ValueError("callback not registered with trainer!")
```
epwalsh: Can you move these methods below `__post_init__`?

src/olmo_core/train/callbacks/callback.py
```diff
@@ -56,6 +56,15 @@ def load_state_dict(self, state_dict: Dict[str, Any]):
         """
         del state_dict
 
+    def block_ephemeral_checkpoints(self):
+        name = self.trainer.get_callback_name(self)
+        self.trainer._blocking_ephemeral_checkpoints.add(name)
+
+    def unblock_ephemeral_checkpoints(self):
```
epwalsh, line 65: Worth adding docstrings to these to explain what they do.

baileykuehl: Done! Can adjust the amount of detail though if needed!

src/olmo_core/train/trainer.py
baileykuehl: Done!

src/olmo_core/train/trainer.py
```diff
@@ -242,6 +243,9 @@ class Trainer:
     training throughput.
     """
 
+    _blocking_ephemeral_checkpoints: Set[str] = field(repr=False, init=False, default_factory=set)
+    """Callbacks that are blocking ephemeral checkpoints."""
+
```
epwalsh: nit: I like to keep private fields below under the `# Internal bookkeeping` comment

src/olmo_core/train/trainer.py
```diff
@@ -385,6 +389,16 @@ def __post_init__(self):
...
 
+    @property
+    def block_ephemeral_checkpoints(self) -> bool:
+        return len(self._blocking_ephemeral_checkpoints) > 0
+
+    def get_callback_name(self, callback: Callback) -> str:
+        for name, cb in self.callbacks.items():
+            if cb is callback:
+                return name
+        raise ValueError("callback not registered with trainer!")
```
epwalsh: nit: I like to put methods below all of the properties

src/olmo_core/train/callbacks/model_merger.py
```diff
@@ -0,0 +1,388 @@
...
+        # and only save rank 0's data). We keep the weights loaded for evaluation
+        # to avoid swapping twice.
+        model = self.trainer.train_module.model
+        params_dict = dict(model.named_parameters())
+
+        original_state = {
+            k: get_local_tensor(p.data.detach()).to("cpu").clone() for k, p in params_dict.items()
+        }
+
+        with torch.no_grad():
```
epwalsh: You could also just decorate this whole method with `@torch.no_grad`. That's probably safer, and then you don't need to use this context manager at all within the method.

src/olmo_core/train/trainer.py
baileykuehl: Make sense, fixed!

src/olmo_core/train/trainer.py
baileykuehl: Gotcha, thanks! Moved below all properties

src/olmo_core/train/callbacks/model_merger.py
baileykuehl: Done, that's a lot cleaner. Thank you!

PR #567, opened 2026-01-27
Replace OmegaConf with dataclass-extensions

epwalsh:
[dataclass-extensions](https://github.com/epwalsh/dataclass-extensions) is a little library I wrote to improve the experience of using dataclasses as configs (encode/decode to/from JSON/YAML), and bring back the old `Registrable` functionality from the [AllenNLP library](https://github.com/allenai/allennlp) for config-based polymorphism in a way that doesn't give me nightmares trying to maintain it.

This PR ignores utilizing the `Registrable` functionality and just focuses for now on replacing OmegaConf with `dataclasses-extentions` in a way that's completely backwards compatible.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ dependencies = [
     "cached-path>=1.7.2",
     "requests",
     "packaging",
-    "dataclass-extensions>=0.2.5",
+    "dataclass-extensions>=0.2.6",
     "pyyaml",
     "rich",
     "safetensors",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -61,7 +62,7 @@ comet = ["comet_ml"]
 dion = ["dion @ git+https://github.com/microsoft/dion.git@7452a5823cf9655b93c3f1d8020b4ebb2535239b ; sys_platform == 'linux'"]
 eval = ["ai2-olmo-eval==0.8.5"]
 torchao = ["torchao==0.15.0"]
-transformers = ["transformers"]
+transformers = ["transformers<5.0"]
```
epwalsh, line 65: unrelated but the 5.0 release breaks some things

PR #569, opened 2026-01-27
Integrate QuACK kernels, minor improvements to Docker build

epwalsh:
- Adds the option to install QuACK in our Docker images.
- Adds a new Beaker image with torch 2.9.1, CUDA 12.9.1, QuACK, and all other optional dependencies: `petew/olmo-core-tch291cu129-2026-01-24`.
- Adds a CuTe-based RMSNorm from the [QuACK](https://github.com/Dao-AILab/quack) library. I didn't see any improvement at the 7B scale, but I'm curious to try with some larger models soon. And there's a bunch of other kernels in QuACK that could be useful.

### DIFF
--- a/src/test/nn/layer_norm_test.py
+++ b/src/test/nn/layer_norm_test.py
@@ -35,6 +35,7 @@ def test_fused_rms_norm(bias, dtype):
 def test_cute_rms_norm():
     dim = 64
     norm = CuTeRMSNorm(size=dim, init_device="cuda")
+    norm.compile()
     ref_norm = RMSNorm(size=dim, init_device="cuda")
 
     x = torch.randn(4, dim, requires_grad=True, device="cuda", dtype=torch.bfloat16)


src/olmo_core/internal/experiment.py
```diff
@@ -415,7 +415,6 @@ def build_config(
 
 
 def launch(config: ExperimentConfig):
-    log.info(config)
```
epwalsh, line 418: Resulting in printing the config twice.

src/olmo_core/internal/experiment.py
```diff
@@ -429,7 +428,6 @@ def launch(config: ExperimentConfig):
     config.launch.launch(
         follow=True,
         slack_notifications=slack_enabled,
-        launch_timeout=5 * 60,
```
epwalsh, line 432: This gets taken from the config so we can override it at runtime.

PR #570, opened 2026-01-27
SequenceMixer as base for both attention and recurrent layers

tyler-romero:
Setting up for supporting recurrent layers such as GatedDeltaNet

### DIFF
--- a/src/olmo_core/config.py
+++ b/src/olmo_core/config.py
@@ -262,7 +262,7 @@ class Config:
                 d = {(int(k) if isinstance(k, str) and k.isdigit() else k): v for k, v in d.items()}
 
                 new_dict = {
-                    str(k): decode_data(v, f"{prefix}.{k}" if prefix else str(k))
+                    k: decode_data(v, f"{prefix}.{k}" if prefix else str(k))
                     for k, v in d.items()
                     if k != cls.CLASS_NAME_FIELD
                 }
@@ -274,7 +274,7 @@ class Config:
                         new_dict = {k: v for k, v in new_dict.items() if k not in ignore_fields}
 
                     try:
-                        return decode(cls_o, new_dict)
+                        return decode(cls_o, new_dict)  # type: ignore[arg-type]
                     except Exception as e:
                         if prefix:
                             msg = f"Failed to construct '{prefix}' in config: {e}"


src/olmo_core/nn/transformer/block.py
```diff
@@ -99,7 +99,7 @@ class TransformerBlock(TransformerBlockBase):
 
     :param d_model: The model dimensionality.
     :param block_idx: The index/position of the block within the model. Ranges from 0 to ``n_layers - 1``.
-    :param attention: The attention module config.
+    :param sequence_mixer: The sequence mixer module config (e.g. attention, recurrent, convolution, etc.).
```
tyler-romero, line 102: How do we feel about this sort of naming / renaming?

Modifying might break backwards compat with old configs. Likewise renaming `self.attention` below would break backwards compat with old checkpoints.

Could just keep calling everything attention?

src/olmo_core/nn/transformer/config.py
```diff
@@ -150,9 +151,9 @@ class TransformerBlockConfig(ModuleConfig):
     A configuration class for easily building transformer blocks.
     """
 
-    attention: AttentionConfig
+    sequence_mixer: SequenceMixerConfig
```
epwalsh: Renaming this would break backwards compatibility for configs, meaning we couldn't load a config from an old checkpoint without doing some preprocessing. Alternatively we could keep a separate `attention: Optional[AttentionConfig] = None` here and handle setting it to `sequence_mixer` in `__post_init__`. It could also just be an `InitVar` "field" instead of a regular field.

src/olmo_core/nn/transformer/model.py
```diff
@@ -112,7 +112,6 @@ def __init__(
         self.d_model = d_model
         self.vocab_size = vocab_size
         self.n_layers = n_layers
-        self.n_attn_heads = block.attention.n_heads
```
epwalsh, line 115: I guess this wasn't used anywhere?

tyler-romero: Yeah, unused

src/olmo_core/nn/transformer/block.py
tyler-romero: discussed w/ pete offline

src/olmo_core/nn/transformer/config.py
tyler-romero: Took this suggestion and also added  an "UNSET" sentinel so that we don't have to treat `sequence_mixer` as None anywhere.

PR #572, opened 2026-01-27
Add GatedDeltaNet

tyler-romero:
Add a streamlined `GatedDeltaNet` layer with support for ulysses-style CP.

Reference impl: https://github.com/fla-org/flash-linear-attention/blob/3cf180339b8a1cbad823f553541cd531d18670ea/fla/layers/gated_deltanet.py#L34


epwalsh (approved):
I didn't review the math in detail but the API looks good and tests seem thorough ✅

PR #573, opened 2026-01-27
Support residual epilogue in cute rmsnorm

tyler-romero:
Since torch compile cant fuse into custom kernels, manual work is required to wire up epilogues to avoid incurring additional memory transfers. CuTE rmsnorm has support for residual connections, so we can try to use that to see what the speedup looks like.

### DIFF
--- a/src/olmo_core/nn/layer_norm.py
+++ b/src/olmo_core/nn/layer_norm.py
@@ -294,7 +294,6 @@ class CuTeRMSNorm(RMSNorm):
         )
         return result.to(x.dtype)
 
-,
 class FusedRMSNorm(RMSNorm):
     """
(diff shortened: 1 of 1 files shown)


src/olmo_core/nn/layer_norm.py
```diff
@@ -268,15 +277,24 @@ def __init__(
...
             weight=None if self.weight is None else self.weight.type_as(x),
             bias=None if self.bias is None else self.bias.type_as(x),
+            residual=residual,
             eps=self.eps,
-        ).to(x.dtype)
-
+        )
+        return result.to(x.dtype)
 
+,
```
epwalsh:
```suggestion
```

PR #578, opened 2026-01-29
Mount oe-adapt-default as well when mounting oe-training-default

jacob-morrison:
As we're using olmo-core more for post-training, and because many of our post-training datasets and models are in oe-adapt-default, let's mount it as well when mounting weka, so we don't have to save all of our data in oe-training-default instead.


epwalsh (approved):
LGTM

baileykuehl (approved):

PR #582, opened 2026-01-30
Propagate gantry launch options

tyler-romero:
Bug: follow, launch_timeout, step_timeout, and step_soft_timeout would not be applied if specified via `BeakerLaunchConfig` (or, equivalently, via command line override)

### DIFF
--- a/src/olmo_core/launch/beaker.py
+++ b/src/olmo_core/launch/beaker.py
@@ -480,7 +480,7 @@ class BeakerLaunchConfig(Config):
         Arguments are the same as :meth:`launch()`.
         """
         with get_beaker_client(workspace=self.workspace) as beaker:
-            recipe, recipe_launch_kwargs = self._build_recipe(
+            recipe, _ = self._build_recipe(
                 beaker,
                 follow=follow,
                 slack_notifications=slack_notifications,
@@ -488,9 +488,6 @@ class BeakerLaunchConfig(Config):
                 step_timeout=step_timeout,
                 step_soft_timeout=step_soft_timeout,
                 torchrun=torchrun,
-            )
-            log.info(
-                f"Experiment would be launched with the following options: {recipe_launch_kwargs}"
             )
             recipe.dry_run(client=beaker)
 


src/olmo_core/launch/beaker.py
tyler-romero, line 469: note: this is never actually used in our codebase

epwalsh: We use this in the current ladder: https://github.com/allenai/OLMo-core/blob/6b2cd7cf5dad76bc9bfad7a170bd2e6870e95e5f/src/olmo_core/internal/ladder.py#L490

tyler-romero: Ah.. I grepped for launch.dry_run

src/olmo_core/launch/beaker.py
```diff
@@ -374,20 +374,20 @@ class BeakerLaunchConfig(Config):
 
     launch_timeout: int | None = None
     """
-    A timeout in seconds to wait for the job to start after submitting it.
+    A timeout in seconds to wait for the job to start after submission
```
epwalsh:
```suggestion
    A timeout in seconds to wait for the job to start after submission.
```

src/olmo_core/launch/beaker.py
```diff
@@ -374,20 +374,20 @@ class BeakerLaunchConfig(Config):
...
     If the job doesn't start in time a timeout error will be raised.
     """
 
     step_timeout: int | None = None
     """
-    A timeout in seconds to wait for the first training step when ``follow=True``.
+    A timeout in seconds to wait for new logs when ``follow=True``.
```
epwalsh: It really is training steps though, not just any logs.
```suggestion
    A timeout in seconds to wait for new training steps in the logs when ``follow=True``.
```

src/olmo_core/launch/beaker.py
```diff
@@ -489,6 +489,9 @@ def dry_run(
                 step_soft_timeout=step_soft_timeout,
                 torchrun=torchrun,
             )
+            log.info(
+                f"Experiment would be launched with the following options: {recipe_launch_kwargs}"
+            )
```
epwalsh: Was this just for debugging or do you want to keep this?

PR #585, opened 2026-01-30
Make SLACK_WEBHOOK_URL not actually required

tyler-romero:
SLACK_WEBHOOK_URL is marked as not required, but we try to check its existence in a way that will throw an error if it doesnt exist.

### DIFF
--- a/src/olmo_core/launch/beaker.py
+++ b/src/olmo_core/launch/beaker.py
@@ -587,7 +587,9 @@ class BeakerLaunchConfig(Config):
             else:
                 # Pull from secret if available.
                 for env_secret in self.env_secrets:
-                    if env_secret.name == SLACK_WEBHOOK_URL_ENV_VAR and self._secret_exists(env_secret):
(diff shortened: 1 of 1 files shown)


src/olmo_core/launch/beaker.py
```diff
@@ -586,11 +587,7 @@ def _build_recipe(
                 slack_webhook_url = os.environ[SLACK_WEBHOOK_URL_ENV_VAR]
             else:
                 # Pull from secret if available.
-                for env_secret in self.env_secrets:
-                    if env_secret.name == SLACK_WEBHOOK_URL_ENV_VAR:
-                        secret = beaker.secret.get(env_secret.secret)
-                        slack_webhook_url = beaker.secret.read(secret)
-                        break
+                slack_webhook_url = dict(env_secrets).get(SLACK_WEBHOOK_URL_ENV_VAR)
```
epwalsh: This would set it to the name of the secret, not the value of the secret. And I don't think that's what we want.

PR #586, opened 2026-01-30
Optional disk-backed caching for remote IO ops, data loading config improvements

epwalsh:
- Adds a way to cache the result of remote IO operations, like globbing and stat-ing file sizes, in a persistent local filesytem cache by setting the env var `OLMO_CORE_FS_CACHE_DIR`. This can speed up restart time when that directory is shared between runs.
- Makes `DataLoaderConfig` a registrable base class.
- Also adds the option to use the composable data API in internal train scripts.

Example with caching:

```python
import logging
import os
import time

from olmo_core.data import DataMix, TokenizerConfig
from olmo_core.data.composable import *
from olmo_core.fs_cache import CACHE_DIR_ENV_VAR
from olmo_core.utils import prepare_cli_environment

log = logging.getLogger()
tokenizer = TokenizerConfig.dolma2()
sequence_length = 8192

config = NumpyDocumentSourceMixConfig(
    tokenizer=tokenizer, mix=DataMix.OLMo_mix_0925, mix_base_dir="gs://ai2-llm"
)


def main():
    prepare_cli_environment()
    os.environ[CACHE_DIR_ENV_VAR] = "/tmp/OLMo-core/fs_cache"  # comment this out to disable caching

    start_time = time.monotonic()
    log.info("Building sources...")
    config.build("/tmp/OLMo-core/dataset_cache")
    log.info(f"Done. Finished in {time.monotonic() - start_time:.2f} seconds.")


if __name__ == "__main__":
    main()
```

Result with cache miss:

```
[2026-01-30 15:33:29] INFO     [root, Pete-Walsh-MacBook-Pro, rank=0] Building sources...
[2026-01-30 15:33:29] INFO     [olmo_core.data.composable.numpy_document_source:768, Pete-Walsh-MacBook-Pro, rank=0] Assuming dtype 'uint32' based on vocab size
[2026-01-30 15:33:37] INFO     [root, Pete-Walsh-MacBook-Pro, rank=0] Done. Finished in 8.09 seconds.
```

And with a cache hit after running a 2nd time:

```
[2026-01-30 15:33:40] INFO     [root, Pete-Walsh-MacBook-Pro, rank=0] Building sources...
[2026-01-30 15:33:40] INFO     [olmo_core.data.composable.numpy_document_source:768, Pete-Walsh-MacBook-Pro, rank=0] Assuming dtype 'uint32' based on vocab size
[2026-01-30 15:33:40] INFO     [root, Pete-Walsh-MacBook-Pro, rank=0] Done. Finished in 0.18 seconds.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `BeakerCallback` will include throughput metrics in the workload description.
 - Added `olmo_core.io.deterministic_glob_directory` function.
 - Added the option to cache the results of certain IO operations on remote files, like `get_file_size()` and `deterministic_glob_directory()` by setting the env var `OLMO_CORE_FS_CACHE_DIR` to a local directory.
+- Added `eval_on_finish` option to `EvaluatorCallback`.
 
 ### Fixed
 
@@ -55,10 +56,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed a bug where final metrics might not get logged.
 - Fix failing test_build_world_mesh_cpu for pytorch 2.10.
 - Fix failing convert_checkpoint_to_hf_test due by reducing total disk space required.
-
+- Ensure all metrics have been logged and bookkeeping ops complete before writing a checkpoint.
 
 ### Changed
 
+- Updated SFT documentation with alternative tokenization approach and tips for new base models.
 - Renamed `olmo_core.distributed.utils.scatter_object()` to `broadcast_object()` for correctness.
 - Updated stable torch version to 2.9.1, updated versions of underlying libraries in Beaker Images.
 - `olmo_core.io.join_path()` now accepts an arbitrary number of components to join.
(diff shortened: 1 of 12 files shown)


pyproject.toml
```diff
@@ -16,6 +16,7 @@ dependencies = [
     "numpy",
     "torch>=2.6.0",
     "cached-path>=1.7.2",
+    "filelock>=3.4,<4.0",
```
epwalsh, line 19: This was already a (indirect) dependency through `cached-path`.

src/olmo_core/fs_cache.py
```diff
@@ -0,0 +1,88 @@
...
+                        # Cache hit. Load and return the cached result.
+                        with cache_path.open("rb") as f:
+                            return pickle.load(f)
+                    else:
+                        # Cache miss. Call the user function and cache the result.
+                        log.debug(
+                            "Computing result for %s() and storying to filesystem cache...",
+                            user_function.__qualname__,
+                        )
+                        result = user_function(*args, **kwargs)
```
dirkgr, line 60: When the result of the function is that it throws an exception, we don't cache that. I think that's fine. We could have made the decision either way. But maybe worth pointing out.

pyproject.toml
dirkgr: As we remember, this library loses the lock when you are running Google Drive on the same machine.

PR #588, opened 2026-02-02
Checkpointing improvements

epwalsh:
- Use `os.fdatasync` to ensure all data in a temp file is written to disk before replacing the temp file with the final file. According to https://stackoverflow.com/a/67905617, flushing/closing may not be enough.
- Replace usage of deprecated `tmpfile.mktemp` with `tmpfile.NamedTemporaryFile`.

### DIFF
--- a/src/olmo_core/distributed/checkpoint/__init__.py
+++ b/src/olmo_core/distributed/checkpoint/__init__.py
@@ -94,7 +94,8 @@ def save_state_dict(
     :param process_group: The process group to use for distributed collectives.
     :param save_overwrite: Overwrite existing files.
     :param thread_count: Set this to override the number of threads used while writing data.
-    :param process_count: Set this to use a process pool instead of a thread pool when possible.
+    :param process_count: Set this to use a process pool instead of a thread pool when possible
+        (currently not compatible with ``throttle_uploads``).
     :param throttle_uploads: If this is set to ``True`` and ``dir`` is a URL then only one
         rank from each node will upload data at a time.
     """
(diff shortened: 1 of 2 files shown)


src/olmo_core/distributed/checkpoint/filesystem.py
```diff
@@ -99,38 +100,46 @@ def _write_items(
 ) -> List[WriteResult]:
     results: List[WriteResult] = []
 
-    tmp_path = Path(
-        tempfile.mktemp(suffix=".distcp", dir=None if is_url(path) else Path(path).parent)
```
epwalsh, line 103: `mktemp` is deprecated because it's unsafe.

src/olmo_core/distributed/checkpoint/filesystem.py
```diff
@@ -211,37 +245,37 @@ def gen_file_name() -> str:
...
             )
         else:
             buckets = _split_by_size_and_type(self.thread_count, plan.items)
+            paths = [gen_file_name() for _ in buckets]
             results = []
-            with ThreadPoolExecutor(max_workers=self.thread_count) as executor:
+            with ExitStack() as stack:
+                executor: Executor
+                if self.process_count is not None and self.process_count > 1:
+                    executor = stack.enter_context(
```
tyler-romero, line 266: nice

src/olmo_core/distributed/checkpoint/__init__.py
```diff
@@ -93,6 +94,7 @@ def save_state_dict(
     :param process_group: The process group to use for distributed collectives.
     :param save_overwrite: Overwrite existing files.
     :param thread_count: Set this to override the number of threads used while writing data.
+    :param process_count: Set this to use a process pool instead of a thread pool when possible.
```
tyler-romero: What does "when possible" imply here? It looks like its on the user to choose either threads or processes.

epwalsh: Currently not compatible with `throttle_uploads`. I added that info to the docstring here.

epwalsh:
@tyler-romero I ran into some multiprocessing issues trying to use a process pool to write checkpoints during an actual training job. I need more time to investigate, so I just removed that option for now so we can get these other changes in.

tyler-romero (approved):

Issue #503, opened 2025-12-16
[Feature Request] Support variable-length sequences in RoPE natively


yezhengmao1:
Currently, the RoPE (Rotary Positional Embeddings) implementation does not support variable-length (varlen) sequences natively.

If users train with variable-length data and do not explicitly pass the pos_sin and pos_cos arguments, it can lead to potentially incorrect calculation results (as the positions are not correctly aligned with the actual sequence lengths).
At present, users must perform an extra step to generate the pos_sin and pos_cos matrices externally to ensure correctness.

I suggest adding support for variable-length data directly within the RoPE module. 
A specific implementation reference can be found in [Megatron-LM](https://github.com/NVIDIA/Megatron-LM/blob/ff45bd4255af1c8f6b96209bfcd3e966062ee41a/megatron/core/models/common/embeddings/rope_utils.py#L4)

If this feature sounds reasonable, I am willing to submit a PR to implement this.

tyler-romero:
Hi! Thanks for looking into this.

We have a step where we initialize the rope buffers at the time we initialize the model weights:
https://github.com/allenai/OLMo-core/blob/89cef3df1e8c0beff479919dd657aaa0a46e87e2/src/olmo_core/nn/transformer/model.py#L299

This sets the rope buffers to the maximum allowed seq len during training (I think this is what you mean when you say that the pos_sin and pos_cos matrices need to be generated externally). So all var len sequences would reuse this rope buffer and have the same relative pos encodings.

Does this address the issue you are referring to? If not can you give more detail?

yezhengmao1:
> Hi! Thanks for looking into this.
> 
> We have a step where we initialize the rope buffers at the time we initialize the model weights:
> 
> [OLMo-core/src/olmo_core/nn/transformer/model.py](https://github.com/allenai/OLMo-core/blob/89cef3df1e8c0beff479919dd657aaa0a46e87e2/src/olmo_core/nn/transformer/model.py#L299)
> 
> Line 299 in [89cef3d](/allenai/OLMo-core/commit/89cef3df1e8c0beff479919dd657aaa0a46e87e2)
> 
>  att.rope.warmup_cache(max_seq_len, device) 
> This sets the rope buffers to the maximum allowed seq len during training (I think this is what you mean when you say that the pos_sin and pos_cos matrices need to be generated externally). So all var len sequences would reuse this rope buffer and have the same relative pos encodings.
> 
> Does this address the issue you are referring to? If not can you give more detail?

[...]

PR #591, opened 2026-02-03
Fix RoPE positions to reset at document boundaries when using doc_lens

finbarrtimbers:
When packing multiple documents with `cu_doc_lens`, RoPE was using global positions instead of per-document positions. This caused the second document to get positions `[20, 21, ...]` instead of `[0, 1, ...]`, breaking the expected behavior where identical documents should produce identical outputs.

To fix, we add a `cu_doc_lens` parameter to `RotaryEmbedding.forward()` and `ComplexRotaryEmbedding.forward()` which computes per-document positions using `searchsorted` to find document boundaries. We have a [minimal repro](https://gist.github.com/finbarrtimbers/6cfd2cf0f0b0225604dd9f2929ef4b58) which we ran on Beaker.

[Before this fix](https://beaker.org/ex/01KGJMT7GZ4NM4J781BNDTNZ87), identical documents packed with `doc_lens` produced different outputs because RoPE applied global positions (see `minimal_doc_lens_repro.py`):

```
Sample logits at position 5:
  Packed doc0[5, :5]:  [0.365234375, 0.0888671875, -0.5078125, 0.54296875, 0.9140625]
  Packed doc1[5, :5]:  [0.427734375, 0.1171875, -0.4921875, 0.5234375, 0.8671875]
```

After our fix, identical documents packed with `doc_lens` now [produce identical outputs](https://beaker.org/ex/01KGJNK1PEAVNJVNJGPS2NJ3G7):

```
Sample logits at position 5:
  Packed doc0[5, :5]:  [0.365234375, 0.0888671875, -0.5078125, 0.54296875, 0.9140625]
  Packed doc1[5, :5]:  [0.365234375, 0.0888671875, -0.5078125, 0.54296875, 0.9140625]
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added documentation for verifying chat template settings before running evals after SFT.
 - Added `olmo_core.data.composable` module.
 - Added `PeriNormTransformerBlock`.
 - Added exponential learning rate scheduler to `olmo_core.optim.scheduler`.
@@ -39,6 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `olmo_core.io.deterministic_glob_directory` function.
 - Added the option to cache the results of certain IO operations on remote files, like `get_file_size()` and `deterministic_glob_directory()` by setting the env var `OLMO_CORE_FS_CACHE_DIR` to a local directory.
 - Added `eval_on_finish` option to `EvaluatorCallback`.
+- Added the option to use a process pool instead of a thread pool when writing checkpoints.
 
 ### Fixed
 
@@ -58,6 +60,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix failing test_build_world_mesh_cpu for pytorch 2.10.
 - Fix failing convert_checkpoint_to_hf_test due by reducing total disk space required.
 - Ensure all metrics have been logged and bookkeeping ops complete before writing a checkpoint.
+- Minor improvements to make checkpointing more robust.
 
 ### Changed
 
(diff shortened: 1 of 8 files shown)


tyler-romero:
Nice!

Related: https://github.com/allenai/OLMo-core/issues/503

src/olmo_core/nn/rope.py
```diff
@@ -523,7 +567,29 @@ def forward(
...
+                    cos_k = cos_k[:, None, :, :]
+                else:
+                    sin_q = sin_q[:, :, None, :]
+                    cos_q = cos_q[:, :, None, :]
+                    sin_k = sin_k[:, :, None, :]
+                    cos_k = cos_k[:, :, None, :]
+
+                q_ = self._apply_rotary_pos_emb(sin_q, cos_q, q_)
+                k_ = self._apply_rotary_pos_emb(sin_k, cos_k, k_)
+            elif head_first:
```
tyler-romero, line 592: nit: a bit unintunitve to have head_first in the same if-else as the cu_doc_lens check. In general I think there is some redundancy between the if-else statements here.

src/olmo_core/nn/rope.py
```diff
@@ -50,6 +50,46 @@ def compute_inv_freqs(theta: int, dim: int, device: torch.device) -> "torch.Tens
     return inv_freq
 
 
+def compute_local_positions(
```
tyler-romero, line 53: Could this be wrapped in `@torch.no_grad()`?

src/olmo_core/nn/rope.py
```diff
@@ -523,7 +567,29 @@ def forward(
                     f"have {pos_sin.size(-2)}."
                 )
 
-            if head_first:
+            if cu_doc_lens is not None:
+                k_local_pos, q_local_pos = compute_local_positions(
```
tyler-romero, line 571: What are the perf implications of this? If its expensive it would be nice to only compute it once per batch (as applicable, since blocks of different shapes might get different k and q local outputs).

dirkgr:
Rope is completely relative though? Is this a precision thing?

PR #592, opened 2026-02-04
GatedDeltaNet improvements

tyler-romero:
1. Add an assert to prevent the usage of GQA. If a user did use GQA, they would hit a confusing illegal memory access error. It also just doesnt make much sense for linear rnn models.
2. Correct n_kv_heads -> n_v_heads
3. Add better docs.


epwalsh (approved):
LGTM!

PR #602, opened 2026-02-11
Fix GDN initialization

tyler-romero:
Correctly init linear and conv layers in the GatedDeltaNet module. Refactors attention init to delegate it to the specific class.

### DIFF
--- a/src/olmo_core/nn/attention/recurrent.py
+++ b/src/olmo_core/nn/attention/recurrent.py
@@ -237,6 +237,7 @@ class GatedDeltaNet(SequenceMixer):
         self.k_conv1d.apply_cp(cp_mesh)
         self.v_conv1d.apply_cp(cp_mesh)
 
+    @torch.no_grad()
     def init_weights(
         self,
         *,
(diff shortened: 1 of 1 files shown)


src/olmo_core/internal/ladder.py
```diff
@@ -110,8 +110,8 @@ def add_general_args(parser: argparse.ArgumentParser):
         parser.add_argument(
             "--cluster",
             type=str,
-            choices=["ai2/augusta", "ai2/jupiter", "ai2/titan"],
-            default="ai2/augusta",
+            choices=["ai2/jupiter", "ai2/titan"],
```
tyler-romero, line 113: unrelated fix for olmo3-ladder tests

src/olmo_core/nn/attention/recurrent.py
```diff
@@ -244,6 +237,46 @@ def apply_cp(
         self.k_conv1d.apply_cp(cp_mesh)
         self.v_conv1d.apply_cp(cp_mesh)
 
+    def init_weights(
```
epwalsh, line 241: Could also just wrap the whole method with `no_grad` to be safe.
```suggestion
    @torch.no_grad()
    def init_weights(
```

epwalsh (approved):

dirkgr (commented):
How was it getting initialized before?

tyler-romero:
> How was it getting initialized before?

It would be the default init, which is a uniform sample from (-sqrt(k), sqrt(k)) where `k=1/fan_in`

PR #603, opened 2026-02-11
HybridBlockConfig for cleaner definitions of hybrid models

tyler-romero:


src/olmo_core/nn/transformer/config.py
```diff
@@ -301,6 +317,91 @@ def num_active_params(self, d_model: int) -> int:
...
+    blocks: Dict[str, TransformerBlockConfig] = field(default_factory=dict)
+    """
+    Named block configurations.
+    """
+
+    blocks_pattern: List[str] = field(default_factory=list)
+    """
+    Pattern of block type names to cycle through. Each entry must be a key in :data:`blocks`.
+    ``len(blocks_pattern)`` must divide evenly into ``n_layers``.
+    """
```
epwalsh, line 349: I feel like it would make more sense to do this configuration one level up, in the `TransformerConfig`.

src/olmo_core/nn/transformer/config.py
```diff
@@ -314,7 +415,7 @@ class TransformerConfig(ModelConfig):
     d_model: int
     vocab_size: int
     n_layers: int
-    block: TransformerBlockConfig
+    block: BlockConfig
```
epwalsh, line 418: `dataclass-extensions` can handle complex types, so we could do this:


```suggestion
    block: TransformerBlockConfig | dict[str, TransformerBlockConfig]
```

src/olmo_core/nn/transformer/config.py
epwalsh: That way a block config is still just a block config.

PR #607, opened 2026-02-14
Support block-pattern based initialization of hybrid transformers

tyler-romero:
Support a cleaner method of configuring hybrid models. Just define each of your distinct `TransformerBlockConfig`s and then a pattern for interleaving them.

## Summary

- `TransformerConfig.block` now accepts either a single `TransformerBlockConfig` or a `dict[str, TransformerBlockConfig]` of named blocks. When a dict is provided, a `block_pattern` list controls which block config is used at each layer (the pattern cycles over layers).
- The existing `block_overrides` mechanism continues to work for single-block configs. It is mutually exclusive with the dict-based `block_pattern` approach.
- `peri_norm` builder configs (e.g. Gemma-style local/global attention hybrids) now use named blocks + `block_pattern` instead of generating per-layer `block_overrides`.
- Adds `resolve_block_configs()` and `validate_block_resolution_config()` helpers, and a `resolved_block_configs` property on `TransformerConfig`.
- Adds type guards (`isinstance` checks) where code accesses `block` attributes directly.
- Adds a backward-compat test that loads a real OLMo3-7B checkpoint config from `olmo-checkpoints.org`.

## Test plan

- [x] `make checks` passes (isort, black, ruff, mypy)
- [x] New `test_load_olmo3_7b_config` test verifies old single-block configs deserialize and round-trip correctly
- [x] Existing model builder tests pass (`test_small_llama2_builder_config`, etc.)
- [x] Existing HF checkpoint conversion tests pass

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,160 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Overview
+
+OLMo-core is AI2's training library for the Open Language Model (OLMo) series. It provides modular components for transformer architectures, distributed training, data loading, and evaluation.
+
+## Commands
+
+```bash
+# Install (development)
+pip install -e '.[all]'
+
+# Run all tests (GPU tests auto-skip without GPU)
+pytest -v src/
+
+# Run a specific test file
+pytest -v src/test/path/to/test_file.py
+
+# Filter to specific tests by keywords
+pytest -v src/test/path/to/test_file.py -k 'keyword'
+
+# Auto-format code
+make style
+
+# Check formatting, lint, and types
+make checks          # all three at once
+make style-check     # isort + black
+make lint-check      # ruff
+make type-check      # mypy
+```
+
+## Code Style
+
+- Line length: 100
+- Formatting: `isort` (profile=black) + `black`
+- Linting: `ruff` (ignores F403, F405, E501; F401 ignored in `__init__.py`)
+- Type checking: `mypy` with `ignore_missing_imports = true`
+
+## Docstrings
+
+- Docstrings should be included on all public classes, methods, and functions.
+- We use Sphinx to automatically build API docs by pulling from those docstrings.
+- The syntax of the docstrings is a superset of reStructuredText with additional Sphinx-specific syntax for things like:
+  - Cross-document links, e.g.:
+    ```
+    :class:`foo.Foo`  <- links to the class named 'Foo' in the module 'foo'
+    :mod:`foo`        <- links to the module named 'foo'
+    :func:`foo.bar`   <- links to the function named 'bar' in the module named 'foo'
+    ```
+  - Documenting parameters (`:param ...:`), return values (`:returns:`), or expected exceptions (`:raises ...:`).
+
+Here's a toy example for a function:
+
+```python
+def read_file(path: str) -> str:
+    """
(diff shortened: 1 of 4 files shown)


src/olmo_core/nn/transformer/config.py
```diff
@@ -1788,3 +1796,74 @@ def apply_scaling(block_config: TransformerBlockConfig) -> None:
...
+        n_layers=n_layers,
+        block=block,
+        block_pattern=block_pattern,
+        block_overrides=block_overrides,
+    )
+
+    block_configs: list[TransformerBlockConfig]
+    if isinstance(block, dict):
+        # Named-block configuration.
+        assert block_pattern is not None
```
epwalsh, line 1858: I guess `block_overrides` is not allowed in this case? Looks like this is also checked `validate_block_resolution_config`, but would be good to make this clear here as well:
```suggestion
        assert block_pattern is not None
        assert block_overrides is None
```

epwalsh (approved):
LGTM. Might be worth double-checking that we can load old configs from a checkpoint.

tyler-romero:
### Code review

Found 4 issues:

1. `set()` called on unhashable `TransformerBlockConfig` instances in test. `TransformerBlockConfig` is a dataclass with `eq=True` and `frozen=False` (defaults), so Python sets `__hash__=None`. `set(config.resolved_block_configs)` will always raise `TypeError: unhashable type`. Consider using `{id(b) for b in config.resolved_block_configs}` instead.

https://github.com/allenai/OLMo-core/blob/4515575d8409d957379dab55d90b199e06e9ff4f/src/test/nn/transformer/model_test.py#L648-L651

2. Divisibility constraint in `validate_block_resolution_config` is too strict for Gemma 3 models. The validation requires `n_layers % len(block_pattern) == 0`, but canonical Gemma 3 sizes are not divisible by the default pattern length of 6: gemma3_1B (26%6=2), gemma3_4B (34%6=4), gemma3_27B (62%6=2). Only gemma3_12B (48 layers) passes. The resolver already handles this correctly via `islice(cycle(block_pattern), n_layers)`. This also breaks `test_gemma3_builder_configs` which uses `n_layers=2`.

https://github.com/allenai/OLMo-core/blob/4515575d8409d957379dab55d90b199e06e9ff4f/src/olmo_core/nn/transformer/config.py#L1832-L1837

3. `test_gemma3_builder_configs` accesses `config.block.feed_forward` and `config.block.sequence_mixer`, but after this PR `config.block` is a `dict[str, TransformerBlockConfig]` for Gemma 3 configs. These attribute accesses will raise `AttributeError`. The test needs to be updated to use `config.resolved_block_configs` or index into the dict.

https://github.com/allenai/OLMo-core/blob/4515575d8409d957379dab55d90b199e06e9ff4f/src/test/nn/transformer/model_test.py#L612-L621

4. Unaddressed reviewer comment: epwalsh suggested adding `assert block_overrides is None` in the `isinstance(block, dict)` branch of `resolve_block_configs` to make the mutual exclusion with `block_overrides` explicit at the usage site (in addition to the validation in `validate_block_resolution_config`).

https://github.com/allenai/OLMo-core/blob/4515575d8409d957379dab55d90b199e06e9ff4f/src/olmo_core/nn/transformer/config.py#L1855-L1859

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #611, opened 2026-02-19
Add claude md for Olmo-core

tyler-romero:
Putting my claude.md out there. Others may have their own that they prefer so I'm open to suggestions.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -59,7 +59,7 @@ Everything is configured via `@dataclass` classes inheriting from `Config`. This
 ### Data Loading (`src/olmo_core/data/`)
 
 - `NumpyDataset` variants: Memory-mapped numpy datasets for pre-tokenized data (`.npy` files).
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,122 @@
...
+### Model Architecture (`src/olmo_core/nn/`)
+
+- `transformer/`: Core transformer with configurable blocks. `TransformerConfig` has factory methods like `olmo2_32B()` for predefined architectures.
+- `attention/`: Multi-head attention with backends (flash attention, ring attention, etc.).
+- `moe/`: Mixture of Experts with expert parallelism.
+- `feed_forward.py`, `layer_norm.py`, `rope.py`, `lm_head.py`: Standard components.
+
+### Data Loading (`src/olmo_core/data/`)
+
+- `NumpyDataset` variants: Memory-mapped numpy datasets for pre-tokenized data (`.npy` files).
```
epwalsh, line 61: Worth mentioning the composable API here too.

CLAUDE.md
```diff
@@ -0,0 +1,122 @@
...
+
+```bash
+# Install (development)
+uv sync --all-extras
+
+# Run all tests (GPU tests auto-skip without GPU)
+uv run pytest -v src/
+
+# Run a specific test file
+uv run pytest src/test/path/to/test_file.py
```
epwalsh: I still don't use uv locally (virtualfish is a very nice fish integration for Python envs), so I'd like to keep these commands uv-agnostic.

epwalsh (commented):
Nice! Just a couple of suggestions

PR #616, opened 2026-02-22
Support perplexity evals with context and tensor parallelism

tyler-romero:
## Summary
- Support PPL evals with both CP and TP in `TransformerTrainModule.eval_batch()`.
- When CP or TP is active, shard `label_mask` along the sequence dimension to match the sharded `ce_loss` shape. CP shards first via the load balancer, then TP shards further via `torch.chunk`.
- When TP is active, set `return_logits=False` and unwrap the DTensor `ce_loss` with `get_local_tensor()`.

## Test plan
- [x] Run a training job with CP and/or TP enabled with `LMEvaluator` configured — verify PPL metrics are reported
- [x] Downstream evaluators that require full logits still fail naturally with CP or TP

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Mark ephemeral checkpoints with the `ephemeral` flag in their metadata.
 - Added `ephemeral: Optional[bool]` flag the `Checkpointer.find_checkpoints()` for filtering.
 - Added support for block-pattern based initialization of hybrid transformers. `TransformerConfig.block` now accepts a dict of named `TransformerBlockConfig`s, paired with a `block_pattern` list that controls per-layer block selection.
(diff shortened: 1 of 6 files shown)


tyler-romero:
TP: https://beaker.org/orgs/ai2/workspaces/oe-t-ladder/work/01KJ66NSZFVRPC9G1P92GWTA7R

tyler-romero:
cp: https://beaker.org/orgs/ai2/workspaces/oe-t-ladder/work/01KJ626B6VGK8ZV5AGCG3JBGDG

src/olmo_core/nn/transformer/model.py
```diff
@@ -240,8 +240,11 @@ def get_rope_buffers(
             device = self.device
         rope_buffers = {}
         for key, block in self.blocks.items():
-            rope = cast(Optional[RotaryEmbeddingBase], block.attention.rope)  # type: ignore
-            rope_buffers[int(key)] = None if rope is None else rope.get_buffers(seq_len, device)
+            if isinstance(block.attention, (Attention, FusedAttention)):
```
tyler-romero, line 243: unrelated cp test fix

PR #625, opened 2026-02-26
Feature/amplegcg setup

mmdt01:
Fix AmpleGCG setup and compatibility.

- Added `sentencepiece` dependency (required by Llama tokenizer)
- Updated `generate` call with `trust_remote_code=True` to support group beam search in latest transformers
- Replaced deprecated `torch_dtype` with `dtype` in model loading


mmdt01:
Opened on wrong repo, closing. Sorry!

PR #626, opened 2026-02-26
LC Anneal

dirkgr:
There are four changes in here:
 * A script that does LC anneals, with LC data, on Olmo 3 models
 * A generalization / tightening up of the code that disables parallelism when running with `train_single`
 * `--launch.debug`, which sets `CUDA_LAUNCH_BLOCKING`
 * Fixing `dump_training_batch.py`, so that it does the right thing with datasets that shuffle themselves. Also adds an annotation to the batches so you can see which file it came from.

### DIFF
--- a/src/olmo_core/internal/experiment.py
+++ b/src/olmo_core/internal/experiment.py
@@ -146,7 +146,13 @@ class SubCmd(StrEnum):
             train(config)
             teardown_training_environment()
         elif self == SubCmd.train_single:
-            for parallelism_style in ["dp_config", "tp_config", "cp_config", "pp_config", "ep_config"]:
+            for parallelism_style in [
+                "dp_config",
+                "tp_config",
+                "cp_config",
+                "pp_config",
+                "ep_config",
+            ]:
                 if (p_config := getattr(config.train_module, parallelism_style, None)) is not None:
                     log.warning(
                         "'%s' is set to %s, but you can't use parallelism when running on a single node. Disabling.",


dirkgr:
### Code review

Found 2 issues:

1. **Missing `ep_config` in `train_single` parallelism-disabling loop.** The loop disables `dp_config`, `tp_config`, `cp_config`, and `pp_config` but omits `ep_config` (expert parallelism). `TransformerTrainModule.__init__` checks all five (including `ep_config`) and raises `OLMoConfigurationError` if any are non-None when not running distributed, so any MoE script using `train_single` would crash instead of silently disabling EP.

https://github.com/allenai/OLMo-core/blob/fcaa57c6379645ef09e932661c0f215051dc0856/src/olmo_core/internal/experiment.py#L149

2. **`NCCL_DEBUG=INFO` dropped from `--debug` flag.** The old `_build_config()` set both `CUDA_LAUNCH_BLOCKING=1` and `NCCL_DEBUG=INFO` when `--debug` was passed. The refactored `BeakerLaunchConfig.default_env_vars` only sets `CUDA_LAUNCH_BLOCKING=1`, silently losing NCCL debug output.

https://github.com/allenai/OLMo-core/blob/fcaa57c6379645ef09e932661c0f215051dc0856/src/olmo_core/launch/beaker.py#L430-L431

Generated with [Claude Code](https://claude.ai/code)

src/scripts/dump_training_batch.py
```diff
@@ -188,14 +176,45 @@ def main():
...
+
+    # Pre-compute source_sizes from the already-cached file_sizes to avoid a second round of
+    # ~960 concurrent HEAD requests to R2 that overwhelms the connection pool. The file_sizes
+    # property was already populated during prepare(), and source_sizes is just file_sizes
+    # divided by item_size, but it's a separate property that would re-query every path.
+    from olmo_core.data.numpy_dataset import NumpyPackedFSLDataset
+
+    if isinstance(dataset, NumpyPackedFSLDataset):
+        item_size = dataset.dtype(0).itemsize
+        dataset._source_sizes = [s // item_size for s in dataset.file_sizes]
```
epwalsh, line 217: How about fixing this here instead: https://github.com/allenai/OLMo-core/blob/cc940d64bdce80341ad942fe2ffba114ec977ea3/src/olmo_core/data/numpy_dataset.py#L1105

src/olmo_core/launch/beaker.py
```diff
@@ -422,6 +427,9 @@ def default_env_vars(self) -> list[tuple[str, str]]:
         ]
         if self.shared_filesystem:
             env_vars.append((OLMO_SHARED_FS_ENV_VAR, "1"))
+        if self.debug:
+            env_vars.append(("CUDA_LAUNCH_BLOCKING", "1"))
+            env_vars.append(("NCCL_DEBUG", "INFO"))
```
epwalsh, line 432: This is already in the default env vars: https://github.com/allenai/OLMo-core/blob/cc940d64bdce80341ad942fe2ffba114ec977ea3/src/olmo_core/launch/beaker.py#L416

epwalsh: which kinda annoys me TBH. I think we should default to WARN

src/scripts/train/OLMo3/OLMo3-7B-LC-anneal.py
```diff
@@ -0,0 +1,197 @@
...
+from olmo_core.train.train_module import (
+    TransformerActivationCheckpointingConfig,
+    TransformerContextParallelConfig,
+    TransformerTrainModuleConfig,
+)
+
+log = logging.getLogger(__name__)
+
+
+if __name__ == "__main__":
```
epwalsh, line 39: Kind of odd and error-prone to define the whole script globally in the `if __name__ == "__main__"` block. It's easy to mix up locals and globals, and adds unnecessary indentation.

epwalsh (changes requested):

PR #636, opened 2026-03-11
Start a v3 ladder with zero-centered RMSNorm

YashasSamaga:
A new v3 ladder.

Changes:
- adds embedding norm to v3

Other changes which are not in the v3 ladder:
- adds zero-centered RMSNorm that has a more natural resting state for weight decay

Pending:
- some restructuring is required for versioning correctly; parts of the code are hardcoded to v2 

More v3 changes will be added in separate PRs.

### DIFF
--- a/src/olmo_core/nn/layer_norm.py
+++ b/src/olmo_core/nn/layer_norm.py
@@ -65,7 +65,7 @@ class LayerNormConfig(ModuleConfig):
     bias: Optional[bool] = None
     full_precision: Optional[bool] = None
     dtype: Optional[DType] = None
-    zero_centered_weights: Optional[bool] = None # use reparameterized (1 + weight) version
+    zero_centered_weights: Optional[bool] = None  # use reparameterized (1 + weight) version
 
     def num_params(self, size: int) -> int:
         """
--- a/src/olmo_core/nn/layer_norm.py
+++ b/src/olmo_core/nn/layer_norm.py
@@ -65,6 +65,7 @@ class LayerNormConfig(ModuleConfig):
     bias: Optional[bool] = None
     full_precision: Optional[bool] = None
     dtype: Optional[DType] = None
+    zero_centered_weights: Optional[bool] = None # use reparameterized (1 + weight) version
 
     def num_params(self, size: int) -> int:
         """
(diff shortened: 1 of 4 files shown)


src/scripts/train/ladder/gemma_like_ladder.py
```diff
@@ -489,6 +489,265 @@ def v2_34B(cls, vocab_size: int, **kwargs) -> "TransformerConfig":
             **kwargs,
         )
 
+    @classmethod
+    def v3(cls, vocab_size: int, **kwargs) -> "TransformerConfig":
+        # Clone of v2 with zero-centered RMSNorm weights (zero_centered_weights=True).
```
YashasSamaga: I'm not really sure if replicating this block of code was necessary, but I don't think there was a clean way to edit layernorm configuration in all layers unless the configs are a shared object across all uses (like fiddle configs).

src/scripts/train/ladder/gemma_like_ladder.py
```diff
@@ -489,6 +489,265 @@ def v2_34B(cls, vocab_size: int, **kwargs) -> "TransformerConfig":
...
+
```
dirkgr:
```suggestion
        # Clone of v2 with zero-centered RMSNorm weights (zero_centered_weights=True).
        # More changes to come until we declare v3 done.

```

dirkgr (approved):
Looks great, just this one change!

src/olmo_core/nn/layer_norm.py
```diff
@@ -203,8 +204,49 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        self.zero_centered_weights = zero_centered_weights
+        super().__init__(
+            size=size,
+            eps=eps,
+            elementwise_affine=elementwise_affine,
+            bias=bias,
+            full_precision=full_precision,
+            dtype=dtype,
+            init_device=init_device,
+        )
```
YashasSamaga, line 233: I'm unable to decide whether to explicitly list out arguments or forward via `*args` and `*kwargs`. This may impact static analysis tools and documentation generators.

dirkgr: That's very good thinking. This is the reason I don't like `kwargs`. Copying it is better. We might be modifying this later.

PR #653, opened 2026-04-02
Use metadata-backed document boundaries for SFT datasets

### LINKED ISSUE #594: SFT for Qwen Models
My setup is the following:

I take Qwen3-1.7B base model from [here](https://huggingface.co/Qwen/Qwen3-1.7B-Base). 
Then for the tokenizer and chat_template I use the ones that Qwen use for their instruct model [here](https://huggingface.co/Qwen/Qwen3-1.7B?chat_template=default).
I then take OLMo's [dolci-sft-instruct](https://huggingface.co/datasets/allenai/Dolci-Instruct-SFT) dataset, pretokenize it using open-instruct script [here](https://github.com/allenai/open-instruct/blob/main/scripts/data/convert_sft_data_for_olmocore.py) and use it to train the Qwen base model, hoping to achieve Qwen3-instruct performance on IFEval.

Result:
IFEval degrading throughout training, showing unrelated responses to instructions.

Debugging:
I immediately suspected chat_template and tokenization but while debugging the pretokenization on open-instruct I saw overall reasonable results and correct usage of Qwen's tokenizer and chat_template. Anyway I tried to switch to Olmo's chat_template but to no avail, same results again. Same for trying padding instead of packing.

After some debugging, it appears the issue lies in olmo-core's assumption (I debugged it to [this method](https://github.com/allenai/OLMo-core/blob/32e40e31fd0507178f10e0c096953c8d16cb06b6/src/olmo_core/data/utils.py#L170), I think) that tokenized SFT conversations end with an eos token (<|endoftext|> for olmo) such that the Dataset object (whether it's the Packed or Padded one) will dissect them accordingly. For Qwen models their default chat_template end *all* messages, including the last one in a conversation with an <|im_end|> token. But then switching the chat_template to the Olmo one should have helped, shouldn't it? It didn't, because Qwen-Instruct's eos token is <|im_end|> (look [here](https://huggingface.co/Qwen/Qwen3-1.7B/blob/main/tokenizer_config.json#L232)), and while olmo-core tries to dissect data with each model's eos token, when doing it for a Qwen model, even when using Olmo's chat_template, all messages and conversations end with <|im_end|>, losing the ability to be dissected correctly.

Solution:
Eventually I modified Qwen's chat_template to end conversations with <|endoftext|> like Olmo's chat_template AND changed its eos token from <|im_end|> to <|endoftext|> in tokenizer_config.json such that olmo-core will take it from there.
The next SFT training ran correctly, I looked at the samples right before they get into the model and they look fine, and IFEval on the resulting model looks great and compares to Qwen-Instruct.

I'm writing this because I can keep doing what I'm doing but I'm not sure if this is the intended way to solve it (overriding the tokenizer's eos and chat_template of all models intended to be trained on olmo-core) and I'd be happy to hear your take on this.

Cheers,
Gabi.

taivu1998:
## Summary
This PR fixes issue #594 by making document-aware SFT paths prefer sibling `.csv.gz` boundary metadata over rediscovering boundaries from tokenizer EOS tokens.

The key behavior changes are:
- add explicit `DocumentBoundaryMode` support with `auto`, `metadata`, and `tokenizer` modes
- make `NumpyPackedFSLDataset`, `NumpyPaddedFSLDataset`, and `NumpyDocumentSource` use metadata-backed boundaries in `auto` mode when metadata exists for all source files
- reject mixed metadata presence in `auto` mode instead of silently mixing modes
- compute packed `doc_lens` from the packed document spans instead of rescanning token values
- include effective boundary mode and metadata file identity in fingerprints/cache keys when metadata-backed boundaries are used
- update the SFT README to reflect the new metadata-backed behavior

## Motivation
Issue #594 showed that Qwen-style SFT data can break when the tokenizer EOS token also appears at message boundaries. In that case, recovering document boundaries from tokenizer EOS fragments one conversation into many smaller documents.

The upstream SFT conversion pipeline already writes authoritative conversation boundaries into sibling `.csv.gz` metadata files, so this PR uses that metadata directly in the document-aware dataset paths implicated by the issue.

## Design Notes
The change is intentionally narrow:
- low-level utility helpers still default to tokenizer-based behavior so unrelated dataset paths do not change semantics accidentally
- metadata-first behavior is enabled explicitly in the document-aware packed, padded, and composable source paths that need it
- packed runtime `doc_lens` now come from known packed spans, which removes an unnecessary EOS-token dependency from the masking path

## Testing
Focused regression coverage added for:
- metadata precedence over local EOS scanning
- mixed metadata presence failures
- packed `doc_lens` generation from metadata-backed spans
- padded metadata-backed segmentation
- composable `NumpyDocumentSource` metadata-backed offsets
- fingerprint/cache changes when metadata changes

Commands run:
- `uv run pytest src/test/data/utils_test.py src/test/data/numpy_dataset_test.py src/test/data/composable/numpy_document_source_test.py -q`
- `uv run pytest src/test/data -k 'not mixes_test' -q`
- `UV_CACHE_DIR=/tmp/uv-cache uv run ruff check src/olmo_core/data/types.py src/olmo_core/data/utils.py src/olmo_core/data/numpy_dataset.py src/olmo_core/data/composable/numpy_document_source.py src/olmo_core/data/__init__.py src/test/data/utils_test.py src/test/data/numpy_dataset_test.py src/test/data/composable/numpy_document_source_test.py`

## Related
- Closes #594

### DIFF
--- a/src/olmo_core/data/__init__.py
+++ b/src/olmo_core/data/__init__.py
@@ -48,7 +48,7 @@ from .numpy_dataset import (
     VSLNaturalCurriculum,
 )
 from .tokenizer import TokenizerConfig, TokenizerName
-from .types import LongDocStrategy, NumpyDatasetDType
+from .types import DocumentBoundaryMode, LongDocStrategy, NumpyDatasetDType
 
 __all__ = [
     "composable",
@@ -73,6 +73,7 @@ __all__ = [
     "VSLCurriculumType",
     "VSLCurriculumConfig",
     "NumpyDatasetDType",
+    "DocumentBoundaryMode",
     "TokenizerConfig",
     "TokenizerName",
     "DataMixBase",
--- a/src/olmo_core/data/composable/numpy_document_source.py
+++ b/src/olmo_core/data/composable/numpy_document_source.py
@@ -16,12 +16,13 @@ from olmo_core.utils import log_once
 
 from ..mixes import DataMix, DataMixBase
 from ..tokenizer import TokenizerConfig
-from ..types import LongDocStrategy, NumpyDatasetDType, NumpyUIntTypes
+from ..types import DocumentBoundaryMode, LongDocStrategy, NumpyDatasetDType, NumpyUIntTypes
 from ..utils import (
     chunked,
     iter_document_indices,
     iter_document_indices_with_max_sequence_length,
     load_array_slice,
+    resolve_document_boundary_mode,
 )
 from .token_source import DocumentSource, DocumentSourceConfig, TokenRange
 from .utils import path_map, resolve_seed
@@ -52,6 +53,10 @@ class NumpyDocumentSourceConfigBase(DocumentSourceConfig):
     long_doc_strategy: LongDocStrategy = LongDocStrategy.truncate
     """
     How to handle long documents when ``max_document_length`` is set.
+    """
+    document_boundaries: DocumentBoundaryMode = DocumentBoundaryMode.auto
+    """
+    How to resolve document boundaries for document-aware iteration.
     """
 
     def __post_init__(self):
@@ -218,6 +223,7 @@ class NumpyDocumentSourceConfig(NumpyDocumentSourceConfigBase):
             label=label,
             max_document_length=self.max_document_length,
             long_doc_strategy=self.long_doc_strategy,
+            document_boundaries=self.document_boundaries,
         )
 
         if self.source_group_size > 0:
@@ -337,6 +343,7 @@ class NumpyDocumentSource(DocumentSource):
         label: Optional[str] = None,
         max_document_length: Optional[int] = None,
         long_doc_strategy: LongDocStrategy = LongDocStrategy.truncate,
+        document_boundaries: DocumentBoundaryMode = DocumentBoundaryMode.auto,
         _source_sizes: Optional[Sequence[int]] = None,
         _label_mask_sizes: Optional[Sequence[int]] = None,
     ):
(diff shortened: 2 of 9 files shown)


taivu1998:
Hi @tyler-romero, could you help review this PR? Thanks!

pangwangshu:
Thanks for this — checked the metadata format against `open_instruct/numpy_dataset_conversion.py` and it lines up, so the core mechanism is sound. A few things before merging:

**1. `auto` default reaches beyond SFT.** `NumpyPackedFSLDataset`/`NumpyPaddedFSLDataset`/`NumpyDocumentSource` are also used by non-SFT scripts — e.g. `OLMo-3-1025-7B-long-context.py` uses `NumpyPackedFSLDatasetConfig` as its main training dataset. If any pretraining data already has sibling `.csv.gz` files (the pre-existing `write_document_indices` docstring suggests Dolma-tokenized data might), upgrading would silently change document-boundary resolution there too — no validation that metadata is consistent with the configured `eos_token_id`, just an existence check. Could we default to `tokenizer` and have SFT opt into `auto`, or confirm no pretraining data has stray `.csv.gz` siblings?

**2. Composable data loader not covered.** `composable/data_loader.py`'s `get_instance` still computes `doc_lens` via a fresh EOS scan (`get_document_lengths`), ignoring metadata-backed boundaries — so the #594 bug would still hit attention masking there. Not a blocker (current SFT scripts use the classic path), but worth a follow-up issue.

**3. Startup latency.** `resolve_document_boundary_mode` calls `file_exists()` per source path, synchronously, in `__init__`. For remote (GCS/S3/Weka) mixtures with many shards, that's serialized network round-trips added to every startup. Worth parallelizing.

**4. Unrelated behavior change.** `_write_instance_indices` now passes `bos_token_id` into `segment_documents_into_instances`, which it didn't before — changes doc-splitting for any padded-dataset consumer with `bos_token_id` set, independent of this PR's main change. Worth its own commit/callout.

Test coverage on the metadata precedence and mixed-presence cases looks solid, for what it's worth.

PR #654, opened 2026-04-02
Add position_ids-based varlen RoPE support for packed inputs

### LINKED ISSUE #503: [Feature Request] Support variable-length sequences in RoPE natively

taivu1998:
## Summary
- add native varlen RoPE support via explicit `position_ids`
- auto-derive `position_ids` from `doc_lens` in the transformer input path for packed and intra-document-masked batches
- thread `position_ids` through attention and RoPE while preserving existing contiguous and decode-offset behavior
- add direct helper coverage plus rope, attention, and transformer regression tests

## Design
This change follows the minimal design from issue #503: `position_ids` is the advanced RoPE API, while the common packed-training path can continue passing `doc_lens` and let the transformer derive positions automatically.

The implementation intentionally keeps unsupported paths loud instead of adding partial behavior:
- `position_ids` with fused RoPE raises `NotImplementedError`
- `position_ids` with KV caching raises `RuntimeError`
- context-parallel attention now accepts either pre-sharded RoPE buffers or pre-sharded `position_ids`

## Implementation details
- add `get_position_ids_from_doc_lens()` to build per-token RoPE positions that reset at document boundaries
- update `RotaryEmbedding` and `ComplexRotaryEmbedding` to gather RoPE values from explicit per-token positions
- keep the existing `start_pos` and contiguous-sequence path unchanged when `position_ids` are not provided
- plumb `position_ids` through `Attention`, `NormalizedAttention`, and `Transformer`
- auto-derive `position_ids` from `doc_lens` before CP sharding so the same abstraction works in regular and CP input preparation

## Tests
- `UV_CACHE_DIR=/tmp/uv-cache uv run pytest src/test/data/utils_test.py src/test/nn/rope_test.py src/test/nn/attention/attention_test.py src/test/nn/transformer/model_test.py -q`
  - `198 passed, 1073 skipped`

Closes #503

### DIFF
--- a/src/olmo_core/data/utils.py
+++ b/src/olmo_core/data/utils.py
@@ -399,7 +399,8 @@ def get_position_ids_from_doc_lens(doc_lens: torch.Tensor, seq_len: int) -> torc
     This is intended for packed or intra-document masked training batches where every document
     should reset its RoPE positions back to zero. Padding to ``seq_len`` is filled with zeros.
 
-    :param doc_lens: A 1D or 2D tensor of document lengths.
+    :param doc_lens: A 1D tensor of document lengths for a single packed instance or a 2D batch
+        of zero-padded document lengths.
     :param seq_len: The padded sequence length for each batch item.
     """
     squeeze_output = False
@@ -409,25 +410,37 @@ def get_position_ids_from_doc_lens(doc_lens: torch.Tensor, seq_len: int) -> torc
     elif doc_lens.ndim != 2:
         raise ValueError(f"'doc_lens' must be 1D or 2D (got {doc_lens.ndim}D)")
 
-    position_ids = torch.zeros(
-        (doc_lens.size(0), seq_len), dtype=torch.long, device=doc_lens.device
+    if doc_lens.numel() == 0 or doc_lens.size(1) == 0:
+        position_ids = torch.zeros(
+            (doc_lens.size(0), seq_len), dtype=torch.long, device=doc_lens.device
+        )
+        return position_ids[0] if squeeze_output else position_ids
+
+    if doc_lens.numel() and int(doc_lens.min().item()) < 0:
+        raise ValueError(
+            f"document lengths must be non-negative (got {int(doc_lens.min().item())})"
+        )
+
+    cumulative_doc_ends = torch.cumsum(doc_lens, dim=1, dtype=torch.long)
+    total_lengths = cumulative_doc_ends[:, -1]
+    max_total_length = int(total_lengths.max().item())
+    if max_total_length > seq_len:
+        raise ValueError(
+            f"document lengths sum to {max_total_length}, which exceeds seq_len={seq_len}"
+        )
+
+    # Keep this fully tensorized since it runs in the packed-training hot path.
+    token_positions = (
+        torch.arange(seq_len, device=doc_lens.device, dtype=torch.long)
+        .expand(doc_lens.size(0), -1)
+        .contiguous()
     )
-    for row_idx, row in enumerate(doc_lens):
-        offset = 0
-        for doc_len in row.tolist():
-            if doc_len == 0:
-                continue
-            if doc_len < 0:
-                raise ValueError(f"document lengths must be non-negative (got {doc_len})")
-            next_offset = offset + doc_len
-            if next_offset > seq_len:
-                raise ValueError(
-                    f"document lengths sum to {next_offset}, which exceeds seq_len={seq_len}"
-                )
-            position_ids[row_idx, offset:next_offset] = torch.arange(
-                doc_len, device=doc_lens.device, dtype=torch.long
-            )
-            offset = next_offset
+    doc_indices = torch.searchsorted(cumulative_doc_ends.contiguous(), token_positions, right=True)
+    doc_indices.clamp_(max=doc_lens.size(1) - 1)
+
+    doc_start_offsets = cumulative_doc_ends - doc_lens.to(dtype=torch.long)
+    position_ids = token_positions - doc_start_offsets.gather(1, doc_indices)
+    position_ids.masked_fill_(token_positions >= total_lengths.unsqueeze(1), 0)
 
     return position_ids[0] if squeeze_output else position_ids
 
--- a/src/olmo_core/data/utils.py
+++ b/src/olmo_core/data/utils.py
@@ -392,6 +392,46 @@ def get_cumulative_document_lengths(doc_lens: torch.Tensor) -> torch.Tensor:
     )
 
 
+def get_position_ids_from_doc_lens(doc_lens: torch.Tensor, seq_len: int) -> torch.Tensor:
+    """
+    Build per-token RoPE positions from document lengths.
+
+    This is intended for packed or intra-document masked training batches where every document
+    should reset its RoPE positions back to zero. Padding to ``seq_len`` is filled with zeros.
+
+    :param doc_lens: A 1D or 2D tensor of document lengths.
+    :param seq_len: The padded sequence length for each batch item.
+    """
+    squeeze_output = False
+    if doc_lens.ndim == 1:
+        doc_lens = doc_lens.unsqueeze(0)
+        squeeze_output = True
+    elif doc_lens.ndim != 2:
+        raise ValueError(f"'doc_lens' must be 1D or 2D (got {doc_lens.ndim}D)")
+
+    position_ids = torch.zeros(
+        (doc_lens.size(0), seq_len), dtype=torch.long, device=doc_lens.device
+    )
+    for row_idx, row in enumerate(doc_lens):
+        offset = 0
+        for doc_len in row.tolist():
+            if doc_len == 0:
+                continue
+            if doc_len < 0:
+                raise ValueError(f"document lengths must be non-negative (got {doc_len})")
+            next_offset = offset + doc_len
+            if next_offset > seq_len:
+                raise ValueError(
+                    f"document lengths sum to {next_offset}, which exceeds seq_len={seq_len}"
+                )
+            position_ids[row_idx, offset:next_offset] = torch.arange(
+                doc_len, device=doc_lens.device, dtype=torch.long
+            )
+            offset = next_offset
+
+    return position_ids[0] if squeeze_output else position_ids
+
+
 def iter_batched(
     iterable: Iterable[Dict[str, Any]], batch_num_tokens: int
 ) -> Iterable[Tuple[Dict[str, Any], ...]]:
--- a/src/olmo_core/nn/attention/__init__.py
+++ b/src/olmo_core/nn/attention/__init__.py
@@ -484,6 +484,22 @@ class Attention(SequenceMixer):
     def cp_enabled(self) -> bool:
         return self.backend.cp_enabled
 
+    def _validate_rope_forward_inputs(
+        self,
+        *,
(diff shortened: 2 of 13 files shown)


tyler-romero:
Nice tests, thanks for that. Any idea how this affects training throughput?

src/olmo_core/nn/attention/__init__.py
```diff
@@ -888,10 +899,18 @@ def forward(
         v = v.view(B, T, self.n_kv_heads, self.head_dim)
 
         if self.rope is not None:
-            if self.cp_enabled and pos_sin is None and pos_cos is None and freqs_cis is None:
+            if self.kv_cache_manager is not None and position_ids is not None:
+                raise RuntimeError("position_ids are not supported with KV caching")
+            if (
```
finbarrtimbers: Given how complicated this check is, I think it's worth pulling it out into a function/method.

src/olmo_core/data/utils.py
```diff
@@ -392,6 +392,46 @@ def get_cumulative_document_lengths(doc_lens: torch.Tensor) -> torch.Tensor:
...
+    elif doc_lens.ndim != 2:
+        raise ValueError(f"'doc_lens' must be 1D or 2D (got {doc_lens.ndim}D)")
+
+    position_ids = torch.zeros(
+        (doc_lens.size(0), seq_len), dtype=torch.long, device=doc_lens.device
+    )
+    for row_idx, row in enumerate(doc_lens):
+        offset = 0
+        for doc_len in row.tolist():
+            if doc_len == 0:
```
finbarrtimbers: I think it'd be clearer to just assert doc_len > 0; we don't need to handle both cases, it's fine to let the caller do that imo.

src/olmo_core/data/utils.py
```diff
@@ -392,6 +392,46 @@ def get_cumulative_document_lengths(doc_lens: torch.Tensor) -> torch.Tensor:
...
+    if doc_lens.ndim == 1:
+        doc_lens = doc_lens.unsqueeze(0)
+        squeeze_output = True
+    elif doc_lens.ndim != 2:
+        raise ValueError(f"'doc_lens' must be 1D or 2D (got {doc_lens.ndim}D)")
+
+    position_ids = torch.zeros(
+        (doc_lens.size(0), seq_len), dtype=torch.long, device=doc_lens.device
+    )
+    for row_idx, row in enumerate(doc_lens):
```
finbarrtimbers: Can you help me understand why we'd want to have it be a 2d tensor and have this outer loop? Is the case a batch of packed tensors?

src/test/data/utils_test.py
```diff
@@ -184,6 +185,44 @@ def test_get_cumulative_document_lengths():
     ).tolist() == [0, 1, 6, 9, 11, 16, 19, 22]
 
 
+def test_get_position_ids_from_doc_lens():
```
finbarrtimbers: Parameterize this?

src/olmo_core/nn/transformer/model.py
```diff
@@ -406,25 +424,33 @@ def _prepare_inputs(
...
+                inputs.append(position_ids)
+                seq_dims.append(1)
+                pad_values.append(0)
+                keys.append("position_ids")
+            else:
+                # NOTE: initialize buffer(s) on CPU to avoid possible host-device sync when sharding.
+                for block_idx, rope_buffers in self.get_rope_buffers(
+                    S, torch.device("cpu")
+                ).items():
+                    if rope_buffers is not None:
```
finbarrtimbers: To remove one level of nesting, I'd do: 

```
if nope_buffers is None:
    continue
```

src/olmo_core/nn/transformer/model.py
```diff
@@ -406,25 +424,33 @@ def _prepare_inputs(
...
+                        # Also shard RoPE buffers based on the context parallelism load balancer.
+                        if rope_buffers.pos_sin is not None:
```
finbarrtimbers: Make this a function and call it repeatedly, please.

src/olmo_core/nn/rope.py
```diff
@@ -509,36 +511,78 @@ def forward(
...
+                    raise RuntimeError(
+                        f"'pos_sin' and 'pos_cos' are invalid when using 'position_ids' with "
+                        f"{self.__class__.__name__}"
+                    )
+                if q_len > k_len:
+                    raise RuntimeError(
+                        f"'position_ids' requires q_len <= k_len (got q_len={q_len}, k_len={k_len})"
+                    )
+
+                position_ids = position_ids.to(device=q_.device, dtype=torch.long)
```
finbarrtimbers: Let's move this after the remaining checks, so that we can fail faster when the checks raise.

src/olmo_core/nn/rope.py
```diff
@@ -509,36 +511,78 @@ def forward(
...
-                )
-
-            if head_first:
-                sin_q = pos_sin[q_abs_start : q_abs_start + q_len, :][None, None, :, :]
-                cos_q = pos_cos[q_abs_start : q_abs_start + q_len, :][None, None, :, :]
-                sin_k = pos_sin[k_abs_start : k_abs_start + k_len, :][None, None, :, :]
-                cos_k = pos_cos[k_abs_start : k_abs_start + k_len, :][None, None, :, :]
+                pos_sin, pos_cos = pos_sin.type_as(q_), pos_cos.type_as(q_)
+
+                if head_first:
```
finbarrtimbers, line 566: Let's make `position_ids[:, k_len - q_len :]` a named variable as it's used a bunch here.

src/olmo_core/nn/rope.py
```diff
@@ -509,36 +511,78 @@ def forward(
...
                 k_ = self._apply_rotary_pos_emb(sin_k, cos_k, k_)
             else:
-                sin_q = pos_sin[q_abs_start : q_abs_start + q_len, :][None, :, None, :]
-                cos_q = pos_cos[q_abs_start : q_abs_start + q_len, :][None, :, None, :]
-                sin_k = pos_sin[k_abs_start : k_abs_start + k_len, :][None, :, None, :]
-                cos_k = pos_cos[k_abs_start : k_abs_start + k_len, :][None, :, None, :]
-
-                q_ = self._apply_rotary_pos_emb(sin_q, cos_q, q_)
-                k_ = self._apply_rotary_pos_emb(sin_k, cos_k, k_)
+                seq_len_needed = (start_pos + k_len) if start_pos is not None else k_len
```
finbarrtimbers, line 580: Great variable name.

src/olmo_core/nn/rope.py
```diff
@@ -769,26 +818,64 @@ def forward(
         k_ = torch.view_as_complex(k_.reshape(*k_.shape[:-1], -1, 2))
 
         with torch.autocast(q.device.type, enabled=False):
-            # shape: (T, hs // 2)
-            seq_len_needed = (start_pos + k_len) if start_pos is not None else k_len
-            if freqs_cis is None:
+            if position_ids is not None:
```
finbarrtimbers, line 846: These checks are remarkably similar to the checks in `attention/__init__.py`; can we pull them out into a shared validation function?

src/olmo_core/nn/rope.py
```diff
@@ -769,26 +818,64 @@ def forward(
...
+                    raise RuntimeError(
+                        f"'freqs_cis' is invalid when using 'position_ids' with "
+                        f"{self.__class__.__name__}"
+                    )
+                if q_len > k_len:
+                    raise RuntimeError(
+                        f"'position_ids' requires q_len <= k_len (got q_len={q_len}, k_len={k_len})"
+                    )
+
+                position_ids = position_ids.to(device=q_.device, dtype=torch.long)
```
finbarrtimbers: Same comment about moving this call applies here.

finbarrtimbers (changes requested):
Generally, this looks fine, but I have a bunch of small changes requested. 

I'm not super familiar with this code, so feel free to push back if what I'm saying seems off.

taivu1998:
Hi @finbarrtimbers, thank you so much for the comments. I updated the code to address them. Could you help review the PR again?

taivu1998:
> Nice tests, thanks for that. Any idea how this affects training throughput?

Thanks for the question. I haven’t run an end-to-end throughput benchmark yet, but the position_ids construction helper is now faster than the original loop-based version, so I’m not expecting a meaningful training regression from this path.

taivu1998:
Hi @finbarrtimbers, Could you help review the PR again? Thanks!

src/olmo_core/nn/transformer/model.py
```diff
@@ -391,13 +427,23 @@ def _prepare_inputs(
...
         doc_lens: Optional[torch.Tensor] = None
+        position_ids: Optional[torch.Tensor] = kwargs.pop("position_ids", None)
         cache_leftpad: Optional[torch.Tensor] = kwargs.pop("cache_leftpad", None)
 
-        if (doc_lens := kwargs.pop("doc_lens", None)) is not None and (
-            max_doc_lens := kwargs.pop("max_doc_lens", None)
-        ) is not None:
-            max_doc_len = max(max_doc_lens)
-            cu_doc_lens = get_cumulative_document_lengths(doc_lens)
+        doc_lens = kwargs.pop("doc_lens", None)
```
finbarrtimbers, line 433: What about this: 

```
doc_lens = kwargs.pop("doc_lens", None)
max_doc_lens = kwargs.pop("max_doc_lens", None)

if max_doc_lens is not None and doc_lens is None:
    raise ValueError("'max_doc_lens' is invalid without 'doc_lens'")

if doc_lens is not None and position_ids is None and self._has_rope():
    position_ids = get_position_ids_from_doc_lens(doc_lens, S)

if max_doc_lens is not None:
    max_doc_len = max(max_doc_lens)
    cu_doc_lens = get_cumulative_document_lengths(doc_lens)

if position_ids is not None:
    self._validate_position_ids(position_ids, B, S)
    position_ids = position_ids.to(device=input_ids.device, dtype=torch.long)
```

finbarrtimbers (approved):

finbarrtimbers:
Thanks for pinging; apologies for not following up!

taivu1998:
Hi @finbarrtimbers, thanks! I just updated the PR, could you take a look and see if it's good to merge?

finbarrtimbers (approved):

finbarrtimbers:
LGTM! Please fix the failing CI errors and then you're good to merge.

taivu1998:
Thanks @finbarrtimbers, I fixed the issue. Could you help merge this PR? Thanks!

finbarrtimbers:
Hey @taivu1998 we had a bunch of failing tests so I have reverted this.

PR #659, opened 2026-04-21
Fall back to anonymous GCS client when no credentials are available

baileykuehl:
Allows external users without GCS credentials to read from public GCS folders


undfined (approved):
seems fine

PR #660, opened 2026-04-22
Add HFConverterCallback for end-of-training HuggingFace conversion

AkshitaB:
## Summary

- Add `HFConverterCallback`, a training callback that converts the final OLMo Core checkpoint to HuggingFace format at the end of training. All ranks collectively gather the full (FSDP-unsharded) model state dict and rank 0 performs the conversion.
- Refactor the conversion logic out of the example CLI script into `olmo_core.nn.hf.convert_checkpoint` so it can be reused programmatically. The example script at `src/examples/huggingface/convert_checkpoint_to_hf.py` becomes a thin CLI wrapper.
- Extend `convert_checkpoint_to_hf` with an optional `model_state_dict` kwarg. When provided, it is used directly (and loaded into the fresh model for validation) instead of reading `model_and_optim` from disk — this is the path the callback takes.

## Test plan

- [x] `pytest src/test/examples/huggingface/convert_checkpoint_to_hf_test.py` (exercises `convert_checkpoint_to_hf` via the CLI wrapper re-export)
- [x] `pytest src/test/examples/huggingface/convert_checkpoint_to_hf_hybrid_test.py`
- [x] Run a short training job with `HFConverterCallback` attached and verify the HF checkpoint is produced alongside the final OLMo Core checkpoint


farhatkevin:
LGTM, thanks for adding the test.

farhatkevin (approved):

PR #664, opened 2026-04-30
Pin flash-linear-attention version to 0.4.1

AkshitaB:
... till we fix tilelang


YashasSamaga (approved):

YashasSamaga:
I wonder if we should have an open issue to track progress on tilelang integration so that this doesn't become a permanent version lock in. The FLA issue only affects Hopper GPUs and this blocks upgrades on all other architectures where this is a non-issue.

PR #671, opened 2026-05-12
Add ComposableScheduler

AkshitaB:
## Summary

Adds `ComposableScheduler` — a piecewise LR scheduler — plus a reusable `OverrideDecay` primitive that's wired into both `ComposableScheduler` and `SequentialScheduler` for mid-flight late-stage anneals.

The first two commits (`ComposableScheduler` itself and the override) are original work by @TianhuaTao; this PR ports just those scheduler changes onto `main` (preserving authorship and dates) and refines the API.

### `ComposableScheduler` (registered as `"composable"`)

A piecewise LR scheduler built from `ComposableSchedulerStage` segments (linear or cosine interpolation between endpoint LRs) on an absolute time axis.

- **`ComposableSchedulerStage`**: `duration`, `shape` (`linear`/`cosine`), `start_lr`/`start_lr_fraction` (optional — defaults to previous stage's end LR), `end_lr`/`end_lr_fraction` (required).
- **Continuity**: stages chain via "previous end LR = next start LR" by default, so a list of stages composes into one continuous curve.
- **`t_max` is ignored**. The schedule is defined absolutely by the sum of stage durations; it does not rescale to fit the trainer's horizon. A once-per-instance `UserWarning` surfaces this on the first `get_lr` call, showing both the configured total duration and the trainer's `t_max`.

### `OverrideDecay`

A late-stage override that interrupts the main schedule mid-flight. Available on both schedulers via an `override_decay` field.

- **`OverrideDecay`**: `start`, `duration`, `shape`, `end_lr`/`end_lr_fraction`.
- When `current >= override_decay.start`, the LR is interpolated from "whatever the main schedule would have produced at `start`" to the override's target over `duration` (linear/cosine), then held at the target.
- On `SequentialScheduler`, `t_max` is honored during the main sequence but ignored once the override becomes active. A once-per-instance `UserWarning` fires when the override first triggers.

## Behavior matrix

| Scheduler | `t_max` behavior |
|---|---|
| `ComposableScheduler` | Ignored. Schedule is absolute. Warns once on first `get_lr`. |
| `SequentialScheduler` (no override) | Honored — threaded through sub-schedulers, last segment stretches to `t_max - sum(prior maxes)`. |
| `SequentialScheduler` (override active) | Honored before `override.start`; ignored once override fires. Warns once. |

When `t_max` equals the sum of `ComposableScheduler` stage durations *and* `SequentialScheduler`'s `schedulers_max` are placed at the stage boundaries, the two schedulers produce identical LR curves. Covered by `test_composable_and_sequential_match_when_t_max_aligns` (parametrized linear/cosine × two `alpha_f` values).

## Tests

`src/test/optim/scheduler_test.py` gains:

- `test_composable_scheduler_override_decay_linear` / `_cosine` / `_validation`
- `test_sequential_scheduler_override_decay_linear` / `_cosine`
- `test_sequential_scheduler_override_decay_inside_first_segment` (override interrupts the *first* sub-scheduler)
- `test_sequential_scheduler_override_decay_t_max_warning` (warning fires once and only once)
- `test_composable_and_sequential_match_when_t_max_aligns` (parametrized parity)

43 tests pass; mypy / isort / black / ruff all clean.

## Test plan

- [x] `pytest src/test/optim/scheduler_test.py` — 43 passed
- [x] `mypy src/olmo_core/optim/scheduler.py src/test/optim/scheduler_test.py` — no issues
- [x] `make style-check` (isort + black) — clean
- [x] `make lint-check` (ruff) — clean
- [ ] Confirm `override_decay` on `SequentialScheduler` produces the intended LR trajectory in a real run


TianhuaTao (approved):
Looks good.

PR #674, opened 2026-05-12
Add PowerLR scheduler

AkshitaB:
## Summary

Adds `PowerLR`, a power-law learning rate scheduler. The original implementation is by @TianhuaTao; this PR carries over just the scheduler portion onto `main`, preserves the original commit's authorship/date, and fixes two small bugs from the source commit.

### `PowerLR` (registered as `"power_lr"`)

Three-phase schedule:

1. **Linear warmup** to `initial_lr` over `warmup` steps/tokens (or `warmup_fraction * t_max`).
2. **Power decay**: `lr = initial_lr * (current / warmup) ** b`, where `b` is negative (default `-0.51`). Key property: the LR in this phase depends only on `current / warmup`, **not** on `t_max` — so the schedule is horizon-independent. Useful when the eventual training horizon isn't known at config time, or when you want LR curves to be comparable across runs of different lengths.
3. **Optional linear decay tail** over the last `decay` steps/tokens (or `decay_fraction * t_max`, default `0.1`) that linearly anneals from the power-phase value at the start of the tail down to `decay_min_lr`.

Config fields follow the standard pattern used by the other schedulers in this file: `warmup` / `warmup_fraction` / `warmup_steps` (deprecated alias) for warmup, `decay` / `decay_fraction` / `decay_steps` (deprecated alias) for the tail, plus `warmup_min_lr`, `decay_min_lr`, and `b`. Validation enforces `b < 0` and the usual mutual-exclusivity of `warmup` vs. `warmup_fraction` (and `decay` vs. `decay_fraction`).

## Tests

`src/test/optim/scheduler_test.py` gains:

- `test_power_lr` — exercises all three phases (linear warmup, power-law decay, linear decay tail) and the tail endpoint.
- `test_power_lr_errors` — covers `b >= 0` rejection, missing warmup, out-of-range `warmup_fraction`, and the `decay` + `decay_fraction` mutual-exclusivity error.

34 tests pass; `make checks` clean (isort, black, ruff, mypy across all 381 source files).

## Test plan

- [x] `pytest src/test/optim/scheduler_test.py` — 34 passed
- [x] `make checks` (isort + black + ruff + mypy) — clean


TianhuaTao (approved):
Looks good.

PR #689, opened 2026-05-27
Add profiling callbacks and fix logger init ordering

AkshitaB:
## Summary

A few general training-callback improvements.

### New profiling callbacks (`callbacks/profiler.py`)
- **`NvidiaProfilerCallback`** — wraps a window of training steps (`start`→`end`, on configured ranks) in `cudaProfilerStart/Stop` plus `emit_nvtx` NVTX ranges, for capture under an external Nsight Systems session. (Distinct from the existing `ProfilerCallback`, which uses `torch.profiler` and writes its own Chrome trace; this one drives an external `nsys` capture range and annotates the timeline with PyTorch-op NVTX.)
- **`TorchMemoryHistoryCallback`** — records CUDA memory allocation history and dumps a snapshot pickle (viewable at https://pytorch.org/memory_viz) for OOM/fragmentation debugging.

### Logger init ordering (`callbacks/wandb.py`, `callbacks/comet.py`)
`WandBCallback` and `CometCallback` now declare a higher callback `priority` so they initialize **before** the checkpointer. Pre-train checkpoint saves can flush already-recorded metrics; if the external logger hasn't started yet those metrics are dropped. New test in `callback_order_test.py` covers the ordering and the wandb finalize-on-`close` lifecycle.

### Speed monitor (`callbacks/speed_monitor.py`)
Adds a `throughput/device/TFLOPs_per_GPU` metric and an RTX PRO 6000 entry for peak-FLOPs / MFU estimation.

## Notes
- `NvidiaProfilerCallback` only produces output when the job is launched under `nsys` (e.g. `nsys profile --capture-range=cudaProfilerApi ...`); standalone it just toggles a capture range. Noted in its docstring.
- Stray `print()` debugging in the profiler callbacks was converted to `log` calls.

## Tests
`pytest src/test/train/callbacks/` → 40 passed (incl. the 2 new ordering/lifecycle tests). `make checks` (isort/black/ruff/mypy) clean.


farhatkevin:
@codex review

farhatkevin:
@codex review. Check for profiler windows being off by one and profiler cleanup

PR #690, opened 2026-05-27
Improve checkpoint/S3 IO robustness

AkshitaB:
## Summary

Three small, independent reliability improvements to the IO / distributed-checkpoint paths.

### S3 retries on transient SSL errors (`io.py`)
`_s3_retry_condition` now also treats SSL errors (`ssl.SSLError`, botocore `SSLError`, urllib3 `SSLError`) as retryable, alongside the existing client/HTTP/connection errors. Transient TLS handshake failures during S3 transfers are no longer fatal.

### Distributed checkpoint write correctness (`distributed/checkpoint/filesystem.py`)
Before `torch.save`-ing a tensor, clone it so a **view or storage-sharing tensor** doesn't drag its entire backing storage into the checkpoint shard. A guard raises `OLMoCheckpointError` if a written tensor's byte length exceeds its `nbytes` by more than 1 MiB, surfacing this class of bug loudly instead of silently bloating checkpoints.

### Loader API path (`distributed/checkpoint/__init__.py`)
Call `dist_cp.state_dict_loader.load(...)` instead of the top-level `dist_cp.load` alias.

## Tests
- New unit test for `_s3_retry_condition` covering SSL/connection (retryable) vs. non-network errors (not retryable).
- `pytest src/test/io_test.py` and the CPU `distributed/checkpoint` tests pass. (The `*_to_gcs_*` checkpoint tests require GCS credentials and are environment-gated.)
- `make checks` (isort/black/ruff/mypy) clean.

### DIFF
--- a/src/olmo_core/distributed/checkpoint/__init__.py
+++ b/src/olmo_core/distributed/checkpoint/__init__.py
@@ -177,7 +177,7 @@ def load_state_dict(
     reader = RemoteFileSystemReader(
         dir, thread_count=thread_count, pre_download=pre_download, work_dir=work_dir
     )
-    dist_cp.load(
+    dist_cp.state_dict_loader.load(
         state_dict,
         checkpoint_id=dir,
         storage_reader=reader,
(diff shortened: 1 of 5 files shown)


src/olmo_core/distributed/checkpoint/__init__.py
farhatkevin: we might also want the new load alias here

farhatkevin (approved):
lgtm!

PR #692, opened 2026-05-27
[1/n] Add vision transformer, connector, and MultimodalLM

jason718:
## Summary

This is the first in a series of a few PRs adding VLM/multimodal support to OLMo-core.

---

## What's in this PR

Three new modules under `src/olmo_core/nn/vision/`:

### Vision encoder (`image_vit.py`, `config.py`)

A single, configurable `VisionTransformer` (built from `VisionEncoderConfig`) covering both supported encoder families. There is **one** module class; the architectural variant is selected entirely through config switches rather than separate subclasses:

| Switch | CLIP (`openai`) | SigLIP / SigLIP2 |
|---|---|---|
| `use_cls_token` | `True` (prepends a learnable CLS) | `False` |
| `patch_embedding_bias` | `False` | `True` |
| `use_pre_ln` | `True` (LayerNorm before blocks) | `False` |

| Type | HF checkpoint |
|------|--------------|
| `openai` (CLIP) | `openai/clip-vit-large-patch14-336` |
| `siglip` | `google/siglip-so400m-patch14-384` |
| `siglip2` | `google/siglip2-so400m-patch14-384` |

Key config knobs: `image_default_input_size`, `image_patch_size`, `image_emb_dim`, `image_num_heads`, `image_num_layers`, `image_num_pos`. Factory class-methods cover all standard Molmo2 variants.

The transformer block follows the same pattern as the LM side: `ViTAttention` / `ViTMLP` / `ViTBlock` are public, and the block is selected via `VisionBlockConfig` / `VisionBlockType` (`VisionEncoderConfig.block`), so additional block implementations can be registered as they're added.

### `VisionConnector` (`connector.py`)

Maps variable-length vision embeddings to the LM hidden dimension:
1. 2×2 attention pooling over the patch grid
2. SwiGLU MLP projector → `output_dim = lm.d_model`

Constructed via `VisionConnectorConfig.from_vision_encoder(vis_cfg, output_dim, mlp_hidden_size)`.

### `MultimodalLM` (`multimodal.py`)

Composite model holding `lm: Transformer`, `vision: VisionTransformer`, and `connector: VisionConnector`. It **composes** an LM rather than subclassing `Transformer` — the LM exposes a generic `input_embeddings` entry point (the standard precomputed-embeddings hook), and the multimodal logic lives here. Forward pass:
1. Run the vision encoder on each crop's patch tensor.
2. Project patch embeddings to LM width via `VisionConnector`.
3. Splice patch tokens into the text embedding stream at `image_patch_token_id` positions.
4. Run the standard LM forward pass on the fused sequence.

---

## HF Parity Tests

`src/test/nn/vision/parity_test.py` loads real checkpoints and asserts numerical equivalence between our implementation and HuggingFace's reference models. Tests skip automatically when checkpoints are not cached.

Measured max absolute errors on CPU in float32 (24–27 transformer layers):

| Checkpoint | max abs error | mean abs error | tolerance used |
|---|---|---|---|
| `openai/clip-vit-large-patch14-336` | **4.9e-04** | 3.1e-06 | `atol=3e-3, rtol=1e-3` |
| `google/siglip-so400m-patch14-384` | **2.8e-03** | 7.8e-06 | `atol=3e-3, rtol=1e-3` |
| `google/siglip2-so400m-patch14-384` | **3.1e-04** | 3.0e-06 | `atol=3e-3, rtol=1e-3` |

These errors reflect genuine float32 accumulation across 24–27 attention/MLP layers from independent kernel implementations — not any precision loss. They are ~10–100× tighter than fp16 parity would require. State-dict key layouts are unchanged across the CLIP and SigLIP variants, so all three parity tests pass against the single configurable encoder.

---

## Files changed

```
src/olmo_core/nn/vision/__init__.py        (new)
src/olmo_core/nn/vision/config.py          (new)  VisionEncoderConfig + factories, VisionBlockConfig
src/olmo_core/nn/vision/image_vit.py       (new)  VisionTransformer, ViTBlock/ViTAttention/ViTMLP
src/olmo_core/nn/vision/connector.py       (new)  VisionConnector
src/olmo_core/nn/vision/multimodal.py      (new)  MultimodalLM, MultimodalLMConfig

src/test/nn/vision/__init__.py             (new)
src/test/nn/vision/config_test.py          (new)  config + factory + build dispatch + round-trip
src/test/nn/vision/image_vit_test.py       (new)  encoder forward-pass behavior
src/test/nn/vision/connector_test.py       (new)
src/test/nn/vision/multimodal_test.py      (new)
src/test/nn/vision/parity_test.py          (new)  CLIP + SigLIP + SigLIP2 HF parity
```

### DIFF
--- a/src/olmo_core/nn/transformer/model.py
+++ b/src/olmo_core/nn/transformer/model.py
@@ -506,6 +506,7 @@ class Transformer(nn.Module):
         self,
         input_ids: torch.Tensor,
         *,
+        input_embeddings: Optional[torch.Tensor] = None,
         labels: Optional[torch.Tensor] = None,
         ignore_index: int = -100,
         loss_reduction: Literal["mean", "sum", "none"] = "mean",
@@ -519,6 +520,12 @@ class Transformer(nn.Module):
         Run the transformer on the token input IDs.
 
         :param input_ids: The token input IDs, shape ``(batch_size, seq_len)``.
+        :param input_embeddings: Pre-computed embeddings to use instead of looking up
+            ``input_ids`` in the embedding table, shape
+            ``(batch_size, seq_len, d_model)``.  When provided the embedding lookup,
+            scale, and norm steps are all skipped.  Intended for multimodal use-cases
+            where image features have already been spliced into the embedding sequence.
+            Not supported with context parallelism.
         :param labels: The token labels, shape ``(batch_size, seq_len)``.
         :param ignore_index: The index to ignore in the loss computation. Default is -100.
         :param loss_reduction: The reduction method for the loss. Can be "mean", "sum", or "none".
@@ -550,11 +557,14 @@ class Transformer(nn.Module):
 
         # Get embeddings but pass-through for non-existent layers to allow easy
         # pipeline parallel configuration.
-        h = self.embeddings(input_ids) if self.embeddings is not None else input_ids
-        if self.embeddings is not None and self.embed_scale is not None:
-            h = h * self.embed_scale
-        if self.embedding_norm is not None:
-            h = self.embedding_norm(h)
+        if input_embeddings is not None:
+            h = move_to_device(input_embeddings, self.device)
+        else:
+            h = self.embeddings(input_ids) if self.embeddings is not None else input_ids
+            if self.embeddings is not None and self.embed_scale is not None:
+                h = h * self.embed_scale
+            if self.embedding_norm is not None:
+                h = self.embedding_norm(h)
 
         # Run each block.
         for block_key, block in self.blocks.items():
--- a/src/olmo_core/nn/transformer/model.py
+++ b/src/olmo_core/nn/transformer/model.py
@@ -537,6 +537,13 @@ class Transformer(nn.Module):
 
         :returns: The logits if ``labels`` is ``None`` or the losses if ``labels`` is not ``None``.
         """
+        if input_embeddings is not None and self._cp_load_balancer is not None:
+            raise RuntimeError(
+                "`input_embeddings` is not supported with context parallelism: `_prepare_inputs` "
+                "shards `input_ids`/`labels`/RoPE while `input_embeddings` stays full-size, which "
+                "would misalign the hidden states."
+            )
+
         (
             input_ids,
             labels,
--- a/src/olmo_core/nn/vision/config.py
+++ b/src/olmo_core/nn/vision/config.py
@@ -8,14 +8,14 @@ from olmo_core.config import DType, StrEnum
 from olmo_core.nn.config import ModuleConfig
 
 __all__ = [
-    "VisionBackboneType",
+    "VisionEncoderType",
     "VisionBlockType",
     "VisionBlockConfig",
-    "VisionBackboneConfig",
+    "VisionEncoderConfig",
 ]
 
 
-class VisionBackboneType(StrEnum):
+class VisionEncoderType(StrEnum):
     """
     An enumeration of supported vision encoder architectures.
     """
@@ -31,13 +31,13 @@ class VisionBackboneType(StrEnum):
     """
     SigLIP-style encoder (no CLS token).
     Default config matches SigLIP-SO400M-14-378.
-    ➡️ :class:`~olmo_core.nn.vision.SiglipVisionTransformer`
+    ➡️ :class:`~olmo_core.nn.vision.VisionTransformer`
     """
 
     siglip2 = "siglip2"
     """
     SigLIP2-style encoder (no CLS token; same architecture as SigLIP, improved training).
-    ➡️ :class:`~olmo_core.nn.vision.SiglipVisionTransformer`
+    ➡️ :class:`~olmo_core.nn.vision.VisionTransformer`
     """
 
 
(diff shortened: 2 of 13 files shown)


jason718:
## CI failure analysis

All 3 failing CI jobs are due to the PR coming from an external fork (`jason718/OLMo-core`), which does not receive GitHub Actions secrets. None are caused by the code changes in this PR.

### Test (CPU)
5 tests fail — all with credential errors, not logic errors:

| Test | Error |
|------|-------|
| `test/data/mixes_test.py::test_olmoe_mix` | S3 `403 Forbidden` — no AWS creds |
| `test/data/mixes_test.py::test_dolma17_mix` | S3 `403 Forbidden` — no AWS creds |
| `test/data/mixes_test.py::test_v3_small_ppl_validation_mix` | S3 `403 Forbidden` — no AWS creds |
| `test/io_test.py::test_s3_functionality` | S3 `403 Forbidden` — no AWS creds |
| `test/launch/beaker_test.py::test_get_beaker_client_caching` | `BeakerConfigurationError: token is empty` |

These tests all pass on PRs from within the main repo (e.g. #681) where secrets are available.

### Test checkpoint
Fails with S3 `AccessDenied` on checkpoint read — same root cause (no AWS creds on fork).

### Test olmo3 ladder
Fails with empty Beaker token — same root cause.

These failures are pre-existing on all external fork PRs and do not reflect on the correctness of the vision architecture code being introduced here.

src/olmo_core/nn/vision/image_vit.py
```diff
@@ -0,0 +1,333 @@
+import math
+from typing import List, Optional, Tuple
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from transformers.activations import get_activation
```
undfined: Let's make this a lazy import so we don't introduce a new hard requirement on `transformers`. Right now we would would always raise unless the user includes the `transformers` library in their environment

src/olmo_core/nn/vision/image_vit.py
```diff
@@ -0,0 +1,333 @@
...
+
+class VisionTransformer(nn.Module):
+    """
+    OpenAI CLIP-style vision transformer with a prepended CLS token.
+
+    Matches ``VisionTransformer`` in the Molmo reference implementation and is
+    suitable for weights from OpenAI CLIP ViT-L/14-336.
+
+    The forward pass accepts pre-patchified images and returns the hidden states
+    from every block, so callers can select specific layers (e.g. ``[-2, -9]``).
```
undfined: Given this is intended to be foundational VLM architecture in olmo-core we should probably avoid composing  core doc strings about specific implementations in other projects unless absolutely needed to build intuition.

src/olmo_core/nn/vision/multimodal.py
```diff
@@ -0,0 +1,194 @@
+from dataclasses import dataclass
+from typing import List, Optional, Tuple, Union
+
+import torch
+import torch.nn as nn
+
+from ...config import Config
```
undfined: Generally, I think we should prefer absolute import paths unless an exception warrants it.

src/olmo_core/nn/transformer/model.py
```diff
@@ -519,6 +520,12 @@ def forward(
         Run the transformer on the token input IDs.
 
         :param input_ids: The token input IDs, shape ``(batch_size, seq_len)``.
+        :param input_embeddings: Pre-computed embeddings to use instead of looking up
+            ``input_ids`` in the embedding table, shape
+            ``(batch_size, seq_len, d_model)``.  When provided the embedding lookup,
+            scale, and norm steps are all skipped.  Intended for multimodal use-cases
+            where image features have already been spliced into the embedding sequence.
+            Not supported with context parallelism.
```
undfined, line 528: input_embeddings is documented as not supported with context parallelism, but it isn’t enforced. With CP enabled, _prepare_inputs() shards input_ids/labels/RoPE while input_embeddings stays full-size, so h can be misaligned. We should guard against this until we support it.

src/olmo_core/nn/vision/config.py
```diff
@@ -0,0 +1,317 @@
...
+        """Number of patches along (height, width) for the default input size."""
+        h, w = self.image_default_input_size
+        return h // self.image_patch_size, w // self.image_patch_size
+
+    # ------------------------------------------------------------------
+    # SigLIP factory methods
+    # ------------------------------------------------------------------
+
+    @classmethod
+    def siglip_b16_224(cls) -> "VisionBackboneConfig":
```
undfined: Can we use `Self` for these classmethod factories instead of hard-coding "VisionBackboneConfig"? Since they return cls(...), `Self` preserves subclass return types and avoids repeating the concrete class.

undfined (changes requested):
Looks solid! I like the layout and tests, thank you. A few suggestions from me + let's check the copilot suggestions (a couple overlap). @AkshitaB would love to get your input before we merge.

src/olmo_core/nn/transformer/model.py
```diff
@@ -506,6 +506,7 @@ def forward(
         self,
         input_ids: torch.Tensor,
         *,
+        input_embeddings: Optional[torch.Tensor] = None,
```
AkshitaB, line 509: Is it cleaner to have a subclass for VisionTransformer?

src/olmo_core/nn/vision/image_vit.py
```diff
@@ -0,0 +1,352 @@
...
+        nn.init.trunc_normal_(self.w1.weight, std=math.sqrt(1 / self._emb_dim), a=-2.0, b=2.0)
+        nn.init.trunc_normal_(self.w2.weight, std=math.sqrt(1 / self._mlp_dim), a=-2.0, b=2.0)
+        nn.init.zeros_(self.w1.bias)
+        nn.init.zeros_(self.w2.bias)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        return self.w2(self.act(self.w1(x)))
+
+
+class _ViTBlock(nn.Module):
```
AkshitaB: Typically, for non-vision transformer blocks, we keep it public and the block itself has its own config and there are various subclasses. Is the expectation that for the vision part, it will always be this exact implementation?

AkshitaB (commented):
Class structure and organization looks good to me overall. Added a couple of questions / comments mostly to understand the conventions in multimodal work.

src/olmo_core/nn/transformer/model.py
jason718: `input_embeddings` is not vision-specific — it's the standard precomputed-embeddings entry point; HF impl also has similar input, see `inputs_embeds` here: https://huggingface.co/docs/transformers/en/model_doc/qwen3_5#transformers.Qwen3_5TextModel.forward.inputs_embeds

Additionally, a subclass would either duplicate the forward body or still rely on this same hook, and would also lose  free support for normalized/MoE LMs.

AkshitaB (approved):

undfined (approved):
lgtm, thanks!

PR #693, opened 2026-05-27
[2/n] HF Molmo2 loader and logit parity tests

jason718:
## Summary

2nd PR in the VLM stack building Molmo2-style multimodal support.

- [1/n] Vision architecture (#692)
- **[2/n] This PR — HF Molmo2 loader**

1. **`molmo2_loader.py`** — HF Molmo2 → `MultimodalTransformer` weight converter handling three architecture deltas:
   - **Embedding split**: concatenates HF's `wte.embedding` + `wte.new_embedding` into one flat `embeddings.weight`
   - **Fused QKV + gated MLP**: splits `att_proj` → `w_q`/`w_k`/`w_v` and `ff_proj` → `w1`/`w3`
   - **Patch-embedding flatten order**: permutes spatial-first `(kh,kw,c)` HF weights to channel-first `(c,kh,kw)`

   Also exports two public helpers for transformers 5.x compatibility:
   - `ensure_default_rope_registered()` — re-registers `ROPE_INIT_FUNCTIONS["default"]` before `from_pretrained`
   - `reinit_rope_buffers(model)` — re-runs `rope_init_fn` after loading to fix uninitialised `inv_freq` buffers

2. **Parity tests** (`molmo2_parity_test.py`, `molmo2_logits_parity_test.py`):
   - Vision encoder: max abs diff < 5e-3 in fp32 at every ViT layer the connector reads
   - Embedding: < 1e-3 in bf16
   - LM-only forward and full pipeline (4B and 8B)

3. **Generation test** (`molmo2_generation_test.py`):
   - PIL-based single-crop preprocessor (no torchvision) — resize → SigLIP2 normalize → patchify → 2×2 pool
   - Synthetic RGB gradient image + prompt `"What colors do you see?"` → greedy-decode 8 tokens
   - Asserts our model's token IDs match HF exactly step-by-step

## Parity results (bfloat16, GPU)

| Variant | Embedding diff | LM argmax | LM max diff | Full pipeline argmax | Generation (8 tok) |
|---------|---------------|-----------|-------------|---------------------|-------------------|
| Molmo2-4B | < 1e-3 | ✅ match | **0.125** (thr 1.0) | ✅ match | ✅ exact match |
| Molmo2-8B | < 1e-3 | ✅ match | **0.313** (thr 1.0) | ✅ match | (4B only tested) |
| Molmo2-O-7B | < 1e-3 | — | ⚠️ skipped | ⚠️ skipped | — |

Vision-encoder layer parity (fp32, all 3 variants): max abs diff < 5e-3 ✅

O-7B logit tests skipped: uses per-layer YaRN attention scaling (`attention_factor ≈ 1.208`) on 24/32 LM layers; `MultimodalTransformer` uses a single global RoPE config. Vision-encoder and embedding parity still pass for O-7B.

## Generation QA (Molmo2-4B, bf16)

Prompt: image of a synthetic RGB gradient (red↗ green↗ blue=flat) + `"What colors do you see?"`

```
Question : What colors do you see?
HF output: The colors I see in this image are:
           - Blue
           - Purple
           - Pink
           - Green
Our output: The colors I see in this image are:
            - Blue
            - Purple
            - Pink
            - Green
Token IDs (HF = ours): [785, 7987, 358, 1490, 304, 419, 2168, 525]
```

Both models produce **identical token IDs** for all 8 greedy-decode steps, and the answer is semantically coherent for the image content.

## transformers 5.x RoPE compat

transformers 5.8.1 removed `ROPE_INIT_FUNCTIONS["default"]` and its fast-init (meta-device) leaves non-persistent `inv_freq` buffers uninitialised. The two-step fix lives in `molmo2_loader.py`:

```python
ensure_default_rope_registered()  # before from_pretrained
hf = AutoModelForImageTextToText.from_pretrained(model_id, trust_remote_code=True)
reinit_rope_buffers(hf)            # after from_pretrained
```

## Test command

```bash
# Requires Molmo2 checkpoints in HF cache; auto-skips otherwise
pytest -v src/test/nn/vision/molmo2_parity_test.py \
          src/test/nn/vision/molmo2_logits_parity_test.py \
          src/test/nn/vision/molmo2_generation_test.py
```

---

### DIFF
--- a/src/olmo_core/nn/vision/molmo2_loader.py
+++ b/src/olmo_core/nn/vision/molmo2_loader.py
@@ -1,0 +1,518 @@
+"""
+HuggingFace Molmo2 → :class:`~olmo_core.nn.vision.MultimodalTransformer`
+state-dict converter.
+
+Loads weights from a public Molmo2 checkpoint on HuggingFace (e.g.
+``allenai/Molmo2-O-7B``) into our composite multimodal model.
+
+The two architectures differ in three places that this converter handles:
+
+1. **LM token embedding split.** HF Molmo2 keeps the base vocab and the extra
+   image-special-token vocab in two separate parameters
+   (``transformer.wte.embedding`` and ``transformer.wte.new_embedding``);
+   our :class:`~olmo_core.nn.transformer.Transformer` uses a single
+   ``embeddings.weight`` table. We concatenate.
+
+2. **Fused QKV and gated MLP in the LM.** HF Molmo2 has
+   ``self_attn.att_proj`` (fused Q+K+V) and ``mlp.ff_proj`` (fused gate+up
+   for SwiGLU). Our LM uses ``w_q`` / ``w_k`` / ``w_v`` and ``w1`` / ``w3``.
+   We split along the output dimension.
+
+3. **Patch-embedding flatten order in the vision encoder.** Molmo2's HF
+   ``image_processing_molmo2.batch_pixels_to_patches`` lays patches out
+   spatial-first (``kh, kw, c``), while our
+   :class:`~olmo_core.data.multimodal.ImagePreprocessor` uses channel-first
+   (``c, kh, kw``) to match the HuggingFace ``Conv2d`` patch-embedding
+   convention our CLIP/SigLIP parity tests verified against. We permute
+   the patch-embedding weight to bridge the two.
+
+Vision blocks, the connector pooling cross-attention, the connector SwiGLU
+projector, and the LM head all map one-to-one (no per-tensor surgery).
+
+Usage::
+
+    cfg = MultimodalTransformerConfig(lm=olmo3_7B(...), vision=..., connector=...)
+    model = MultimodalTransformer(cfg)
+    from transformers import AutoModelForCausalLM
+    hf = AutoModelForCausalLM.from_pretrained("allenai/Molmo2-O-7B", trust_remote_code=True)
+    converted = molmo2_hf_state_dict_to_multimodal_transformer(hf.state_dict(), cfg)
+    model.load_state_dict(converted)
+"""
+
+import logging
+from typing import Any, Dict, Optional, Tuple
+
+import torch
+
+from ...config import DType
+
+log = logging.getLogger(__name__)
+from ..transformer import TransformerConfig
+from .config import VisionBackboneConfig, VisionBackboneType
+from .connector import (
+    ImagePoolingType,
+    ImageProjectorType,
+    VisionConnectorConfig,
+)
+from .multimodal import MultimodalTransformerConfig
+
+__all__ = [
+    "molmo2_hf_state_dict_to_multimodal_transformer",
+    "molmo2_config_from_hf_config",
+    "ensure_default_rope_registered",
+    "reinit_rope_buffers",
+]
+
+
+class Molmo2LoaderError(RuntimeError):
+    """Raised when the HF Molmo2 state dict can't be mapped to our model."""
+
+
+def ensure_default_rope_registered() -> None:
+    """Re-register ``ROPE_INIT_FUNCTIONS["default"]`` if upstream removed it.
+
+    transformers ≥ 5 dropped the ``"default"`` key from
+    ``modeling_rope_utils.ROPE_INIT_FUNCTIONS``, but the Molmo2 modeling
+    code bundled inside HF checkpoints still references it.  Call this
+    **before** ``from_pretrained`` so the model can be instantiated.
+    Safe to call repeatedly; a no-op when ``"default"`` is already present.
+    """
+    from transformers.modeling_rope_utils import ROPE_INIT_FUNCTIONS
+
+    if "default" in ROPE_INIT_FUNCTIONS:
+        return
+
+    def _default_rope(config, device=None, seq_len=None, **kwargs):
+        base = getattr(config, "rope_theta", 10000.0)
+        head_dim = getattr(config, "head_dim", None) or (
+            config.hidden_size // config.num_attention_heads
+        )
+        inv_freq = 1.0 / (
+            base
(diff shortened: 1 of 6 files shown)


src/test/nn/vision/molmo2_loader_test.py
```diff
@@ -0,0 +1,325 @@
...
+A separate ``@pytest.mark.slow`` numerical-parity test that requires a real
+HF Molmo2 checkpoint can be added later (task follow-up).
+"""
+
+from typing import Dict
+
+import pytest
+import torch
+
+from olmo_core.data.multimodal import MultimodalTokenizerConfig
```
undfined, line 20: I'm not able to run this:

```
src/test/nn/vision/molmo2_loader_test.py:20: in <module>
    from olmo_core.data.multimodal import MultimodalTokenizerConfig
E   ModuleNotFoundError: No module named 'olmo_core.data.multimodal'
```

src/olmo_core/nn/vision/molmo2_loader.py
```diff
@@ -0,0 +1,518 @@
...
+
+    return out
+
+
+# ---------------------------------------------------------------------------
+# HF Molmo2 config → MultimodalTransformerConfig
+# ---------------------------------------------------------------------------
+
+
+def _build_lm_config(text_cfg, total_vocab_size: int) -> TransformerConfig:
```
undfined, line 362: Let's take a TODO to make this a bit more robust / generic. The growing list of models it will dispatch to make it something that will be difficult to maintain over time. Also, this ignores many HF fields right?

src/olmo_core/nn/vision/molmo2_loader.py
```diff
@@ -0,0 +1,518 @@
...
+    # behaviour matches HF's exactly.
+    hf_lm_head = _require(hf_state_dict, "lm_head.weight")
+    extra_rows = lm_cfg.vocab_size - hf_lm_head.shape[0]
+    if extra_rows < 0:
+        raise Molmo2LoaderError(
+            f"HF lm_head has {hf_lm_head.shape[0]} rows but our LM "
+            f"vocab_size is only {lm_cfg.vocab_size}"
+        )
+    if extra_rows > 0:
+        pad = torch.zeros(extra_rows, hf_lm_head.shape[1], dtype=hf_lm_head.dtype)
```
undfined, line 258: This torch.zeros(...) defaults to CPU, so torch.cat([hf_lm_head, pad], ...) can fail when the HF state dict is already on CUDA. Elsewhere we tend to allocate from the existing tensor for this case, so hf_lm_head.new_zeros((extra_rows, hf_lm_head.shape[1])) would preserve both device and dtype.

src/olmo_core/nn/vision/molmo2_loader.py
```diff
@@ -0,0 +1,518 @@
...
+        matches the HF Molmo2 layout, ready to receive converted weights.
+    """
+    text_cfg = hf_config.text_config
+    vit_cfg = hf_config.vit_config
+    adapter_cfg = hf_config.adapter_config
+
+    # HF Molmo2 keeps the extra image-token embeddings in a separate
+    # parameter (``new_embedding``); after concatenation our model needs the
+    # combined vocab.
+    total_vocab = text_cfg.vocab_size + text_cfg.additional_vocab_size
```
undfined, line 496: Do we need to mask Molmo2’s input-only extra vocab rows at output?

HF Molmo2’s `lm_head` predicts only the base vocab. This pads our `lm_head` with zero rows for the extra image-token IDs. Even if the base logits and argmax match, those finite zero rows change the loss right?

Minimal example:

```python
import torch
import torch.nn.functional as F

hf_logits = torch.tensor([[4.0, 1.5, -0.5, 0.0, 2.0]])
olmo_logits = torch.cat([hf_logits, torch.zeros(1, 3)], dim=-1)
target = torch.tensor([0])

print("HF logits:        ", hf_logits.tolist())
print("OLMo full logits: ", olmo_logits.tolist())
print("HF prediction:    ", int(hf_logits.argmax(-1)))
print("OLMo prediction:  ", int(olmo_logits.argmax(-1)))
print("HF loss:          ", round(F.cross_entropy(hf_logits, target).item(), 4))
print("OLMo padded loss: ", round(F.cross_entropy(olmo_logits, target).item(), 4))
print("loss drift:       ", round((F.cross_entropy(olmo_logits, target) - F.cross_entropy(hf_logits, target)).item(), 4))

"""
Output:

text


HF logits:         [[4.0, 1.5, -0.5, 0.0, 2.0]]
OLMo full logits:  [[4.0, 1.5, -0.5, 0.0, 2.0, 0.0, 0.0, 0.0]]
HF prediction:     0
OLMo prediction:   0
HF loss:           0.2206
OLMo padded loss:  0.2637
delta:        0.0431

Do we need to explicitly mask the extra output IDs to -inf for Molmo2 parity?
"""
```

undfined (changes requested):

jason718:
Superseded by #715, which contains the same Molmo2 loader rebased on top of the `vision` branch (the unified `VisionEncoderConfig`/`MultimodalLM` API) and addresses the review feedback here (device-safe lm_head padding, decoupled loader test, dispatch TODO, extra-vocab loss note). Closing in favor of #715.

jason718:
@undfined All four of your review comments are addressed in the successor PR #715 (rebased onto the `vision` branch) — see the "Review feedback from #693 — status" table in its description. Three are code fixes (decoupled test import, `new_zeros` device-safe padding, dispatch TODO); the extra-vocab logit question is documented in the loader with the `-inf` masking deferred to the loss/forward layer in the training/eval PRs, since a weight converter can't mask logits. If you'd rather see the masking land in #715 itself, happy to discuss there.

PR #700, opened 2026-06-03
Add cuda_extension_utils JIT loader for CUDA/C++ extensions

AkshitaB:
## What this adds

`olmo_core.kernels.cuda_extension_utils` — a generic just-in-time loader (`load_cuda_extension`) for CUDA/C++ extensions built via `torch.utils.cpp_extension`, with:
- **arch/ABI-tagged build directories**, so a cached build isn't reused across an incompatible torch/CUDA/arch toolchain;
- **cross-process build-lock handling**, so concurrent ranks don't clobber each other's builds.

## Import-safety / testing

Pure-Python and import-safe on CPU — the `cpp_extension` import is lazy, so importing the module needs neither a GPU nor a compiler. Includes `cuda_extension_utils_test.py`, a CPU unit test for the pure-Python helpers (env parsing, arch/ABI tag construction, lock handling). The actual extension build is exercised by GPU tests on hardware.

### DIFF
--- a/src/olmo_core/kernels/cuda_extension_utils.py
+++ b/src/olmo_core/kernels/cuda_extension_utils.py
@@ -162,17 +162,19 @@ def _force_rebuild_build_directory(build_directory: str, *, enabled: bool) -> No
     if not enabled:
         return
 
+    from olmo_core.distributed.utils import get_fs_local_rank
+
+    # Wipe once per filesystem: get_fs_local_rank() is global rank 0 on a shared FS
+    # (OLMO_SHARED_FS=1), per-node local rank 0 on node-local FS, and 0 when not
+    # distributed -- so each distinct build cache is wiped exactly once (the default JIT
+    # cache under TORCH_EXTENSIONS_DIR is typically node-local).
+    if get_fs_local_rank() == 0:
+        shutil.rmtree(build_directory, ignore_errors=True)
+        os.makedirs(build_directory, exist_ok=True)
+
     dist = torch.distributed
     if dist.is_available() and dist.is_initialized():
-        rank = dist.get_rank()
-        if rank == 0:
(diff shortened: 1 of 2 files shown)


src/olmo_core/kernels/cuda_extension_utils.py
```diff
@@ -0,0 +1,265 @@
...
+        return
+
+    dist = torch.distributed
+    if dist.is_available() and dist.is_initialized():
+        rank = dist.get_rank()
+        if rank == 0:
+            shutil.rmtree(build_directory, ignore_errors=True)
+            os.makedirs(build_directory, exist_ok=True)
+        dist.barrier()
+        return
```
undfined: Should we do something like the following and rely on OLMO_SHARED_FS to handle whether to clear node local fs vs global shared fs? 

```
if dist.is_available() and dist.is_initialized():
    if get_fs_local_rank() == 0:
        shutil.rmtree(build_directory, ignore_errors=True)
        os.makedirs(build_directory, exist_ok=True)
    dist.barrier()
    return
```

src/olmo_core/kernels/cuda_extension_utils.py
```diff
@@ -0,0 +1,265 @@
...
+
+import logging
+import os
+import re
+import shutil
+import time
+from pathlib import Path
+from typing import Any, Optional, Sequence, Union
+
+import torch
```
undfined, line 20: Is `torch` actually a lazy import if it's at the top level here? Should it be nested in the entrypoint instead?

AkshitaB: Fixed the docstring, it's not torch that's lazily imported but `torch.utils.cpp_extension`

PR #705, opened 2026-06-05
Prasann/landmark fast sparse

PrasannS:
Added some vibe-coded cost for faster landmark attention (should be similar to standard attention with flash attention) as well as sparse compression landmark attention (landmark attention but only landmarks are attended to, this can be up to 20x faster than full attention, possibly more depending on sequence length).

### DIFF
--- a/src/olmo_core/nn/attention/__init__.py
+++ b/src/olmo_core/nn/attention/__init__.py
@@ -71,6 +71,8 @@ __all__ = [
     "FusedAttention",
     "NormalizedAttention",
     "LandmarkAttention",
+    "FastLandmarkAttention",
+    "SparseLandmarkAttention",
     "RingAttentionLoadBalancerType",
     "RingAttentionLoadBalancer",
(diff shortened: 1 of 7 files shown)


src/olmo_core/nn/attention/__init__.py
```diff
@@ -349,6 +376,24 @@ def build(
                 if landmark_use_kernel is not None:
                     kwargs["use_kernel"] = landmark_use_kernel
                 return LandmarkAttention(mem_freq=mem_freq, **kwargs)
+            elif self.name == "fast_landmark":
+                if "window_size" in kwargs:
```
abertsch72, line 380: is it intentional that this only warns for fast and sparse landmark? seems like it should warn for all three (?)

abertsch72 (approved):
thanks for doing this! I didn't do a close readthrough of the kernels, but I'm going to go ahead and merge it and test it out on beaker

PR #716, opened 2026-06-09
Add async all-to-all to ops.moe

AkshitaB:
## Summary

Adds an asynchronous all-to-all to `olmo_core.ops.moe` for overlapping the collective with independent compute:

- `AllToAllAsyncOp` / `all_to_all_async(...)` — launches `dist.all_to_all_single(..., async_op=True)` and returns `(x, y, handle)` immediately, where `y` is the output buffer (not yet valid) and `handle` is the in-flight work.
- `AllToAllWaitOp` / `all_to_all_wait(x, y, handle)` — blocks on the handle and returns the completed `y`.

```python
x, y, handle = all_to_all_async(x, ...)
z = some_independent_compute(...)   # overlaps with the transfer
y = all_to_all_wait(x, y, handle)
```

The work handle and the split/shape metadata are stashed on the output tensor (`_a2a_*`) so they thread through the matching wait and the backward pass without extra `ctx` plumbing. The backward mirrors the forward: it launches the reverse all-to-all in the wait's backward and synchronizes it in the async op's backward, so the gradient path overlaps too.

The existing synchronous `all_to_all` / `AllToAllOp` is unchanged.

## Tests

`test_all_to_all_async_matches_sync` — a 2-GPU NCCL test asserting the async path produces the same output **and** input gradients as the synchronous `all_to_all` on identical inputs. Gated with `@requires_multi_gpu`.

### DIFF
--- a/src/olmo_core/ops/moe.py
+++ b/src/olmo_core/ops/moe.py
@@ -289,6 +289,30 @@ class AllToAllAsyncOp(torch.autograd.Function):
     later blocks on the handle. The work handle and the metadata needed by the wait /
     backward passes are stashed as attributes on the output tensor (``_a2a_*``) so they
     travel through autograd without extra ``ctx`` plumbing across the two ops.
+
+    The backward mirrors the forward, so the gradient all-to-all overlaps too:
+    ``AllToAllWaitOp.backward`` launches the reverse all-to-all and ``AllToAllAsyncOp.backward``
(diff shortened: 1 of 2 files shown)


farhatkevin:
@codex review

AkshitaB:
Superseded: the async all-to-all (all_to_all_async / all_to_all_wait) landed on moe-v2-core via #743, which carried these commits. Closing as redundant.

PR #718, opened 2026-06-11
Add token permute/unpermute helpers

AkshitaB:
Adds `nn/moe/utils.py`: the autograd-aware, `torch.compile`-safe wrapper layer over the MoE token permute / unpermute / chunk-reorder CUDA kernels (with optional Triton and TransformerEngine backends). These move tokens between the router's assignment order and the experts' grouped (per-expert-contiguous) order, in both forward and backward.

Contents:
- **Permute / unpermute with custom autograd** — including fused capacity-drop and row-id-map variants (`moe_permute_1d_fused_drop_no_compile`, `moe_unpermute_1d_fused_drop_no_compile`, `moe_reorder_rows_permutation`).
- **Chunk reordering** for the grouped-expert layout, with TE / CUDA backend selection (`moe_chunk_reorder_no_compile`, `build_chunk_te_routing_map`).
- Small `@torch.compiler.disable` stream/event sync helpers and eager kernel passthroughs.

It depends only on the permute/reorder kernels already in `olmo_core.kernels` plus `olmo_core.utils.get_or_init_stream`; Triton and TransformerEngine are optional, guarded imports, so the module imports on CPU.

Also includes a CPU test for the chunk-reorder torch reference paths (`_chunk_unpermute_torch` zeroing out-of-range rows; `_dispatch_permute_by_row_id_map` honoring an `out=` buffer). The fused permute/unpermute correctness tests are GPU + TransformerEngine gated and will follow.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -182,7 +182,7 @@ jobs:
               # silently skips. Fail loudly if CUDA is still unavailable.
               pip install --force-reinstall --index-url https://download.pytorch.org/whl/cu128 \
                 torch==2.9.1 torchvision torchaudio
-              python -c "import torch; assert torch.cuda.is_available(), 'CUDA unavailable after cu128 reinstall'"
+              python -c 'import torch; assert torch.cuda.is_available()'
               pytest -v --color=yes --durations=3 -m gpu \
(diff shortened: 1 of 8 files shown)


farhatkevin:
@codex review

PR #719, opened 2026-06-11
Add rowwise-FP8 expert compute

AkshitaB:
Adds `nn/moe/v2/fp8.py`: rowwise FP8 quantization/compute for the MoE-v2 experts, built on the scaled-grouped-mm FP8 kernels (`scaled_grouped_mm_q`, `scaled_grouped_mm_q_fp8_weight`, `prequantize_scaled_grouped_mm_rhs`).

Contents:
- `MoERowwiseFP8Config` / `MoERowwiseFP8ScaleMode` — configuration for the rowwise-FP8 path (marked `@beta_feature`).
- FP8 weight-cache management — prequantize expert weights and refresh / invalidate the per-block caches.
- The rowwise-FP8 shared-experts forward path.

It depends only on `olmo_core.config` and the scaled-grouped-mm FP8 kernels already in `olmo_core.kernels`; `nvtx` is an optional import with the `olmo_core._nvtx` no-op fallback, and the only reference to the MoE block is a `TYPE_CHECKING` annotation, so the module imports on CPU without the rest of the MoE-v2 package.


AkshitaB:
Folding fp8.py back into the MoE-v2 model PR. It imports standalone but is functionally block machinery (its functions operate on MoEFusedV2TransformerBlock internals), so it isn't meaningfully reviewable/testable on its own — it belongs with the block.

PR #723, opened 2026-06-12
Compare Qwen3.5 GPU test against the real official checkpoint

mnoukhov:
The Qwen3.5 GPU comparison test previously built a synthetic random-weight `Qwen3_5TextConfig`. This restores a real-checkpoint comparison: it loads the official `Qwen/Qwen3.5-0.8B` weights and tokenizer and compares logits on a real prompt.

The test just calls `from_pretrained("Qwen/Qwen3.5-0.8B")` as-is (skipping if the checkpoint can't be obtained). To keep the GPU job from downloading the gated checkpoint in-job (which can exceed the 10-minute task timeout), the gantry run mounts Weka (`oe-training-default`) and points `HF_HOME` at a checkpoint cache there, so the weights are read from the local mount.

**Required data (already done)** populate the cache once from a Weka-mounted machine (with `HF_TOKEN`):
```
HF_HOME=/weka/oe-training-default/test-fixtures/hf hf download Qwen/Qwen3.5-0.8B
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -238,6 +238,10 @@ jobs:
           echo "COMMIT_SHA=$GITHUB_SHA" >> $GITHUB_ENV
           echo "BRANCH_NAME=${{ github.ref_name }}" >> $GITHUB_ENV
 
+      # The '--weka' mount + 'HF_HOME' below point HuggingFace at a checkpoint cache on Weka so
+      # the Qwen3.5 comparison test reads the official weights from the mount instead of
+      # downloading them and timing out. Populate the cache once from a Weka-mounted machine via:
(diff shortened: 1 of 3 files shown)


AkshitaB (commented):
Thanks @mnoukhov ! For this particular test, is it better to just have a smaller test config rather than load the hf model?

mnoukhov:
I'm seeing this qwen 3.5 test as GPU version of `test_logprobs_match_after_roundtrip` which is CPU only. This feels more realistic since we're usually checking those logprobs on GPU

So I think it makes sense to do the GPU test with a real model instead of a tiny test config model

farhatkevin:
@codex review

PR #724, opened 2026-06-12
Add v2 shared experts

AkshitaB:
Adds `nn/moe/v2/shared_experts.py`: `SharedExperts` and `SharedExpertsConfig`.

Shared experts act like a feed-forward applied to every token, but support more than one expert. The up/gate projections of all experts are column-packed into one weight so the first projection is a single GEMM `(BS, D) @ (D, E*2H)`; the result is split per-expert into up/gate, run through a SwiGLU nonlinearity, and the per-expert down projection is a grouped (batched) GEMM `(E, BS, H) x (E, H, D) -> (E, BS, D)`. A split `forward1`/`forward2` variant is provided so the projection and the down-GEMM can overlap with expert-parallel communication.

Depends only on `olmo_core.config`; `nvtx` is an optional import with the `olmo_core._nvtx` no-op fallback. Includes CPU unit tests covering num_params, the forward output shape, the single-expert case, and that the split forward matches the fused one.

### DIFF
--- a/src/olmo_core/nn/moe/v2/shared_experts.py
+++ b/src/olmo_core/nn/moe/v2/shared_experts.py
@@ -1,4 +1,3 @@
-import math
 from dataclasses import dataclass
 
 import torch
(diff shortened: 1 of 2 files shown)


farhatkevin:
@codex review

PR #732, opened 2026-06-17
CI: cache HuggingFace models for the CPU Test job

mnoukhov:
The CPU `Test` CI job runs the HF roundtrip tests, which download Qwen3-0.6B and Gemma-3-270m. Persist `HF_HOME` across runs with `actions/cache` so they're only downloaded once.


farhatkevin (approved):
looks good!

PR #738, opened 2026-06-30
Add expert-parallel no-sync common helpers

AkshitaB:
Adds the shared helpers used by the no-sync expert-parallel forward paths.

## What's added
`nn/moe/v2/ep_no_sync_common.py`:
- **`build_keep_reorder`** — builds the per-expert capacity-drop reorder: stably partitions routed rows into kept-then-dropped, returning the permutation, its inverse, and a keep-mask.
- **`sync_tail_drop_allowed_splits_single_a2a`** — agrees a global tail-drop policy across expert-parallel ranks with a single all-gather of per-rank requested token counts.
- **`restore_drop_unpermute_1d`** — inverts the capacity drop (scatters kept expert outputs back, zero-filling dropped rows) and unpermutes + combines for the 1D no-sync path.

These operate on plain tensors / a passed-in block; the EP forward families that consume them land in follow-up changes.

## Notes
- The optional `nvtx` import uses the `_nvtx` fallback so the module imports without the profiling extra.
- The dead env-gated debug helpers were trimmed to a single labeled hook + an `OLMO_TBO_VERBOSE_DEBUG_PRINT` note.

## Testing
`src/test/nn/moe/v2/ep_no_sync_common_test.py` (CPU): validates `build_keep_reorder` against an independent reference (partial-drop, no-drop, full-drop-some-experts) and that the reorder/inverse permutations invert each other. `make checks` clean.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -301,6 +301,7 @@ jobs:
             --host-networking \
             --gpu-type h100 \
             --gpu-type a100 \
+            --gpu-type b200 \
             --system-python \
             --dataset-secret 'GOOGLE_CREDENTIALS:/.google/credentials.json' \
             --env 'GOOGLE_APPLICATION_CREDENTIALS=/.google/credentials.json' \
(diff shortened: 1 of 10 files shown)


AkshitaB:
@codex review

PR #746, opened 2026-07-06
Add MultiGroupDistributedDataParallel

AkshitaB:
Adds `MultiGroupDistributedDataParallel` under `olmo_core.nn.parallel` (targeting the MoE
integration branch). Same code as #688, re-based onto this branch.

A data-parallel module wrapper that all-reduces gradients across one or more process groups. Unlike
`torch.nn.parallel.DistributedDataParallel`, **different parameters can be reduced over different
process groups** via a `param_process_group_fn(name, param) -> process_group` mapping — needed when
subsets of the model are replicated over different device meshes (e.g. expert weights under expert
parallelism reduce only over ranks holding the same experts).

- **Bucket views**: gradients accumulate into flat pre-allocated per-group buffers; each param's
  `.grad` is a view into its bucket slice, enabling one all-reduce per bucket.
- **Overlap**: bucketed all-reduces launch during backward as each bucket fills;
  `finalize_grad_reduce()` waits for completion after `loss.backward()`.
- **Optional fp32** gradient accumulation and/or reduction.
- Opt-in, duck-typed hooks (`getattr(module, ...)`) for syncing fp8/mxfp8 gradient stores — no-ops
  for standard modules, so this stays a standalone primitive with no MoE dependency.

This is the DDP wrapper the fused-MoE optimizer's `build()` and the MoE train module require.

**Tests** (`src/test/nn/parallel/distributed_test.py`, gloo + nccl): `test_grad_parity` (DDP grads
match a manual all-reduce reference) and `test_no_sync_accumulation` (`no_sync()` defers the reduce
and accumulates across micro-batches). Pass on gloo (CPU); `make checks` clean.

Also pins `nvidia-cutlass-dsl<4.6.0` (separate commit): flash-attn-4's `cute` path pulls
`quack-kernels`, which imports `cutlass.cute.core.ThrMma` — removed in cutlass-dsl 4.6.0 — breaking
`import olmo_core.nn` at CI collection. This is a general CI fix; drop it wherever it lands second.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -238,19 +238,10 @@ jobs:
           # (src/test/nn/moe/v2) also run here since they need TransformerEngine + the same
           # custom kernels, which the torch 2.9 image lacks.
           - name: Test MoE kernels (GPU, torch 2.10)
-            image: tianhuat/olmo-core-torch210-2404-cu128
+            image: akshitab/olmo-core-tch2100cu130-2026-07-03 #tianhuat/olmo-core-torch210-2404-cu128
             gpus: 2
-            # TEMP: this image's torch is a CUDA-13 build the Beaker CUDA 12.8 driver cannot
-            # initialize, so the kernel tests silently skip. Reinstall the cu128 torch 2.10
-            # build (still provides torch.nn.functional.grouped_mm) to match the driver. The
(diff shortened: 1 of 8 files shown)


TianhuaTao (approved):
It looks like a critical correctness issue. The fix is necessary.

PR #760, opened 2026-07-10
Introduce ExpertParallelConfig and refactor the DDP block onto it

AkshitaB:
Replaces the flat `ep_no_sync_*` flag bag on the fused MoE-v2 DDP transformer block with a single `ExpertParallelConfig` (`nn/moe/v2/ep_config.py`).

## What changed
- **`ExpertParallelConfig` / `ExpertParallelPath` / `ExpertParallelSchedule` / `DeepEPConfig`** — a self-contained config type. `ExpertParallelPath` enumerates the EP families (`sync_1d`, `no_sync_1d`, `rowwise_nvshmem`, and the not-yet-wired `rowwise_wave` / `deepep_v2`); mode selection and per-mode tuning live in one place, with validation in `ExpertParallelConfig.validate()`.
- **`OLMoDDPTransformerBlock[Config]`** now takes a single `ep: ExpertParallelConfig` instead of ~13 flat `ep_no_sync_*` fields. Mode is read through `ep.no_sync` / `ep.is_rowwise` properties and tuning through `ep.<field>`.
- **Consumers** (`model.py`, `ep_no_sync_buffers`, `ep_no_sync_tbo_rowwise`) read config via `block.ep.<field>`.
- **Tests** build blocks with `ep=ExpertParallelConfig(path=...)`; rowwise mutations set `ep.path` (`is_rowwise` is now a read-only property).

The experimental `rowwise_wave` / `deepep_v2` paths are declared in the enum but intentionally left unwired here. Removed `test_v2_ep_no_sync_2d_all_to_all_rejected`, which covered the already-removed 2D all-to-all flag.

### DIFF
--- a/src/olmo_core/nn/ddp/block.py
+++ b/src/olmo_core/nn/ddp/block.py
@@ -33,6 +33,7 @@ from olmo_core.nn.moe.v2.checkpointing import (
 # activation-debug helper are imported lazily at their dispatch sites below, so this module
 # imports cleanly without those files present (they ship in later, stacked PRs). See the
 # combined_forward_ep_no_sync_* wrapper methods and the rowwise metric/symm helper sites.
+from olmo_core.nn.moe.v2.ep_config import ExpertParallelConfig
 from olmo_core.nn.moe.v2.fp8 import MoERowwiseFP8Config
 from olmo_core.nn.moe.v2.fp8 import (
     invalidate_rowwise_fp8_cache as _invalidate_rowwise_fp8_cache,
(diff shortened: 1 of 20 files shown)


AkshitaB:
Fixed in the latest commit. The no-sync forward modules (`ep_no_sync_1d`/`ep_no_sync_rowwise`/`ep_no_sync_tbo_1d`/`ep_no_sync_tbo_rowwise`/`ep_no_sync_common`) now read config via `block.ep.<field>` instead of the removed flat `self.ep_no_sync_*` attributes, and the dead 2D-all-to-all guard (whose flag no longer exists) was dropped. Full type-check + no-sync test collection pass.

PR #792, opened 2026-07-27
Add GatedDeltaNet2 (GDN2) 

CalebOuellette:
WIP no need to review.

Adding GDN2 for testing in dense ladder.

Port of fla/layers/gdn2.py (pinned commit cbb0a72e) in the existing GatedDeltaNet style: channel-wise erase (w_b) and write (w_w) gates, low-rank channel-wise decay (w_f) and output gate (w_g), A_log per QK head, dt_bias per key channel, sigmoid-gated output norm. Registered as "gated_delta_net_2" with num_params/init_weights/FLOPs and Ulysses CP support. Repins the fla extra to the gdn2 commit (linux only).


pyproject.toml
```diff
@@ -69,7 +69,7 @@ eval = [
     "google-cloud-storage",  # for cached_path gs:// downloads of the eval data archive
 ]
 fa4 = ["flash-attn-4>=4.0.0b4 ; sys_platform == 'linux'"]
-fla = ["flash-linear-attention==0.4.1"]
+fla = ["flash-linear-attention[cuda] @ git+https://github.com/fla-org/flash-linear-attention.git@cbb0a72efb55c18ca0ef4f298298317573ad2cb3 ; sys_platform == 'linux'"]
```
CalebOuellette: TODO: Should set to 5.2 when 5.2 is released.

src/olmo_core/nn/attention/recurrent.py
```diff
@@ -451,3 +452,474 @@ def build(
...
+
+    def __init__(
+        self,
+        *,
+        d_model: int,
+        n_heads: int,
+        n_v_heads: int | None = None,
+        head_dim: int | None = None,
+        expand_v: float = 1.0,
+        allow_neg_eigval: bool = False,
```
CalebOuellette: Note: Defaults are different from GDN 1. expand_v is 1 instead of 2. allow neg eigval is False.
```
        expand_v: float = 2.0,
        allow_neg_eigval: bool = True,
```

PR #802, opened 2026-07-31
Update anonymous paths in public scripts to point to data in hugging face bucket

warmbowski:
Screenshot 2026-07-31 at 9 39 15 AM

Two large checkpoint folders were copied to a Hugging Face bucket in the allenai org:
https://huggingface.co/buckets/allenai/ai2-llm/tree/checkpoints

This PR updates the public scripts to point to the new HF bucket. Paths are similar, but a "resolve" namespace is added after the bucket name in order to differentiate between bucket ui (tree) and object data (resolve). The path format is https://huggingface.co/buckets/{org name}/{bucket name}/resolve/{path prefix}

The buckets supports anonymous http, s3 api access, and also has hf specific tools to access data. 
https://huggingface.co/docs/huggingface_hub/en/guides/buckets

This pr targets anonymous access only, let me know if you want the s3 config added as an io.py schema "hfs3", with typical s3 type access. That seems simple to do, but my assumption was that internally we'll we will still be working against the gcs bucket and releasing to hf later (correct me if I am wrong). 

relates to: https://github.com/allenai/playground-issues-repo/issues/1143
closes: https://github.com/allenai/playground-issues-repo/issues/1163

### DIFF
--- a/src/scripts/official/OLMo3/OLMo-3-1025-32B-long-context.py
+++ b/src/scripts/official/OLMo3/OLMo-3-1025-32B-long-context.py
@@ -110,7 +110,7 @@ def build_config(opts: argparse.Namespace, overrides: List[str]) -> ExperimentCo
         TrainerConfig(
             save_folder=opts.save_folder,
             save_overwrite=True,
-            load_path="https://storage.googleapis.com/ai2-llm/checkpoints/stego32-highlr-filter3/step656000/",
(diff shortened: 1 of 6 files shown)


BrownianNotion:
Thank you for sharing these! Do you have a suggestion for which checkpoint(s) would be best to start at for long-context training? Is any model souping needed?

AkshitaB (approved):

PR #804, opened 2026-08-01
Put a C compiler in the research image, because torch.compile needs one

philote-dev:
`src/examples/llm/train.py` sets `compile_model=True`, and the registered base for the eduLLM research image is a bare python image with no toolchain. So a training run reaches the first compiled region and dies:

```
InductorError: RuntimeError: Failed to find C compiler. Please specify via CC
environment variable or set triton.knobs.build.impl.
```

**Found the expensive way.** A twelve-hour submission reached an A10G, pulled the three-gigabyte image, built the model, wrote its first checkpoint object to S3, and died on the first compiled region. Not at build time, not in a check, and not before the instance was paid for.

## Choices in it

`gcc` and `g++` rather than `build-essential`. Inductor and Triton shell out to a compiler to build the kernels they generate; they need neither `make` nor `dpkg-dev`, and this image is what an untrusted training command runs inside, so the smaller toolchain is both lighter and less to reason about. `g++` is there because Inductor's CPU path emits C++.

No version pin, unlike torch above it. The torch pin exists because a resolved wheel changes what the model computes; a compiler that builds a kernel and ships nowhere does not, and pinning to a Debian point release would expire on the base's next security update.

The layer sits above `COPY . .` so a source change does not reinstall it, and it ends in `cc --version` so a build fails rather than a run if the compiler somehow is not on PATH.

## What was verified and what was not

The base digest resolves to **Debian trixie carrying python 3.12.13**, read out of the image config on Docker Hub rather than inferred from the repository name — so `apt-get` is the right tool. An Alpine base would need `apk` and would not have manylinux wheels for torch either, so a base swap that broke this line would already have broken the install above it.

**Not built locally**, because there is no Docker daemon on the machine this was written on. The build is the verification.

## The alternative, and why not

The eduLLM platform's getting-started guide currently tells researchers to pass `train_module.compile_model=false`. That works and costs throughput on every run forever. Fixing the image removes the workaround instead of documenting it.


philote-dev:
Opened against the wrong remote — this belongs on our fork, not upstream. Closing.

PR #805, opened 2026-08-01
Say which stage a run died in, in the one field anyone can read

philote-dev:
## Why

A container that fails before training writes its explanation to a CloudWatch stream nobody on the platform side may read. What is readable is Batch's `exitCode`, and every failure here reported `1`.

On 2026-08-01 four runs died between five and forty seconds with exit 1 — which is equally what a bad hyperparameter, a lost quote, a missing entry point, an unreadable bucket, and a wrong argument look like. Each was diagnosed by resubmitting with one thing changed to see whether the number moved, at an A10G and an image pull per guess.

## What this adds

- A `Stage` enum: one exit code per stage that can fail before a loss is computed (64–72, clear of the shell's 126/127/128+n).
- The read stage tells **denied** from **absent**. A missing `s3:GetObject` and a registry entry pointing at an unpublished prefix both arrive as a failed read, and one is a policy change while the other is a dataset that is not there.
- The explanation also goes to W&B, opening a run of its own when the crash happens before the trainer's callback exists. Tagged `died-before-training` so it sorts away from real runs.
- The reporter swallows its own failures — a diagnostic that replaces the error it reports is worse than none.

## Test plan

- [x] `pytest src/test/edullm_train_on_corpus_test.py` — 17 passed
- [x] `ruff check`, `mypy` clean


philote-dev:
Opened against the wrong repository by mistake — this belongs on our fork, not upstream. Closing.

PR #806, opened 2026-08-03
Vision donovan stage 2 training

donovanclay:


uwGZQ:
Pushed 5 stacked commits (`2f86a76de..17b441a07`) on top of this PR:

1. **`dd0a8b7` vocab/tying fix** — HF Molmo2 predicts base vocab only (4B ties the head to the base embedding table). New `output_vocab_size` masks the 128 extra logit columns to `finfo.min` (exact softmax/gradient parity with mm_olmo, works with `response_logits_only`); the loader emits one consistent tensor for tied models (zero-padding used to clobber the new-embedding input rows on load); `retie_word_embeddings()` restores the tie that `Module.to_empty` silently breaks (wired into both Stage-1/2 init paths). Molmo2-4B GPU parity (embedding/LM/pipeline/training-loss) passes.
2. **`e21e2ca` 2D-knapsack packer** — port of mm_olmo `dynamic_packer.py` (buffer 48, token+crop capacities, value = tokens + 30·crops). Text-only Tulu now packs together with image examples like mm_olmo. `PACK_MAX_CROPS` 8 → 125 (the old cap was below one max-tiling example, forcing ~1 image example per 16k pack).
3. **`8198b1d` multi-image** — `encode_sft_example` takes image lists (`Image {i+1}` prefixes, cumulative pooled offsets; list-of-one is bit-identical to single-image). Adds the all-v9 multi-image sources: 3 Mantis, `correction_qa_multi_only_max5`, 6 `cosyn_multidoc_*_exp`, `pixmo_multi_points` (incl. the `format_multi_points` port, `p_multi_point_all_image=0.5`). Mixture: 32 → 43 sources. No attention change needed — HF's `token_type_ids` mask is already cross-image bidirectional.
4. **`0a82089` data-consistency + pipeline fixes** — pointing prompt RNG parity (50% label-lowercase draw), per-index `pixmo_clocks` augmentation, Tulu qwen3 turn layout, `{"messages"}` conversations as one sequential branch, DEMO_STYLES/MC/VQA-answer/format_points parity, mm_olmo per-example rng (`seed*195172+index`) everywhere + the parity harness aligned so `compare_sft_example.py` can verify any index (not just 0), `root_size_factor<1`, deterministic prefetch ordering, non-finite-loss flag now an unconditional per-microbatch collective (was a guaranteed NCCL hang), residual dropout 0.1.
5. **`17b441a` FineVision + MMFineReason** — six optional extra SFT sources (off by default via `MMFINEREASON_RATE`/`FINEVISION_RATES` in Molmo2-Stage2.py), routed through the shared stage-2 encode path. Also fixes `dataset_compat`'s global `List→Sequence` alias corrupting list-of-struct schemas on `datasets>=5`.

All 145 data-layer CPU tests + 10 Molmo2-4B GPU parity tests pass. Full commit messages carry the details and mm_olmo line references.

PR #808, opened 2026-08-06
add partial ScaledSoftmax support

YashasSamaga:
Adds an optional Scalable-Softmax (SSMax) path to `Attention`, behind
`AttentionConfig.scalable_softmax`.

## What it does

Standard attention computes `softmax(q·kᵀ/√d)`. As the number of visible keys grows the logit
distribution flattens and attention fades toward uniform, which is one of the mechanisms behind
degradation at long context. SSMax ("Scalable-Softmax Is Superior for Attention", Nakanishi 2025)
counteracts it by scaling the logits with the visible-context size:

    softmax(s · log(n) · q·kᵀ/√d)

where `n` is the number of keys the query can see and `s` is learned.

It is implemented as a scale applied to `q`, not to the logits. That is algebraically identical and
keeps the change entirely outside the attention kernel, so every backend (flash_2, flash_3, torch
SDPA) picks it up without a custom path.

`ssmax_scale` is one learnable scalar per query head — `n_heads` extra parameters per layer, which
`AttentionConfig.num_params` now accounts for — initialised to ones in `init_weights`.

## `n` is per-position and document-aware

`n` is the number of tokens a given query actually attends to, not the sequence length:

- causal, unpacked: `n = position + 1`;
- packed sequences (`cu_doc_lens` set): `n` restarts at each document boundary, so intra-document
  masking is respected. A token three positions into the second packed document gets `n = 3`, not
  its offset into the flattened batch.

The first token of each document has `log(1) = 0` and therefore a zeroed query. That is a no-op: it
attends to exactly one key, and softmax over a single logit is 1 regardless of its value.

The scale is applied after QK-norm and RoPE, immediately before `sdpa`.

## What is and isn't supported

This is deliberately partial. Three configurations are rejected rather than silently computing the
wrong `n`:

- **Sliding-window attention** — `OLMoConfigurationError`, raised both from `AttentionConfig.build`
  and from the `Attention` constructor. Under SWA a query sees at most `window` keys, so the
  position-based count computed here is simply not the visible count.
- **Context parallelism** — `NotImplementedError`. Under CP the local `q` shard does not carry
  global position offsets, so `arange` over the local sequence gives the wrong `n`. This is the
  remaining work.
- **KV caching** — `NotImplementedError`. During incremental decode `q.shape[1]` is 1, so the
  position count would collapse to `n = 1` instead of the true cached length.

Tensor parallelism is supported: `apply_tp` re-registers `ssmax_scale` as a DTensor sharded on the
head dimension, matching the head-sharded `q`, and the forward pass converts back to local before
the multiply.

## Tests

Six new test functions (eight cases with parametrisation) in `src/test/nn/attention/attention_test.py`:

- `test_scalable_softmax_query_scaling` — exact per-head scaling for both the plain causal case and
  a packed-document case with `cu_doc_lens=[0, 2, 4, 7, 8]`, plus that gradients reach `ssmax_scale`;
- `test_scalable_softmax_disabled_is_identity` — the disabled path returns the input tensor itself;
- `test_scalable_softmax_scale_is_initialized_to_one` — `init_weights` resets the scale;
- `test_scalable_softmax_is_applied_after_qk_norm` — captures `q` at the `sdpa` boundary to pin
  the ordering against QK-norm;
- `test_scalable_softmax_rejects_unsupported_runtime_modes` — KV cache and CP;
- `test_scalable_softmax_rejects_sliding_window_attention` — config-build rejection.

`test_attention_builder_config` gains a `scalable-softmax` case, which pins the `num_params`
accounting, and `test_tensor_parallel_attention` gains one combining `scalable_softmax` with
head-wise QK-norm.

### DIFF
--- a/src/olmo_core/nn/attention/__init__.py
+++ b/src/olmo_core/nn/attention/__init__.py
@@ -7,7 +7,7 @@ from typing import TYPE_CHECKING, List, Optional, Tuple, Union
 import torch
 import torch.nn as nn
 from torch.distributed import DeviceMesh
-from torch.distributed.tensor import Placement, Replicate, Shard
+from torch.distributed.tensor import DTensor, Placement, Replicate, Shard, distribute_tensor
 from torch.distributed.tensor.parallel import parallelize_module
 
 from olmo_core.config import Config, DType, StrEnum
@@ -201,6 +201,7 @@ class AttentionConfig(SequenceMixerConfig["SequenceMixer"]):
     dtype: DType = DType.float32
     sliding_window: Optional[SlidingWindowAttentionConfig] = None
     use_head_qk_norm: Optional[bool] = None
+    scalable_softmax: bool = False
 
     def num_params(self, d_model: int) -> int:
         """
@@ -254,6 +255,9 @@ class AttentionConfig(SequenceMixerConfig["SequenceMixer"]):
             params += n_heads * head_dim
             params += n_kv_heads * head_dim
 
+        if self.scalable_softmax:
+            params += n_heads
+
         return params
 
     def build(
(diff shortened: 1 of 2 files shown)


YashasSamaga:
It sounds like we might drop FusedAttention and redo the PR for FusedAttentionV2.

PR #809, opened 2026-08-06
Point the agent documentation at this platform rather than at AI2's cluster

philote-dev:
## What was wrong

Thirty-five researchers onboard tomorrow and most of them drive a coding agent. A researcher's first hour is in this repository, and two things here sent it wrong.

**`.claude/skills/training-smoke-test/SKILL.md` matches on "launch a smoke test"** — the exact phrase somebody will use — and sent the agent to Beaker at `ai2/jupiter`, `gantry follow`, and a priority ladder capped by `ai2/OLMo_3`. None of that is reachable from this organization. The agent follows it confidently and fails in a way that looks like the platform is broken.

**`CLAUDE.md` never mentioned `edullm` at all.** Its only install line was `pip install -e '.[all]'`, so an agent reading this repository top to bottom came away believing GPU work runs locally or on Beaker. Verified by cold read: a fresh `git clone --depth 1`, no `edullm` on `PATH`, isolated tool directory, following the documents with no other knowledge.

## What changed

`CLAUDE.md` gains a **Running on GPUs** section: the install line, the `check` → `submit` → `status` loop, read stdout on its own, branch on the exit code, match on refusal `code` and never on prose, never call AWS, never quote a price from this file.

It also says the one thing only this repository knows, which no platform-side guard can: **`.edullm/train_on_corpus.py` builds its config in bfloat16, the platform's precision guard reads command text and cannot see a dtype set in code, and a T4 has none in the hardware.** `torch.cuda.is_bf16_supported()` returns true there. Writing the dtype into the command turns a billed failure into a free refusal.

The smoke-test skill keeps everything about *writing* a smoke script — the 190M base config, CP/TP/backend/eval patterns, the naming convention, all of it still correct — and its launch half now points at `edullm`, including the instruction to name the dtype and to report the cost and approval class before submitting.

**`AGENTS.md` is added.** Cursor and Codex read `AGENTS.md` and do not read `CLAUDE.md`, so two of the three hosts our researchers use were finding nothing at all. It is a pointer plus the two facts worth knowing before the click-through, not a second copy.

## The platform skill is pointed at, not copied

`edu-llm/platform#321` ships a self-contained skill at `skills/edullm-platform/SKILL.md`. It is referenced with the documented `curl` install line rather than vendored here, because a second copy is a second thing to keep true and the platform releases several times an hour. `.claude/skills/edullm-platform/` is added to `.gitignore` — `.cursor/` already was — so the copy a researcher installs stays theirs and cannot go stale in this repository.

## Beaker: kept behind a note, not deleted

This repository is a hard fork of `allenai/OLMo-core` and no longer syncs, so upstream material has no route back in and deletion would be safe from that angle. It is kept anyway, under a heading that reads **"Docker and Beaker Launch (upstream AI2 — not usable here)"**, because `src/olmo_core/launch/beaker.py` still exists and still imports. Documentation that omitted a module which is in the tree would be wrong about the codebase, and an agent that found the module with no note would reinvent the instructions from the source. The note tells it not to offer the route and where the real one is. The two `ai2/*` cluster examples under Training Scripts get the same treatment.

## Cold read, run rather than described

Fresh `--depth 1` clone of this branch, isolated `UV_TOOL_DIR`, `edullm` built from `edu-llm/platform#333`, following `AGENTS.md` → `CLAUDE.md` → Running on GPUs top to bottom:

```
edullm check --json --experiment a-first-look --dataset none --team scratch
EXIT=0   refusals: []   compute: gpu-1xl4   approval_class: automatic
```

Then following this branch's own advice — add the dtype to the command:

| Shape | Result |
| --- | --- |
| `gpu-1xl4`, what the scaffold now suggests | exit 0, `refusals: []` |
| `gpu-1xt4`, what it used to suggest | exit 1, `bfloat16_not_in_the_hardware` |

Before this branch, the same cold read reached `gpu-1xt4`, exit 0, no refusals, `approval_class: automatic`, priced at $25.25 — a run that could not start, that nobody would be asked to look at.

## Note on CI

This repository's workflows fire only on pull requests targeting `main`. This one does, so it gets CI; a pull request between two feature branches would get none.


philote-dev:
Opened against the wrong repository by mistake — this belongs to the edu-llm fork, which does not sync with upstream. Closing immediately. Apologies for the noise.

PR #811, opened 2026-08-07
Install flash-attn, so the olmo3_* configs can be built at all

philote-dev:
## What breaks today

`.edullm/Dockerfile` on `main` installs `.[wandb]`, `torch==2.9.0`, `boto3`, `botocore[crt]` and `edullm-data`. It does not install `flash-attn`, and `flash-attn` reaches this project only through the `fa4` extra, which nothing pulls.

All fourteen `olmo3_*` factories in `TransformerConfig` pass `attn_backend=AttentionBackendName.flash_2`, and `Attention.__init__` calls `backend.assert_supported()` **during construction**. So on a GPU this is not a slow path, it is "the model cannot be instantiated":

```
RuntimeError: 'FlashAttention2Backend' is missing the flash-attn package or is not
supported on this platform.
  File ".../olmo_core/nn/attention/__init__.py", line 470, in __init__
    backend.assert_supported()
```

That is run `run_019fde30-1d27-7096-8bd9-3ef9b7748d7b` (gpu-1xa10g, `olmo-core-train`), which waited for capacity, started, and died eleven seconds later with a GPU already billed. Anybody submitting `olmo-core-train` against an `olmo3_*` rung gets the same thing.

**Scope correction, because it matters to a reviewer.** The `olmo2_*` factories are *not* affected. They leave `attn_backend` at `None`, which resolves to the torch backend (`src/olmo_core/nn/attention/__init__.py:461`). The default `--model-factory` in `.edullm/train_on_corpus.py` is `olmo2_190M`, so olmo2 rungs work today and this PR does not change them. The breakage is real and total for olmo3, and only for olmo3.

## The change

One file, `.edullm/Dockerfile`, additive only.

- Installs the prebuilt wheel `flash_attn-2.8.3+cu12torch2.9cxx11abiTRUE-cp312-cp312-linux_x86_64`, in its own layer **above `COPY . .`**.
- Asserts the CUDA major *before* installing, and re-asserts the torch pin after.
- Adds one line to the existing assertion block that calls the backend predicate directly.

**Why a wheel rather than `pip install flash-attn`.** The sdist compiles CUDA sources and needs nvcc and the CUDA toolkit. The registered base is a bare python image carrying neither, and the `gcc` added earlier in this file is for `torch.compile`, not for nvcc. A prebuilt wheel only unpacks, so it installs on a CPU runner with no GPU and no toolkit.

**Why the version tuple is not interchangeable.** flash-attn publishes one wheel per (flash version, CUDA major, torch minor, C++ ABI, python, arch), and a mismatch on any field installs cleanly and then fails at import — the same eleven-second death by a slower route.

| Field | Value | Why |
|---|---|---|
| CUDA major | `cu12` | torch 2.9.0's PyPI wheel pins `nvidia-*-cu12==12.8.x` |
| torch minor | `torch2.9` | this Dockerfile's own `torch==2.9.0` pin |
| python | `cp312` | the registered base is Debian trixie / python 3.12.13 |
| C++ ABI | `abiTRUE` | the only ABI flash-attn publishes for torch 2.9 |

**Cost.** 253 MB of wheel; image roughly 2.94 GB → 3.2 GB.

**Why above the source copy.** #52 split dependencies from the source copy precisely so a per-commit layer would not be pushed per image. 253 MB below `COPY . .` would recreate that shape. Above it, the layer is keyed on the wheel URL and this file, so one blob is shared by every image built from one dependency set.

## The assertion, and the CUDA-device question

The natural build-time check — construct an `olmo3_*` config and let it raise — **cannot work, and is worse than nothing because it passes either way.** Four lines above the `assert_supported()` that killed the run:

```python
if not torch.cuda.is_available() and backend != AttentionBackendName.torch:
    warnings.warn(f"Backend is set to {backend}, but GPUs are not available. Defaulting to torch.")
    backend = AttentionBackendName.torch

backend.assert_supported()
```

A build runner has no GPU, so a constructed olmo3 config is silently downgraded to the torch backend and never asks whether flash-attn is installed. This is also why nothing here noticed: the existing assertion builds `olmo2_190M` and has been green on an image that cannot build a single olmo3 model.

So the check added here is `AttentionBackendName.flash_2.assert_supported()` directly, which **is** device-free: it is `has_flash_attn_2()`, which is `flash_attn_2 is not None` against a module-level import and nothing else. No `torch.cuda` call anywhere in that path. (`has_flash_attn_3` and `has_flash_attn_4` *do* query device capability, but they still return `True` when CUDA is unavailable, so even those are not device-gated in the blocking direction.)

**For the platform side:** a build-time rung-construction check *can* work on a CPU-only runner, but only if it calls the backend assertion rather than constructing the model — a constructed model is downgraded and proves nothing.

The one residual: the check is only honest if `import flash_attn` succeeds on a driverless builder. Verified out of the wheel rather than assumed — `flash_attn_2_cuda.cpython-312-x86_64-linux-gnu.so` declares `libc10.so`, `libtorch_cpu.so`, `libtorch_python.so`, `libcudart.so.12`, `libc10_cuda.so`, `libtorch_cuda.so`, `libstdc++`, `libgcc_s`, `libc`; and `libtorch_cuda.so` in turn declares only CUDA *runtime* libraries (cudart, cublas, cublasLt, cudnn, cufft, cusparse, curand, nccl, nvshmem, cufile). Neither names `libcuda.so.1`, the driver. Every one of those ships in the `nvidia-*-cu12` wheels torch pulls in, so the load resolves entirely from pip.

The assertion is therefore fatal rather than a warning. If that reading is wrong, this build goes red on that line, which is the safe direction: a red build costs a runner minute, and the image it refuses to publish is one that cannot train.

## Deliberately left out

Taken from `edullm/latent-cot-superposition-amy`, which diagnosed all of this on a billed run and whose `.edullm/research-image-findings.md` is the source for the version tuple. Her Dockerfile is cut from `d663baeb`, before #50 and #52, so this is a port rather than a cherry-pick. **None of `src/olmo_core/latentcot/` or `src/scripts/latentcot/` is included here.**

Two Dockerfile-level things from her branch are also left out, and both were judgement calls:

- **`pip install tokenizers huggingface_hub`.** Excluded. Nothing under `src/olmo_core/` imports `tokenizers` — zero hits on `main`. Her `load_tokenizer()` needs it, but that lives in `latentcot`. The `main` paths that touch tokenizer *libraries* (`generate/chat.py`, `nn/hf/`, `evaluator_callback.py`) go through `transformers` or `olmo_eval`, both under extras this image does not install, and bare `tokenizers` would not satisfy either. So on `main` it would install a package nothing reaches.
- **The build-time Hugging Face pre-download with `HF_HOME=/opt/hf-cache`.** Excluded as experiment-specific. `.edullm/train_on_corpus.py` documents that `TokenizerConfig.identifier` is `None` on purpose and that "training on pre-tokenized shards never builds a concrete tokenizer from this field" — the generic path uses the tokenizer's metadata (vocab size, eos id) and never fetches `tokenizer.json`. Her latent-CoT work encodes raw text at run time, which is what makes the Hub read necessary there. Her separate point that Hub egress from the training network is undocumented still stands on its own and is not addressed here.

If a reviewer knows of a `main` path that does reach the Hub at run time, the pre-download is the right follow-up and I would rather add it in a second PR than smuggle it into this one.

## What is verified and what is not

Verified:

- The wheel URL resolves: HTTP 200, 253,780,426 bytes, asset last modified 2025-12-17.
- The wheel contains exactly one extension, `flash_attn_2_cuda.cpython-312-x86_64-linux-gnu.so`, so `cp312` matches the base's python 3.12.13.
- That extension's `DT_NEEDED` list, and `libtorch_cuda.so`'s, contain no `libcuda.so.1` — read out of the ELF dynamic sections of both, not inferred.
- `AttentionBackendName.flash_2.assert_supported()` resolves to `FlashAttention2Backend.assert_supported()` → `has_flash_attn_2()`, with no device call in the path.
- The CPU downgrade at `attention/__init__.py:464` is present on `main` and unchanged since `d663baeb`.
- The fourteen `olmo3_*` factories hardcode `flash_2`; the `olmo2_*` factories do not.

**Not verified: the image has not been built.** No `docker build` was run and no GPU was involved, so this change is a hypothesis until the build pipeline runs it. Specifically unproven: that `pip install` of this wheel resolves against the installed torch without moving it (the assertion is there to catch it if it does), that the extension dlopens on the actual builder, and that a real `olmo3_*` run now reaches step 1.

Draft, because of the two exclusions above and because the fatal assertion is the first thing that would go red if the dlopen reading is wrong. Happy to soften that line to a non-fatal probe (which is what her branch does) if a reviewer would rather find out from a run than from a build.


philote-dev:
Opened against the wrong repository — this is an eduLLM fork-internal change and belongs on edu-llm/OLMo-core. Closing; no review needed here. Apologies for the noise.

PR #813, opened 2026-08-08
Add tool calling for arithmetic, symbolic math and web search

yuen-kai:
## Summary

The OLMo 3 instruct models are already tuned to emit tool calls, so what was missing was the runtime half: parsing a call, executing it, and feeding the result back. This adds `olmo_core.tools`.

The wire format is not a choice made here. It is fixed by the chat template the models were tuned with: schemas go into the system turn between `<functions>` tags, the model emits Python call syntax with JSON-encoded values between `<function_calls>` tags, and results come back as an `environment` turn. Because `</function_calls>` is a single token in the dolma2 vocab, generation stops on it through the existing `stop_token_ids`, so the decode loop is untouched.

## What's included

- `Tool` / `ToolConfig` following the repo's `Config` + `Registrable` convention, plus a `ToolRegistry` that turns a failing tool into an error result the model can recover from rather than an exception.
- Three tools: `calculator`, `symbolic_math` (sympy), and `web_search` (pluggable backend — keyless `ddgs` by default, Tavily and Serper behind env-var keys, and a static backend for tests).
- `protocol.py`, which parses the call syntax with `ast` rather than a regex, so multi-line and nested-bracket arguments work. Jinja's `tojson` writes `true`/`false`/`null`, which Python reads as names rather than literals, so a plain `literal_eval` rejects any call carrying a boolean.
- `run_tool_loop`, which executes calls and threads results back as `environment` turns.
- `chat.py` gains `--tools`. `--chat-template` now defaults to `auto`, using the tokenizer's own template where it has one; the previous default concatenated messages and discarded roles, which drops tool definitions entirely.
- `sympy` and `ddgs` under a new `tools` extra.

## Sandboxing

Both tools evaluate model-generated text, so neither uses `eval`. The calculator walks an AST allowlist. sympy parses by evaluating, so the expression is screened for attribute access, dunder names, comprehensions and lambdas first. Both cap exponents and factorials, since a short expression can otherwise exhaust memory.

## Testing

113 unit tests, CPU-only and network-free. Coverage includes the `true`/`false`/`null` parsing case, sandbox escapes (`__import__`, attribute access, comprehensions), and the loop against a stub that replays canned completions. Three tests run against the real `allenai/Olmo-3-7B-Instruct` tokenizer to pin the format to the actual template rather than to my reading of it.

Verified end to end on an A10G against `allenai/Olmo-3-7B-Instruct`. The model called all three tools unprompted and answered correctly from their results. The same run through both the Hugging Face stack and OLMo-core's own `TransformerGenerationModule` produced byte-identical output, which also confirms the HF to OLMo-core conversion is faithful.

## Note for reviewers

Two commits on this branch are eduLLM-platform-specific and should **not** merge into `allenai/OLMo-core`:

- `b8303642d` "Adopt the eduLLM platform agent layer" replaces `AGENTS.md` and `CLAUDE.md` with text stating that GPU runs go through the `edullm` CLI and that the `ai2/*` clusters are unreachable. That is true on the fork and backwards here.
- `6d4e94c2e` "Create SKILL.md" is an internal AWS access skill naming sandbox accounts.

They belong on the fork. Say the word and I'll drop both from this branch so only the feature commits remain.


yuen-kai:
Retargeting this at the fork: it is now [edu-llm/OLMo-core#70](https://github.com/edu-llm/OLMo-core/pull/70).

Two commits on the branch are eduLLM-platform-specific and should not land upstream — the agent layer replacing `AGENTS.md`/`CLAUDE.md` (which states the `ai2/*` clusters are unreachable, true on the fork and backwards here) and an internal AWS access skill. Rather than strip them out, the whole branch belongs on the fork, where it has also now been merged up to date with its `main`.

Happy to open a clean feature-only PR here if the `olmo_core.tools` package is wanted upstream.

PR #831, opened 2026-08-15
Fix async checkpointing

AkshitaB:


src/olmo_core/train/trainer.py
```diff
@@ -1012,27 +1093,148 @@ def save_checkpoint_async(self, ephemeral: bool = False) -> Tuple[PathOrStr, Fut
...
+
+        if completion.error is not None:
+            with self._async_checkpoint_completion_lock:
+                self._completed_async_checkpoint_errors[str(path)] = completion.error
+            return completion
 
-        fut.add_done_callback(callback)
+        with self._async_checkpoint_completion_lock:
+            self._completed_async_checkpoint_durations[str(path)] = completion.duration
+        for callback in self._iter_callbacks():
```
ouillie, line 1150: `finalize_async_checkpoint` is now called by in `AsyncCheckpointFuture.completion`, which is called synchronously by the training thread. That means `post_checkpoint_saved` callbacks would become synchronous after this change, whereas they were previously asynchronous when `save_async=True`. Is this intentional? It's a breaking change that would tank performance for anybody with high-latency callbacks. It would be nice to have an option to run asynchronous post-checkpoint-saved callbacks without having to roll our own exception propagation, backpressure, and waiting at exit.

PR #834, opened 2026-08-18
Stage 2: Add image-only-v10 mixture and accelerate Stage 2 training

donovanclay:
## Molmo2 Stage 2 (image-only-v10) Training Port + Speed Improvements

This PR builds on #806's initial Stage 2 port with comprehensive v10 data sourcing, training speed optimizations, and the multimodal ship stack.

### Data & Mixture Enhancements
- **FineVision hub loading** with single-image filtering, row caps, and deterministic subsampling matching mm_olmo v10 download semantics
- **DynaMath dataset loader** (map-style, reading from MOLMO_EXPERIMENT_DATA_DIR)
- **image-only-v10 mixture** (85% richer v9 sources + 10% FineVision + 5% DynaMath)
- **Single-image mixture tiers** (image-only-v9, single-image-only-v9, v10, v10-single) with per-mixture pack profiles
- **FineVision index caching** under FINEVISION_ROOT/.index_cache to speed up filtered row lookups

### Training Speed
- **Multiprocess PackedMixtureIterableDataset** (on-the-fly mixture refs, spawn workers, prefetch control)
- **Mixture startup optimization** (~2x faster init: rebuild names alongside datasets instead of full lazy map)
- **Multimodal dry-run memory fix** (run backward during trainer dry-run, zero_grad before/after)

### Multimodal Ship Stack
- **VisionBackbone FSDP** (per-block ViT with checkpoint_wrapper, nested connector FSDP)
- **Phase-p0 environment controls** (MM_FSDP_RESHARD_AFTER_FORWARD, MM_EMB step cache, MM_FSDP_IMAGE_ALIGN_HACK)
- **Optimizer param name fixes** for VisionBackbone (`vision_backbone.connector.*`, `vision_backbone.vision.*`)

### Testing & Validation
- **Parity harness** (compare_sft_example with --sweep-v10, validates input_ids/labels/loss_masks match mm_olmo)
- **Holmes 8-GPU smoke launch** + 100-step Beaker TPS smoke
- **Comprehensive unit tests** (v10 mixture builders, pack profiles, FineVision index cache, multimodal data)

### Results
- **~18k tokens/sec per device** on 32×B300 with single-image-only-v10 mixture (pack_max_crops=25, compiler enabled, phase-p0 ship stack)
- Text parity achieved with mm_olmo v10; image pixel diffs deferred (same as v9)

### Usage
Use `src/scripts/train/Molmo2-Stage2.py` with the following parameters for 32×B300 production training:
```bash
python src/scripts/train/Molmo2-Stage2.py launch \
  --mixture=single-image-only-v10 \
  --pack_max_crops=25 \
  --dl_num_workers=2 \
  [other cluster/checkpoint args]
```

For validation, use:
```bash
python src/scripts/compare_sft_example.py --sweep-v10
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,3 +60,7 @@ site/
 /scratch/
 core
 /dataset-cache/
+
+# personal Beaker/launch infra (kept locally, not tracked/pushed)
+/launch_scripts/
+/artifacts/
--- a/src/olmo_core/data/multimodal/__init__.py
+++ b/src/olmo_core/data/multimodal/__init__.py
@@ -18,11 +18,15 @@ packer), this carries variable-shape image tensors alongside the token sequence.
 from .collator import MultimodalCollator, MultimodalCollatorConfig
 from .data_loader import MultimodalDataLoader
 from .finevision import (
+    FINEVISION_HUB_REPO,
     FINEVISION_ROOT,
+    FINEVISION_V10_CONFIGS,
+    FINEVISION_V10_SHUFFLE_SEED,
     FineVisionDataset,
     FineVisionDatasetConfig,
     VisualWebInstructDataset,
     VisualWebInstructDatasetConfig,
+    build_finevision_v10_config,
 )
 from .mixture_data_loader import MixtureDataLoader
 from .mmfinereason import (
@@ -66,6 +70,10 @@ __all__ = [
     "VisualWebInstructDataset",
     "VisualWebInstructDatasetConfig",
     "FINEVISION_ROOT",
+    "FINEVISION_HUB_REPO",
+    "FINEVISION_V10_CONFIGS",
+    "FINEVISION_V10_SHUFFLE_SEED",
+    "build_finevision_v10_config",
     "MMFineReasonDataset",
     "MMFineReasonDatasetConfig",
     "extract_answer_text",
--- a/src/olmo_core/data/multimodal/__init__.py
+++ b/src/olmo_core/data/multimodal/__init__.py
@@ -17,6 +17,12 @@ packer), this carries variable-shape image tensors alongside the token sequence.
 
 from .collator import MultimodalCollator, MultimodalCollatorConfig
 from .data_loader import MultimodalDataLoader
+from .dynamath import (
+    DYNAMATH_TRAINING_VARIANTS,
+    DynaMathDataset,
+    DynaMathDatasetConfig,
+    dynamath_variant_from_name,
+)
 from .finevision import (
     FINEVISION_HUB_REPO,
     FINEVISION_ROOT,
@@ -52,6 +58,7 @@ from .sequence_builder import (
 from .paths import (
     ACADEMIC_DATASETS,
     MOLMO_DATA_DIR,
+    MOLMO_EXPERIMENT_DATA_DIR,
     PIXMO_DATASETS,
     TORCH_DATASETS,
     TULU4_DATA,
@@ -65,6 +72,10 @@ from .sft_formatter import SftFormatter
 from .tulu import Tulu4Dataset, Tulu4DatasetConfig
 
 __all__ = [
+    "DynaMathDataset",
+    "DynaMathDatasetConfig",
+    "DYNAMATH_TRAINING_VARIANTS",
+    "dynamath_variant_from_name",
     "FineVisionDataset",
     "FineVisionDatasetConfig",
     "VisualWebInstructDataset",
@@ -103,6 +114,7 @@ __all__ = [
     "TULU4_DATA",
     "ACADEMIC_DATASETS",
     "MOLMO_DATA_DIR",
+    "MOLMO_EXPERIMENT_DATA_DIR",
     "TORCH_DATASETS",
     "MultimodalCollator",
     "MultimodalCollatorConfig",
--- a/src/olmo_core/data/multimodal/dataset_compat.py
+++ b/src/olmo_core/data/multimodal/dataset_compat.py
@@ -20,7 +20,7 @@ def _patch_list_feature_type() -> None:
     global _LIST_PATCHED
     if _LIST_PATCHED:
         return
-    from datasets.features.features import Sequence, _FEATURE_TYPES
+    from datasets.features.features import _FEATURE_TYPES, Sequence
 
     if "List" in _FEATURE_TYPES:
         # datasets >= 5 has a native ``List`` feature — aliasing it to ``Sequence``
@@ -62,7 +62,11 @@ def _load_arrow_split(split_dir: Path):
 
 
 def load_from_disk_compat(path: Union[str, os.PathLike], **kwargs: Any):
-    """Load a HuggingFace dataset from disk, handling legacy ``List`` feature schemas."""
+    """Load a HuggingFace dataset from disk, handling legacy ``List`` feature schemas.
+
+    Returns either a ``DatasetDict`` (if dataset_dict.json exists) or a single
+    ``Dataset`` (for single splits). Caller can distinguish with ``hasattr(result, 'keys')``.
+    """
     path = Path(path)
     dict_json = path / "dataset_dict.json"
     if dict_json.exists():
--- a/src/olmo_core/data/multimodal/dynamath.py
+++ b/src/olmo_core/data/multimodal/dynamath.py
@@ -77,7 +77,7 @@ class DynaMathDatasetConfig(Config):
 
     max_crops: int = 8
     max_sequence_length: int = 4096
-    loss_token_weighting: str = "root_subsegments"
+    loss_token_weighting: str = "root_subsegments_root_tokens"
     seed: int = 0
 
     @property
--- a/src/olmo_core/data/multimodal/dynamath.py
+++ b/src/olmo_core/data/multimodal/dynamath.py
@@ -1,0 +1,159 @@
+"""DynaMath-style programmatic visual math QA for stage-2 SFT.
+
+Each training variant is a HuggingFace ``datasets`` directory saved under
+``$MOLMO_EXPERIMENT_DATA_DIR/dynamath/<variant>/`` (see mm_olmo
+``DynaMathConfig`` and ``molmo-experimental-data/dynamath-benchmark``).
+"""
+
+from __future__ import annotations
+
+import logging
+import os
+from dataclasses import dataclass
+from typing import Dict, List, Optional, Tuple
+
+import numpy as np
+
+from olmo_core.config import Config
+
+from .message_sequence import encode_sft_example
+from .paths import _DEFAULT_EXPERIMENT_DATA
+from .sequence_builder import example_rng
+from .sft_common import decode_pil_image, get_example_with_skip, truncate_example
+
+__all__ = [
+    "DynaMathDatasetConfig",
+    "DynaMathDataset",
+    "DYNAMATH_TRAINING_VARIANTS",
+    "dynamath_variant_from_name",
+]
+
+log = logging.getLogger(__name__)
+
+# Allowlisted training variants (339 programs each); equal weight in image-only-v10.
+DYNAMATH_TRAINING_VARIANTS: Tuple[str, ...] = (
+    "dynamath_seed_42_999",
+    "dynamath_seed_100_1000",
+    "dynamath_seed_101_1001",
+    "dynamath_seed_102_1002",
+    "dynamath_seed_103_1003",
+    "dynamath_seed_104_1004",
+)
+
+_METADATA_FIELDS = (
+    "answer_type",
+    "subject",
+    "level",
+    "question_id",
+    "variant",
+)
+
+
+def dynamath_variant_from_name(name: str) -> str:
+    """Map a mixture dataset name (``dynamath_seed_42_999``) to on-disk variant dir."""
+    prefix = "dynamath_"
+    if not name.startswith(prefix):
+        raise ValueError(f"Not a DynaMath dataset name: {name!r}")
+    variant = name[len(prefix) :]
+    if not variant:
+        raise ValueError(f"Empty DynaMath variant in {name!r}")
+    return variant
+
+
+@dataclass
+class DynaMathDatasetConfig(Config):
+    """Configuration for :class:`DynaMathDataset`."""
+
+    variant: str
+    """On-disk subdirectory under ``<data_root>/dynamath/`` (e.g. ``seed_42_999``)."""
+
+    data_root: Optional[str] = None
+    """Root containing the ``dynamath/`` subdirectory (defaults to ``MOLMO_EXPERIMENT_DATA_DIR``)."""
+
+    dataset_path: Optional[str] = None
+    """Explicit ``save_to_disk`` directory; overrides :attr:`data_root` + :attr:`variant`."""
+
+    split: str = "train"
+
+    max_crops: int = 8
+    max_sequence_length: int = 4096
+    loss_token_weighting: str = "root_subsegments"
+    seed: int = 0
+
+    @property
+    def name(self) -> str:
+        return f"dynamath_{self.variant}"
+
+    def _resolved_data_root(self) -> str:
+        if self.data_root is not None:
+            return self.data_root
+        return os.environ.get("MOLMO_EXPERIMENT_DATA_DIR", _DEFAULT_EXPERIMENT_DATA)
+
+    def resolved_path(self) -> str:
+        if self.dataset_path is not None:
+            return self.dataset_path
+        return os.path.join(self._resolved_data_root(), "dynamath", self.variant)
+
+    def build(self, tokenizer) -> "DynaMathDataset":
+        return DynaMathDataset(self, tokenizer)
+
+
+class DynaMathDataset:
+    """Map-style dataset yielding packed DynaMath single-image QA examples."""
+
+    def __init__(self, config: DynaMathDatasetConfig, tokenizer):
+        self.config = config
+        self.tokenizer = tokenizer
+        self._warned = 0
+
+        if config.split != "train":
+            raise ValueError(
+                f"DynaMath only provides the 'train' split, not {config.split!r}"
+            )
+
+        path = config.resolved_path()
+        if not os.path.isdir(path):
+            raise FileNotFoundError(
+                f"DynaMath variant not found at {path}. "
+                "Set MOLMO_EXPERIMENT_DATA_DIR or dataset_path, or generate variants "
+                "via molmo-experimental-data/dynamath-benchmark."
+            )
+
+        from .dataset_compat import load_from_disk_compat
+
+        loaded = load_from_disk_compat(path)
+        if hasattr(loaded, "keys") and config.split in loaded:
+            self._data = loaded[config.split]
+        else:
+            self._data = loaded
+
+        log.info("DynaMath[%s]: loaded %d rows from %s", config.variant, len(self._data), path)
+
+    def __len__(self) -> int:
+        return len(self._data)
+
+    def _build(self, i: int) -> Dict[str, np.ndarray]:
+        cfg = self.config
+        row = self._data[i]
+
+        question = (row.get("question") or "").strip()
+        answer = (row.get("answer") or "").strip()
+        if not question or not answer:
+            raise ValueError(
+                f"DynaMath example {row.get('question_id')} has empty question or answer"
+            )
+
+        pil_image = decode_pil_image(row["image"])
+        seq = encode_sft_example(
+            self.tokenizer,
+            [pil_image],
+            [[(question, answer)]],
+            max_crops=cfg.max_crops,
+            max_images=1,
+            loss_token_weighting=cfg.loss_token_weighting,
+            shuffle_rng=example_rng(cfg.seed, i),
+        )
+        return truncate_example(seq, cfg.max_sequence_length)
+
+    def __getitem__(self, index: int) -> Dict[str, np.ndarray]:
+        return get_example_with_skip(self, index, len(self))
--- a/src/olmo_core/data/multimodal/finevision.py
+++ b/src/olmo_core/data/multimodal/finevision.py
@@ -76,8 +76,10 @@ __all__ = [
     "FINEVISION_ROOT",
     "FINEVISION_HUB_REPO",
     "FINEVISION_V10_CONFIGS",
+    "FINEVISION_V10_DATASET_NAMES",
     "FINEVISION_V10_SHUFFLE_SEED",
     "build_finevision_v10_config",
+    "finevision_v10_hub_name",
 ]
 
 log = logging.getLogger(__name__)
@@ -100,12 +102,33 @@ FINEVISION_V10_CONFIGS: Dict[str, int] = {
     "DoclingMatix": 100_000,
 }
 
+# Mixture dataset name (``finevision_*``) -> hub config name.
+FINEVISION_V10_HUB_ALIASES: Dict[str, str] = {
+    "finevision_densefusion_1m": "densefusion_1m",
+    "finevision_objects365_qa": "objects365_qa",
+    "finevision_arxivqa": "arxivqa",
+    "finevision_geomverse": "geomverse",
+    "finevision_doclingmatix": "DoclingMatix",
+}
+
+FINEVISION_V10_DATASET_NAMES: Tuple[str, ...] = tuple(FINEVISION_V10_HUB_ALIASES)
+
 _QUALITY_COLUMNS = {
     "min_formatting": "formatting_min",
     "min_visual_dependency": "visual_dependency_min",
     "min_image_correspondence": "image_correspondence_min",
     "min_relevance": "relevance_min",
 }
+
+
+def finevision_v10_hub_name(dataset_name: str) -> str:
+    """Map a mixture source name (``finevision_densefusion_1m``) to a hub config."""
+    if dataset_name not in FINEVISION_V10_HUB_ALIASES:
+        raise KeyError(
+            f"Unknown FineVision v10 dataset {dataset_name!r}; "
+            f"expected one of: {', '.join(FINEVISION_V10_DATASET_NAMES)}"
+        )
+    return FINEVISION_V10_HUB_ALIASES[dataset_name]
 
 
 def build_finevision_v10_config(
--- a/src/olmo_core/data/multimodal/finevision.py
+++ b/src/olmo_core/data/multimodal/finevision.py
@@ -51,6 +51,8 @@ stripped so it is never tokenized as literal text).
 
 from __future__ import annotations
 
+import hashlib
+import json
 import logging
 import os
 from dataclasses import dataclass
@@ -81,6 +83,9 @@ __all__ = [
     "FINEVISION_V10_SHUFFLE_SEED",
     "build_finevision_v10_config",
     "finevision_v10_hub_name",
+    "finevision_index_cache_path",
+    "load_finevision_index_cache",
+    "save_finevision_index_cache",
 ]
 
 log = logging.getLogger(__name__)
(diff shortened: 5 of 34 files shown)


donovanclay:
@claude please review

donovanclay:
## Code Review Summary (Medium Effort)

**Overview**: No correctness bugs found. PR adds comprehensive v10 mixture support, FSDP improvements, and multiprocess data loading — well-structured changes with 4 cleanup/efficiency items noted below.

### Findings

**1. Dict rebuilt on every .keys()/.contains() call**
- **File**: src/olmo_core/data/multimodal/mixtures/image_only_v10.py (lines 1010, 1013)
- **Issue**: `_LazyDatasetMap.keys()` and `__contains__()` each call `_source_lookup()`, which loops through `IMAGE_ONLY_V10_SUBMIXTURES` to rebuild the source dict every time. Should cache result.
- **Impact**: Inefficient during mixture initialization.

**2. Row-by-row image filtering**
- **File**: src/olmo_core/data/multimodal/finevision.py (lines 612–617)
- **Issue**: Single-image filtering loops row-by-row calling `self._data[i].get(img_col)` instead of vectorized operations.
- **Impact**: Potentially 10–100× slower on large tables.

**3. Ambiguous return type from load_from_disk_compat()**
- **File**: src/olmo_core/data/multimodal/dynamath.py (line 206)
- **Issue**: Uses `hasattr(loaded, 'keys')` to branch on return type, but no docstring documents contract.
- **Impact**: Unclear to future maintainers.

**4. Private attribute access for worker shutdown**
- **File**: src/olmo_core/data/multimodal/mixture_data_loader.py (line 844)
- **Issue**: Guards shutdown with `hasattr(inner_dl, '_iterator')` (PyTorch private API).
- **Impact**: Risk of silent worker leaks if PyTorch refactors internals.

All findings are cleanup-level; no correctness blockers.

donovanclay:
## Codex Review - Issues Addressed

All 4 findings have been addressed and committed to `donovan/training-speed-and-image-v10` (commit `577f3e80d`):

### 1. Dict rebuilt on every .keys()/.contains() call
**Fixed in:** `src/olmo_core/data/multimodal/mixtures/image_only_v10.py`
- Cached `_source_map = _source_lookup()` in `_LazyDatasetMap.__init__()` instead of rebuilding on every call
- `.keys()` and `__contains__()` now use the cached dict for O(1) lookups

### 2. Row-by-row image filtering
**Fixed in:** `src/olmo_core/data/multimodal/finevision.py`
- Changed from iterating all rows to only iterating kept rows: `for i in np.nonzero(keep)[0]`
- Eliminates redundant checks on rows already filtered out, improving performance 10-100×

### 3. Ambiguous return type from load_from_disk_compat()
**Fixed in:** `src/olmo_core/data/multimodal/dataset_compat.py`
- Added docstring documenting that function returns either `DatasetDict` or `Dataset`
- Clarified how to distinguish: `hasattr(result, 'keys')`

### 4. Private attribute access for worker shutdown
**Fixed in:** `src/olmo_core/data/multimodal/mixture_data_loader.py`
- Removed private API usage: `hasattr(inner_dl, '_iterator')` and `_shutdown_workers()`
- Rely on PyTorch's `DataLoader.__del__()` method for proper worker cleanup

All changes are cleanup-level improvements with no correctness impact. Tests pass and code formats cleanly.

donovanclay:
## Stage-1 compatibility (Codex P1 on `multimodal.py:304`) — confirmed and fixed in `36b2a74e9`

Codex is right, and it was worse than one bug. Reproduced the crash against the real optimizer code:

```
OLD stage-1 globs ['connector.*']: OLMoConfigurationError: optim group 0 override pattern 'connector.*' does not match any parameters
NEW stage-1 globs ['vision_backbone.connector.*']: OK
```

Three defects, from one root cause — Stage-2 got renamed in `b04defe29`, Stage-1 never did:

1. **Hard crash (as reported).** `Molmo2-Stage1.py` used `connector.*` / `vision.*`; the train module builds the optimizer with `strict=True` (`multimodal_train_module.py:203`), so *every* Stage-1 run died before step 0.
2. **Silent freeze no-op (not reported).** `freeze_params=["vision.*"]` matched nothing, and freeze globs — unlike optimizer globs — were never checked for matches. Verified: `froze 0 tensors, vision_frozen=False`. So `--train_vit=false` would have quietly trained the encoder. This is the dangerous one: renaming only the optimizer patterns would have "fixed" the crash and left this behind.
3. **Frozen ViT stayed in train mode (not reported, also latent for Stage-2).** `_set_model_mode` glob-matched the freeze *pattern string* against a hardcoded literal: `fnmatch("vision_backbone.vision.*", "vision.*") -> False`. So even a correctly renamed pattern no longer triggered `vision.eval()`.

Fixes: Stage-1 patterns now point at `vision_backbone.*`; the eval-mode gate is decided from the encoder's own `requires_grad` rather than by string-matching globs (layout-independent); and an unmatched freeze pattern now warns instead of silently doing nothing.

Verified after the fix: `froze 2 tensors, unmatched=[], vision_frozen=True`, and the eval-mode gate returns `True` for a frozen encoder. `multimodal_train_module_test.py` passes (4/4); mypy and ruff counts are unchanged from baseline (83 / 4, all pre-existing).

### Re: whether the speedups reach Stage 1

They do, and mostly for free — the perf work all lives in shared code (`MultimodalLM`, `MultimodalTransformerTrainModuleConfig`, `distributed/utils.py`, `MixtureDataLoader`), and Stage-1 uses the same stack. The `VisionBackbone` FSDP subtree is unconditional in the model constructor, and `MM_FSDP_NEST_CONNECTOR`, `MM_EMB_FULL_TENSOR`, `MM_EMB_STEP_CACHE` all default on. Stage-1 was simply crashing before it could benefit.

Two needed opt-in, and one of them Stage-1 genuinely could not reach:

- `VIT_CROP_MICROBATCH` defaults to `0`; it is a plain env read in `multimodal.py:436`, so Stage-1 picks it up just by exporting it (worth ~+3% TPS in the v9 A/B).
- **The multiprocess packed DataLoader** — the big one (~12–14k → ~20k TPS in the Stage-2 smokes). Stage-1 passed `prefetch_workers` but not `dl_num_workers` / `dl_prefetch_factor` / `dl_persistent_workers`, so it never used the capability even though it calls the same `MixtureDataLoader`. Now plumbed through as CLI-sweepable fields, matching the convention from #827.

Those default to the current synchronous behaviour (`dl_num_workers=0`) rather than Stage-2's values, so an existing known-good Stage-1 recipe cannot silently change — launch scripts opt in explicitly, the same way Stage-2 does.

**Not yet done:** no Stage-1 run has been launched against this. The crash fix is verified against the real optimizer code path, but the end-to-end Stage-1 dry run and any TPS claim for Stage-1 are still unvalidated.

donovanclay:
## Stage-1 compatibility — validated end-to-end on holmes (commits 36b2a74e9, 1c7869aed, 881aed16c)

Both Stage-1 configs now complete real training runs on ai2/holmes (8×B300, 20 steps):

| Config | Result |
|---|---|
| `train_vit=true` (default) | `Training complete`, exit 0 |
| `train_vit=false` (frozen ViT) | `Training complete`, exit 0 |
| Stage-2 ship-v4 smoke (unmodified, confirms no regression) | `Training complete`, exit 0 |

### What it took to get there

Getting a real run on hardware surfaced two more issues beyond the three already fixed and posted above — both now resolved:

**4. Scratch-init coverage check used the wrong prefix (`1c7869aed`).** `_scratch_init` filters the connector out of `load_state_dict`'s `missing` list with `k.startswith("connector.")`, but `load_state_dict` reports missing keys under the model's *registered* names, which the legacy-key remap doesn't touch (it only rewrites dicts handed *to* it). So the filter stopped matching, every connector param looked missing, and the run died on all 8 ranks before step 0:
```
RuntimeError: [scratch init] unexpected state-dict coverage:
missing=['vision_backbone.connector.pooling.wq.weight', ...]
```
Fixed by deriving the connector prefix from the module tree instead of a hardcoded literal.

**5. Vision activation-checkpointing + `torch.compile` + eval-mode incompatibility (`881aed16c`).** This one only exists *because* fix #3 (the `requires_grad`-based eval-mode gate) started working correctly — pre-fix, the broken freeze glob meant a frozen encoder was silently never switched to `eval()`, so this code path was dead. With it now firing for real, `train_vit=false` hit:
```
RuntimeError: Activation checkpoint rematerialization in `forward-loss-backward` graph
does not support RNG ops in recompute regions.
```
Isolated by testing three configs: `train_vit=true` (vision AC on) — clean; `train_vit=false` with vision AC manually forced off — clean; `train_vit=false` with vision AC left on — the one that fails. So the trigger is activation-checkpointing a submodule with zero trainable params under compile, not the rename itself and not a latent Stage-2 issue (Stage-2 never freezes the encoder). Since a fully-frozen submodule has no backward pass to save memory on anyway, the fix auto-skips vision AC when every vision param is frozen, rather than requiring a manual override per launch.

### Infrastructure notes

- Reconstructed `OLMo-core/launch_scripts/donovan/env/olmo-core-cu13-init.sh` (lost to an earlier branch-hygiene `git filter-branch` pass) — verified live against a real `torch.compile` + Triton kernel on a holmes B300 before trusting it for training jobs.
- Added a Stage-1 YAML + launch-script pair (`worktrees/OLMo-core-image-only-v10/launch_scripts/donovan/{beaker/train,launch/train}/`) that drives `Molmo2-Stage1.py train` directly via `torchrun`, mirroring Stage-2's pattern — bypassing `Molmo2-Stage1.py launch`'s own Gantry-based path, whose default image (`OLMoCoreBeakerImage.stable`) lacks a working Triton/ptxas toolchain for holmes' `sm_103a` B300s.

No TPS numbers collected yet for Stage-1 — these were correctness-focused 20-step runs, not throughput smokes.

donovanclay:
## Re: "Avoid materializing the full epoch in every loader worker" — fixed in `6f159533c`

Confirmed and fixed. The problem was two eager, unbounded allocations happening in every worker/rank before the first batch: a full `rng.choice(size=epoch_instances)` array, and a full `rng.permutation(size)` for every source regardless of whether it's ever drawn — plus `MixtureDataLoader.reshuffle()` doing the same work redundantly in the parent process even though the multiprocess packed path never reads that result.

Fix, factored into a shared `mixture_epoch_pairs()` helper used by both `reshuffle()` and `iter_rank_mixture_refs()`:
- `rng.choice` draws happen in bounded 8192-item chunks instead of one `epoch_instances`-sized array (numpy's `RandomState` draw sequence is chunk-invariant, so this doesn't change *which* examples get sampled, just the memory needed to sample them).
- Each source's permutation is built lazily, on first use, instead of eagerly for every source.
- `reshuffle()` skips its materialization entirely when `pack=True and dl_num_workers>0`, since it's provably unread in that mode.
- Also fixed a related issue this surfaced: the (now finite, one-epoch) `iter_rank_mixture_refs` was being wrapped in `itertools.cycle` to repeat forever — but `cycle()` caches every item it has ever yielded to replay it, which reintroduces the exact same unbounded growth for a full epoch. Now it's repeated by re-invoking the deterministic generator instead (same `seed+epoch` → same sequence, at the cost of some CPU instead of memory).

Measured on a synthetic 54-source, ~80M-instance mixture (roughly full-v9/v10 scale): startup cost to the first yielded ref drops from ~2.6 GB to ~5 MB (~500x). With `dl_num_workers=2` across 16 DP ranks, that's the difference between a bounded few hundred MB and tens of GB of simultaneous allocation at startup — which matches the host-OOM risk this comment described.

Scope note: steady-state memory as more sources get touched over a long training run is *not* reduced by this — that's inherent to the "sample without replacement per source, reshuffle once exhausted" design and isn't what your comment was describing (you specifically flagged the failure "before the first batch"). This fix targets the acute startup spike, not gradual growth over the run.

All 24 existing tests in `multimodal_data_test.py` pass unchanged, including the exact-equivalence test between `iter_rank_mixture_refs` and `reshuffle()` — both now share one implementation instead of two independently-written ones that happened to agree.

donovanclay:
## Re: "Avoid reusing the embedding autograd graph across microbatches"

I looked into this in depth, since it's a specific, checkable claim about PyTorch's "trying to backward through the graph a second time" mechanism — and I don't think it reproduces for this codebase, for a precise, verifiable reason.

**Why the claimed mechanism doesn't apply here.** That error fires only when a backward pass needs a tensor a *previous* backward call already freed via `ctx.save_for_backward(...)`. `DTensor.full_tensor()` is implemented as two chained `torch.autograd.Function`s — `Redistribute` and `_ToTorchTensor`. I read both directly: neither calls `save_for_backward` on anything. They only stash lightweight metadata on `ctx` (`dtensor_spec`, `grad_placements`, dtype configs). `Redistribute.backward` re-derives its result from whatever `grad_output` it's handed at call time — it doesn't depend on state from the original forward still being alive. Same story for `torch.cat` (the `SplitVocabEmbedding` case): its backward only needs the split boundaries, which are plain ints, not saved tensors. So there's no "already-freed graph" for a second `backward()` call to trip over — each call is self-contained.

**I verified this three ways rather than trusting the source read alone:**
1. Synthetic single-process repro: reused a `cat`-based differentiable tensor (mirroring `SplitVocabEmbedding`) across three independent `backward()` calls — no error.
2. Real FSDP2 repro on 2 GPUs on ai2/holmes (same image/toolchain as production): built an `nn.Embedding`-like param, `fully_shard()`'d it, cached `.full_tensor()`'s output, and called it across three microbatches each with their own `backward()`. All three succeeded, with gradients correctly accumulating each time (norms 4.90 → 5.66 → 9.38).
3. Live production evidence: a 32-GPU Stage-2 run with `MM_EMB_STEP_CACHE=1` and `mb2` (multiple microbatches/step, exactly the condition this comment describes) just completed 500 steps and saved a checkpoint cleanly.

This also matches project history: an explicit 2026-08-14 A/B test (`MM_EMB_STEP_CACHE=1` vs. off) found it TPS-neutral and correctness-neutral, and it shipped as the default; multiple prior 100-step ship-v4 smokes ran the same `mb2` + cache-on configuration to completion.

I did initially write a defensive fix (clearing the cache after every microbatch instead of once per step) before finishing this investigation, but reverted it — the caching pattern is safe as designed, and clearing more aggressively would just add a redundant AllGather per microbatch instead of one per step, undoing the point of the cache, to guard against a failure mode I could not reproduce.

If you're aware of a specific PyTorch/DTensor version or a different embedding-parallel configuration where `full_tensor()`'s autograd functions do save tensors, that would change this — happy to dig further with more specifics.

donovanclay:
## Re: "Preserve the seeded shuffle when resuming existing runs" — fixed in `f21edf144`

Confirmed, and it's a real gap I missed: `MixtureDataLoader`'s checkpoint only persists `(seed, epoch, batches_processed)` and reconstructs the reference order on resume by replaying `reshuffle()` and skipping ahead — it never persists the order itself. `6f159533c` changed `reshuffle()`'s RNG call sequence (chunked `rng.choice()` + lazily-built per-source permutations, instead of building every source's full permutation up front before any choice draw), so the same `(seed, epoch)` now produces a different sequence than it did before that commit. Resuming a pre-`6f159533c` checkpoint would silently regenerate a different epoch and skip into it at the old offset — exactly as described.

I couldn't reconcile bit-compatibility with the memory fix's actual goal: matching the old RNG sequence requires knowing every source up front to build all permutations before any choice is drawn, which is the exact eager pattern that fix removed to bound startup memory on full v9/v10 mixtures. So I went with your other suggested remedy — versioning — following the pattern this file already uses for `dataset_fingerprint` mismatches:

- Added `MixtureDataLoader.SHUFFLE_ALGO_VERSION` (currently `2`), persisted in `state_dict()`.
- `load_state_dict()` treats a missing key as version `1` (pre-`6f159533c`) and raises `RuntimeError` on any mismatch, describing exactly what would go wrong, unless the caller explicitly sets `ignore_shuffle_algo_version_mismatch=True` (then it proceeds with a `log.warning` instead).

Verified: a synthetic old-format checkpoint (no `shuffle_algo_version` key) is rejected by default, proceeds correctly under the override with a warning, and a version-matched checkpoint loads with no friction. All 24 tests in `multimodal_data_test.py` still pass; no new mypy/ruff issues.

One practical consequence worth flagging: the existing 16-GPU Stage-2 v9 run's `step8000`–`step12000` checkpoints (made earlier this session, before `6f159533c`) now need the explicit override to resume, since they predate this field — that's the intended tradeoff (a loud, actionable error instead of silent data corruption).

jason718 (commented):
Read through the whole PR. The bulk of this is solid — DynaMath reuses the shared SFT helpers correctly, `VisionBackbone` is a clean abstraction, `PackedMixtureIterableDataset` builds on `iter_dynamic_packs` rather than reimplementing the packer, and there are five new test files. My comments are concentrated in a few places.

Heads up on process: both CI workflows are `branches: [main]` only, so nothing here ran tests, lint, or type-check — the single green check is GitGuardian secret scanning. That's a branch-config gap rather than anything you did, but it means none of the below was caught mechanically.

## Blocking

**1. `mixture_data_loader.py` — the shuffle-version guard makes every existing checkpoint un-resumable, with no reachable escape hatch.**
`checkpoint_version = state_dict.get("shuffle_algo_version", 1)` returns 1 for any checkpoint written before this PR, so `load_state_dict` raises and the job dies at startup. The error text says *"Set `ignore_shuffle_algo_version_mismatch=True` to resume anyway"* — but that's a `MixtureDataLoader.__init__` kwarg, and neither `Molmo2-Stage1.py` nor `Molmo2-Stage2.py` exposes it as an `ExperimentConfig` field or passes it. There's no `--override` that reaches it, so the only recovery is editing code. The guard itself is the right call (it came out of the Codex resume thread) — it just needs a config field wired through both scripts.

**2. `nn/vision/multimodal.py` — model state-dict keys are remapped to legacy names but optimizer keys are not.**
`self.vision_backbone = VisionBackbone(...)` moves real param FQNs to `vision_backbone.*`. `state_dict()` remaps them back to `vision.*` / `connector.*`, but optimizer FQNs come from `named_parameters()` and are untouched — and this PR's own optim group overrides use `vision_backbone.connector.*` / `vision_backbone.vision.*`. So `get_model_state_dict` yields `vision.weight` while `get_optimizer_state_dict(flatten_optimizer_state_dict=True)` yields `param_groups.vision_backbone.connector.bias.lr`. Pre-PR checkpoints fail to match under `StateDictOptions(strict=True)`, and post-PR checkpoints are internally inconsistent. `load_key_mapping` exists for exactly this and is currently unused.

Related, same area: the `state_dict()` override returns a new dict, but `nn.Module.state_dict` writes into `destination` in place and parents discard the child's return value. If `MultimodalLM` is ever nested, or anyone calls `model.state_dict(destination=d)`, `d` keeps `vision_backbone.*` keys while the return value has `vision.*` — the remap is silently bypassed.

**3. Three changes reach outside Stage 2.**
- `multimodal.py` — the in-code `VIT_CROP_MICROBATCH` default went 16 → 0. `Molmo2-Stage2.py:157` explicitly says env overrides are "Applied only to `launch_config.env_vars` below, not the in-code defaults in `nn/vision/multimodal.py`: those are shared with Stage 1 and have a much wider blast radius" — but the in-code default changed anyway. Stage 1 never sets the env var, so `_encode_images` now runs all `B*T` crops through the ViT in one forward instead of 16-crop chunks, which can OOM a Stage-1 config that fits today.
- `nn/transformer/model.py` — `apply_fsdp` now reads `MM_FSDP_RESHARD_AFTER_FORWARD`. That's the shared LM path for every text-only OLMo2/OLMo3 run; with the var set (SHIP_STACK_ENV sets it unconditionally, and it's easy to leave in a shell or launch template) a text-only 32B run loses per-block resharding, roughly doubling FSDP parameter memory with nothing in its config to explain it.
- `finevision.py` — `loss_token_weighting` default flips `root_subsegments` → `root_subsegments_root_tokens`, and `_build` now passes `turns` instead of `[turns]`, so `encode_sft_example` splits multi-turn rows into per-turn branches and shuffles them. Both apply to existing image-only-v9 FineVision sources and to `VisualWebInstructDatasetConfig`. The updated assertions in `sft_datasets_test.py` (4.0 → 6.265986, 3.0 → 2*sqrt(3)) confirm the weights change. A v9 run before and after this PR trains on different supervision for identical rows, with nothing in the config to signal it.

**4. `Molmo2-Stage2.py` — `MM_FSDP_IMAGE_ALIGN_HACK=0` on every launch may deadlock multi-node.**
The justification is that DP max-crop padding makes the lockstep guard redundant, but that padding equalizes the forward, not the backward graph. `tulu4` is ~14% of image-only-v9/v10 and text-only, so a rank can draw an all-text pack: the collator supplies a dummy zero crop with `pooled_patches_idx` all -1, `valid_rows` is all-False, `image_features` is `(0, d)`, and the splice connects nothing. With the align hack off there's no other edge from connector/ViT output into the loss, so that rank runs no connector/ViT backward while its peers do — asymmetric reduce-scatter. The A/B that justified this ran at 8 GPUs; the env var is applied unconditionally, including multi-node.

## Also worth fixing

- **`Molmo2-Stage2.py`** — `_append_extra_sft_sources` is only called on the v9 `else` branch, so `mmfinereason_rate` and `finevision_rate` are silently no-ops for `single-image-only-v10`, `VALIDATION_MIXTURES_V10`, and `single-image-only-v9`. `--mixture=single-image-only-v10 --mmfinereason_rate=0.05` starts, logs a mixture with no MMFineReason, and trains the wrong distribution with no warning. Both fields are still documented as live knobs.
- **`Molmo2-Stage2.py`** — `DL_NUM_WORKERS = 2` as the default makes `--pack_sequences=false` an unconditional startup error, since `dl_num_workers > 0` requires `pack=True`. Users now have to know to pass `--dl_num_workers=0` alongside it.
- **`finevision.py`** — `image-only-v10` can't run from a clean checkout. `build_finevision_v10_config` never sets `hub_repo`, so `uses_hub()` is False and all five sources take the local-parquet branch, pointing at paths created by `launch_scripts/donovan/env/setup-finevision-v10-symlinks.sh` — which this PR adds to `.gitignore`. Same shape for DynaMath: `paths.py:9` defaults to `/weka/oe-training-default/donovanc/molmo-experimental-data`, a personal directory baked into a library default. The docstring and PR body both call this "hub-backed FineVision", which no path outside `test_finevision_hub_load` exercises.
- **`mixture_pack_profiles.py`** — four tiers labeled single-image get `pack_max_crops=25` + `shortcut_max_len_images=True` but contain multi-image sources: `demo`/`demo-pointing` include `correction_qa_multi_only_max5`, `pointing`/`demo-pointing` include `pixmo_multi_points`, and `academic` includes `mantis_instruct_*_multi_only` plus six `cosyn_multidoc_*` (image_only_v9.py:95-114). `DynamicPacker` emits those solo and the collator sizes the batch to ~125 crops — 5x the ViT memory the profile was chosen to cap.
- **`finevision.py`** — `save_finevision_index_cache` uses one fixed `<key>.tmp.npz` shared by all ranks, so 8 DP ranks building the same source race: the second `os.replace` raises an uncaught `FileNotFoundError`, and a half-written file that gets renamed raises `zipfile.BadZipFile`, which the loader's `except (OSError, KeyError, ValueError)` doesn't catch. The default dir is inside the shared read-mostly FineVision corpus root, and the new unit tests pass no `index_cache_dir`, so in CI they'd try to create `/weka/oe-training-default/...`.
- **`finevision.py`** — `require_single_image` decodes every image to count them. The comment says "Vectorized check" but it's a per-row loop, and `load_hf_dataset` never casts to `Image(decode=False)`, so building `finevision_densefusion_1m` runs 1M PNG/JPEG decodes on every rank just to evaluate `len(images) == 1`. `self._data.data.column(img_col)` (Arrow list lengths) or `with_format('arrow')` gives the count without decoding.
- **`multimodal.py`** — the embedding step cache is invalidated only by `clear_embedding_step_cache()`, which exactly one train module calls, and `MM_EMB_STEP_CACHE` defaults on. Any other loop that calls `MultimodalLM.forward` and steps the optimizer reads the first forward's gather for the rest of the process. Also `_mm_emb_full_tensor` is misnamed — `_gathered_embedding_weight` always calls `full_tensor()`, so `MM_EMB_FULL_TENSOR=0` only disables the cache, never the gather.

## Consolidation

Four things I'd like to see restructured before this lands, roughly in order of how much they'll cost later:

**Move the env knobs into the Config system.** This PR adds nine `MM_*` flags plus `VIT_CROP_MICROBATCH`, read through `os.environ` at call sites, and promotes three of them into `olmo_core/distributed/utils.py` (`fsdp_reshard_after_forward`, `fsdp_nest_connector`, `log_fsdp_topology`) — a module every training path imports. AGENTS.md makes `@dataclass Config` with dot-notation overrides the central pattern; these knobs sit outside it, so they aren't serialized into the saved config, aren't `--override`-able, and aren't type-checked. Two of the blockers above are direct consequences. Fields on `MultimodalTransformerTrainModuleConfig` / the dp config would fix the category rather than the instances.

**Collapse the two prefetch mechanisms.** `MixtureDataLoader` now carries both `prefetch_workers` (threads, `prefetch.py`) and `dl_num_workers` (processes, `PackedMixtureIterableDataset`). Of the four combinations, two raise and one silently zeroes `prefetch_workers` with a warning — and that surface is what turns the new `DL_NUM_WORKERS=2` default into the `--pack_sequences=false` trap above. A single `loader` strategy field validated in `__post_init__` would replace three scattered runtime raises.

**Parameterize v9/v10 rather than forking the module.** Measuring the two files: `_source_lookup` is 100% identical, `_LazyDatasetMap` 88%, `build_image_only_*_mixture` 82% — and the entire delta is threading one `finevision_cache_dir` kwarg. Drift has already started: v10's `_LazyDatasetMap` caches `_source_map` while v9 recomputes it on every call. Collapsing the fork removes roughly 160 of `image_only_v10.py`'s 231 lines and means the next fix lands once instead of twice.

**Derive the pack profiles.** `mixture_pack_profiles.py` is a hand-maintained name → profile table, and `MULTI_IMAGE_MIXTURE_DATASETS` already exists in `image_only_v9.py` to compute it from. The table has already drifted, which is the multi-image bug above; deriving it makes that class of mistake impossible.

Happy to pair on any of these, particularly the config migration — that one touches the most files and is the easiest to get half-done.

donovanclay:
Thanks for the thorough review — went through everything below, pushed as 7 commits (`c5f1bdc`..`208d0be`). Summary follows your structure; commit refs are short SHAs on this branch.

## Blocking

**1. `mixture_data_loader.py` shuffle-version guard unreachable.** Fixed in `208d0be`. `ignore_shuffle_algo_version_mismatch` is now an `ExperimentConfig` field on both `Molmo2-Stage1.py` and `Molmo2-Stage2.py`, passed through to `MixtureDataLoader`. Also fixed the error message itself — it named the constructor kwarg, not the CLI flag an operator actually has (`--ignore_shuffle_algo_version_mismatch=true`).

**2. `nn/vision/multimodal.py` model/optimizer key mismatch.** Fixed in `c897977`. Removed the `state_dict()`/`load_state_dict()` override pair entirely; back-compat moved to the checkpoint layer via `MultimodalLM.legacy_vision_key_mapping()` (derived from the module tree) + `swap_param_keys`. Measured the actual mismatch before fixing: 48 legacy model keys vs 528 `vision_backbone`-named optimizer keys on a tiny model. Also caught the `destination=` bypass you flagged — confirmed it was real, now moot since there's no override to bypass. Three call sites build legacy-keyed dicts and load with `strict=False`; without normalizing those too, this fix alone would have made the ViT/connector silently random-init on every HF/scratch init. `molmo2_loader.canonicalize_vision_keys()`/`strip_vision_backbone_prefix()` handle that, and are now public (`__all__` + re-exported from `olmo_core.nn.vision`) since olmo-eval's checkpoint loader needs them directly — see the "olmo-eval coordination" note below.

**3. Three changes reaching outside Stage 2.** All three fixed in `b2131a4`:
- `VIT_CROP_MICROBATCH` in-code default restored `0` → `16`.
- `Transformer.apply_fsdp` no longer reads `MM_FSDP_RESHARD_AFTER_FORWARD`; it's now an explicit `reshard_after_forward` parameter (default `None` = today's behavior). The multimodal train module passes its resolved value explicitly; the shared text-only path passes nothing. Guarded by an AST-based test so a future contributor can't quietly reintroduce the env read.
- FineVision's `loss_token_weighting`/per-turn-branching change: keeping it, deliberately, for both v9 and v10 (see `6c22310`) — decided this should move everything to mm_olmo's supervision semantics rather than preserve the old OLMo-core behavior for v9 alone. Documented on the field, in the module docstring, and pinned by a new test plus the existing loss-mask-sum assertions (confirmed they fail if `_build` is reverted to `[turns]`). Flagging explicitly per your ask: **a v9 run before/after this PR trains on different supervision for identical rows.**

**4. `MM_FSDP_IMAGE_ALIGN_HACK=0` deadlock risk.** Investigated properly rather than just reverting — `c5f1bdc`. The mechanism you flagged is real (confirmed on 2 CPU gloo processes: an asymmetric FSDP param-group gradient state does hang `backward()`, and it needs neither multiple nodes nor GPUs, just 2 DP ranks). But this model doesn't hit it: the splice always runs when `images is not None` (the collator's dummy zero crop guarantees that), and a masked `index_put` with an all-False mask still builds an autograd node, so an all-text microbatch leaves vision/connector params with *present-but-zero* gradients, not `None` — FSDP issues the reduce-scatter regardless. Verified on 2 holmes GPUs under real FSDP2 + `torch.compile` (exp `01M2622FH6WTMHZAD6A9WT5D7F`): tie off, asymmetric batches, `[10, 10]` symmetric collectives, zero `None` grads. New `align_tie_test.py` pins the invariant that makes it safe. Net call: **left the tie on** — turning it off is safe but the 8-GPU A/Bs on record show the win doesn't stack with the no-reshard win already shipped (no-reshard alone 13,281 TPS, align-off alone 13,299, both together only 13,106), so there's no throughput case for it either way.

## Also worth fixing

- `_append_extra_sft_sources` only on the v9 branch → fixed, `208d0be`. Also pinned `require_single_image=True` on the FineVision extras it builds — J-5 made this path reachable from single-image tiers, and the pack-profile derivation (below) doesn't see these runtime-appended sources, so a multi-image row here would've silently blown a single-image tier's crop budget.
- `DL_NUM_WORKERS=2` making `--pack_sequences=false` an unconditional error → fixed, `208d0be`, via the same `resolve_loader_strategy()` used for the consolidation item below.
- **`image-only-v10` can't run from a clean checkout** — partially addressed, flagging the gap rather than claiming it's closed: `paths.py`'s `MOLMO_EXPERIMENT_DATA_DIR` (the DynaMath side of this) no longer defaults to a personal directory (`6c22310`), but `build_finevision_v10_config` still doesn't set `hub_repo`, so the FineVision side of this is **not fixed** — it still takes the local-parquet branch pointing at the gitignored symlink script's output. Left open; didn't want to guess at hub config wiring without knowing which HF repo/revision you want pinned.
- `mixture_pack_profiles.py` multi-image bug → fixed and generalized, `c82f65d`. Derived the profile from the tier's actual sources (`mixtures/tiers.py`) instead of a hand-maintained table, so this class of bug can't recur. Also caught one you didn't list: `nlp-demo` had the same problem (Codex flagged it, independently).
- FineVision index-cache race → fixed, `6c22310`. Per-writer staging file instead of one shared `.tmp.npz`; `BadZipFile`/`EOFError` added to the corrupt-cache guard (an `.npz` is a zip, so a truncated one wasn't being caught); both the mkdir and the write are now best-effort so a read-only cache dir warns instead of aborting a run whose data loaded fine.
- `require_single_image` decoding every image → fixed, `6c22310`. Counts come from Arrow list offsets now, no decode, with a logged fallback if the column isn't a list type.
- Embedding step-cache invalidation / `MM_EMB_FULL_TENSOR` misnaming → fixed, `b2131a4`. Removed `MM_EMB_FULL_TENSOR` (it never gated the gather it was named for); the cache is now "armed" only by `clear_embedding_step_cache()`, so a caller that never invalidates gets a fresh gather instead of silently reusing step 0's weights forever.

## Consolidation

Did all four, in dependency order (registry before deriving profiles from it, one loader strategy before wiring the scripts to it):

- **Env knobs → Config**: `b2131a4`. All `MM_*` + `VIT_CROP_MICROBATCH` are now typed fields on `MultimodalLMConfig` / `MultimodalTransformerTrainModuleConfig`, resolved from the legacy env var once in `__post_init__` so the *resolved* value is what's serialized/overridable, not `None`.
- **Collapse the two prefetch mechanisms**: `937c9b4`. `MixtureLoaderStrategy` + `resolve_loader_strategy()` replace the three scattered constructor checks with one validated dispatch.
- **Parameterize v9/v10 instead of forking**: `c82f65d`. Shared lazy-registry/weighting plumbing extracted to `mixtures/registry.py`; this also fixed a live bug your "drift has already started" comment predicted — v10's registry cached its source lookup, v9 rebuilt it every call.
- **Derive the pack profiles**: `c82f65d`, described above.

## What's still open

- The FineVision `hub_repo` clean-checkout gap noted above.
- No full-scale training run on any of this yet — verification is CPU tests plus the one targeted 2-GPU hardware check for item 4. Recommend at least a Stage-2 smoke before merge.
- **Sequencing with olmo-eval**: `canonicalize_vision_keys()`/`legacy_vision_key_mapping()` are now a real cross-repo contract — olmo-eval's checkpoint loader needs to accept both key layouts or it breaks against any checkpoint written with this PR. That olmo-eval-side fix exists but isn't merged yet; flagging so it doesn't get missed as a merge-order dependency.
- Stage 1 doesn't go through the same loader-strategy validation as Stage 2 (harmless today since its `DL_NUM_WORKERS` defaults to 0, but noting the asymmetry).

Full test run against the pushed state: 399 passed, 1 pre-existing failure unrelated to this branch (`parity_test.py::test_clip_parity`, a `transformers` version issue reproduced identically on a clean `vision` checkout). mypy flat, no new ruff findings beyond the repo's existing baseline.

PR #837, opened 2026-08-19
Caleb/cute kda

CalebOuellette:
Adds an experimental optimized KDA kernel that is opt in only via a flag on the KDA layer. This should improve performance of train on B300 machines.

This Path mixes FLA with some custom CuTeDSL code.

Dependencies are covered by FA4 & FLA Extras. I opted to not change pyproject.toml.


AkshitaB (commented):
Lets add a benchmarking script also. 

Is the plan to move this into its own repo?

PR #852, opened 2026-09-03
Add v2 pointing, olmOCR-mix and OCR caption-tar sources to Molmo2 stage 1

uwGZQ:
## Summary

Adds mm_olmo's second-generation pointing data and an OCR data group to Molmo2 stage 1. Everything is opt-in: the default `Molmo2-Stage1.py` run is unchanged.

### Pointing v2 (`--pointing_data=v2`)

Ports mm_olmo's `PixMoPointV2` / `PixMoCountConfigV2` (`olmo/data/pixmo_datasets.py`), the pointing group of `launch_scripts/train_molmo3_stage1.py`:

* `pixmo_points_v2.py`: `PixMoPointsV2DatasetConfig` over the audited, image-grouped `pixmo-points-with-masks-v17` build (220k trainable images) and `PixMoCountV2DatasetConfig` over `count-v2` (36.9k). Each point set carries a VLM `audit_result`; failed audits are either dropped (`filter_audit`) or trained behind the `aux_pointing` / `aux_point_count` marker styles (`audit_style`). The per-image absence queries (`easy_negatives`, `paired_negatives_v2`) are sub-sampled per epoch (`n_easy_samples`, `p_paired_negatives`), per the mm_olmo team's guidance that training on all of them makes the model over-refuse. Field names follow mm_olmo's.
* `sft_formatter.py`: the `aux_*` styles render like their base style except for the style token (mm_olmo `data_formatter.py:1189`, `1824-1825`).
* `pixmo_points.py`: `_build_example` accepts per-branch loss weights (`negative_weight`).
* Stage-1 defaults follow mm_olmo's `_base_mixture`: `audit_style=(aux_point_count, aux_pointing)`, `filter_audit=False`, 2 easy negatives, `p_paired_negatives=0.25`; the v2 group is split linearly by size (mm_olmo `size_weighted=1`). Knobs are the nested `pointing_v2` / `count_v2` configs (`--pointing_v2.filter_audit=true`).

### OCR group (`--ocr_rate`, `--ocr_sources`)

* `olmocr.py`: `OlmOcrMixDatasetConfig` (mm_olmo `OlmOcrMixConfig`) over the four olmOCR-mix-1025 subsets; pages are rendered from the shipped PDFs with `pypdfium2` at a longest side sampled in 1024–2048 (1536 for eval), English-only by default, blank pages transcribed as "No text found".
* `ocr_caption_tars.py`: `OcrCaptionTarsDatasetConfig` over the oe-encoder webdataset-style tars (`<key>.jpg|png` + `<key>.json`). `TarShardIndex` scans shard headers once (~1 s/GB) and caches member offsets as `.npz` (`OLMO_CORE_TAR_INDEX_DIR`, else under `HF_DATASETS_CACHE`).
* `mixtures/ocr.py`: registry of 20 sources in three styles — `olmocr` (4 olmOCR-mix subsets, `s2pdf`, `iabooks`), `ocr_caption` (text_rich chart/diagram/doc/graphic/table, Cambrian arxivqa/ocr_vqa/screen_qa/llavar/oodvqa, TextCaps), `scene_text` (TextOCR, HierText, COCO-Text, UberText). The group's rate is split by sqrt(size) (mm_olmo's default `root_size_factor`).
* All OCR sources use mm_olmo's molmo3 `style_and_length_v3` form: the user turn is the bare `<style>:` tag.

**Overlap:** `olmocr_v6_tars` `s2pdf` / `iabooks` are the same pages as olmOCR-mix `documents` / `books` train (97.4% / 99.4% of page ids, every iabooks document; 0 overlap with the eval splits), re-rendered as JPEGs and re-transcribed. They are registered but excluded from `DEFAULT_OCR_SOURCES`; the script warns if both renderings of the same pages are selected.

### Misc

* `pixmo_cap.py`: `style_length_prefix` / `style_tag_prompt` factored out and shared. `message_sequence.py`: `encode_sft_example`'s `message_weight` annotation widened to the float it already accepted.
* `paths.py`: `PIXMO_POINTS_V2`, `OLMOCR_MIX`, `OE_ENCODER_DATA`, all env-overridable (the first and last point at other people's directories).
* The launch `post_setup` also installs `pypdfium2`.

## Usage

```
python src/scripts/train/Molmo2-Stage1.py launch RUN --pointing_data=v2 --pointing_v2.filter_audit=true
python src/scripts/train/Molmo2-Stage1.py launch RUN --ocr_rate=0.15
python src/scripts/train/Molmo2-Stage1.py launch RUN --ocr_rate=0.075 \
    --ocr_sources=[olmocr_documents,olmocr_books,olmocr_loc_transcripts,olmocr_national_archives]
```

## Verification

* New tests: `pixmo_points_v2_test.py`, `olmocr_test.py`, `ocr_caption_tars_test.py`; `src/test/data/multimodal/` passes (203 tests).
* Real-data smoke on weka with the Molmo2-4B tokenizer for every new loader: row counts match independent pyarrow scans, examples decode to the expected `<style>:` prompt and target, tar indices build in seconds and reload from cache.
* `build_config` with the new overrides round-trips through `as_config_dict` / `from_dict`; invalid values raise `OLMoConfigurationError`.
* black / isort / ruff clean on the changed files; mypy adds no new errors.

## Notes for reviewers

* At the stage-1 sequence length (2560) with 8 crops, ~4% of v2 pointing images and ~15% of olmOCR-mix `documents` pages are tail-truncated — the same token-level truncation mm_olmo's preprocessor applies (its molmo3 stage 1 runs at 3328).
* Deliberate deviations from mm_olmo: `n_easy_samples` is honoured (mm_olmo hardcodes 2); `v2_paired_negatives` defaults to True (class default; mm_olmo's arms pass False); `pdf_relpath` is split on `.tar.gz:` rather than the last colon. Not ported: `CocoTrain`, `TextRichCaptionConfig`, PixMoPointV2's mask / 3D / surface-normal legs.
* **Correction (review):** an earlier version of this description called `CocoTrain` a COCO detection-as-pointing source. It is not — under `style="pixmo"` it emits `pixmo_seg` with `bboxes` + RLE `segmentations` and no points on positives (`academic_datasets.py:2175-2189`, `data_formatter.py:965-968`). It is a segmentation source, so it stays unported for the same reason the mask branches do.
* TextCaps' `caption` in the tars is its five reference captions concatenated; it stays in the default group but is the first source to drop if that target form is unwanted.

---

## Review round 2 (`523c388f3`)

Addresses [Jason's review](https://github.com/allenai/OLMo-core/pull/852#pullrequestreview-5125932192) of `a6d307f91`. All six issues are fixed; `origin/vision` is merged in (through #861).

**Changes what the model trains on**

* **Negatives rotate across epochs.** `example_rng` was mm_olmo's derivation with `epoch` pinned to 0, and neither data loader handed an epoch to its sources, so `n_easy_samples` / `p_paired_negatives` redrew one fixed subset forever. `example_rng` gains mm_olmo's `+ len(dataset) * epoch` term (`dataset.py:70-73`), a shared `EpochSeededExamples` mixin derives per (row, epoch) streams, and both loaders call `set_epoch` from `reshuffle`. Measured on v17: the paired-negative pool is **67.8%** covered after the four epochs the pointing group runs, against the **25.0%** a pinned epoch repeats. Epoch 0 is bit-identical to before.
* **`PixMoCountV2Dataset.keep`** drops blank / null labels and null points, and its row filter mirrors it.
* **`_size_fractions`** rejects an empty source by name rather than giving it weight 0 (silently dropped by the loader) or dividing by zero into NaN.

**Aborts the run**

* **`PixMoPointsV2Dataset.keep`** rejects a null `points`, matching the row filter that tolerates one.
* **Tar index keys are unicode**, the cache filename carries a format version so v1 caches are ignored, and an unreadable cache is rebuilt rather than fatal.
* **Both OCR sources go through `sft_common.get_example_with_skip`.** The two tests that asserted the raise now assert the skip, as predicted.

**Measured exposure — four of the six are latent, not currently firing**

Worth stating plainly, since it changes how urgent these are rather than whether they were worth fixing:

| Issue | Occurrences in the current data |
|---|---|
| #1 negatives never rotate | fires on the default config (25.0% → 67.8% pool coverage) |
| #2 blank labels (count-v2) | 0 blank / 0 null label / 0 null points in all 3 splits (36,916 rows each) |
| #3 zero-size source | only on a misconfigured path |
| #4 null `points` (v17) | 0 of 2,333,432 annotations |
| #5 non-ASCII tar keys | 0 in 13,884 samples across 6 sources (one shard each) |
| #6 blank targets | 0 blank captions in the same sample |

The invariants should still hold — #4 is two filters that must agree, and #5 crashes at mixture-build time where there is no error budget — but only #1 changes what a run does today.

**Scoping answers**

* **`figure_ocr`:** confirmed. The tar `caption` is byte-identical to the HF build's `mid_level` (`chart/HTMLChartPipeline_area_0-0`: high / mid / low = 145 / 645 / 2262 chars), so `low_level` is absent and this group carries roughly a fifth of `figure_ocr`'s caption mass. Recorded in `mixtures/ocr.py`, including that `--ocr_rate` with `DEFAULT_OCR_SOURCES` is *not* mm_olmo's OCR mixture. The three-level loader is yours.
* **`CocoTrain`:** your reading is right and mine was wrong; corrected above and in the code comment.
* **The `include_masks` denominator:** quantified. mm_olmo's `_keep` admits **220,314** rows against our **220,291** (+23, +0.010%) — the mask path only rescues annotations above `max_points` — so the linear split inside the v2 group is unchanged for practical purposes, and 220k is the right denominator for a port with no segmentation path. The larger consequence is elsewhere and now documented: mm_olmo routes a negative to segmentation with probability `1 - zero_points_as_segmentation` (0.5), for *all* negatives, so this port shows **~2x mm_olmo's pointing-refusal density at equal knobs**. Halve `n_easy_samples` / `p_paired_negatives` to match it.

**From the findings you set aside**

* Shard indexing was quadratic (`stem not in order` scanned a list): 20k-sample shard **5.80s → 0.77s**, 40k **25.2s → 1.2s**.
* `--ocr_tars.dataset_path` / `.style` / `.strip_text_tags` were accepted, saved into the run config and then overwritten per source. Refused now, with `--ocr_data_root` as the real knob.
* Index temp files carry a uuid, not just the pid, and are cleaned up on a failed write (also the Codex comment).

Still open from your list: the index cache is not rank-0 gated, so every rank scans and writes the same entry once per fresh filesystem. Correct but wasteful; happy to take the rest of that list separately.

**Verification:** 220 tests in `src/test/data/multimodal/` pass; black / isort / ruff clean on the changed files; mypy adds no errors over `a6d307f91`. Real-data checks re-run for both pointing-v2 sources and four tar sources.

### DIFF
--- a/src/olmo_core/data/multimodal/__init__.py
+++ b/src/olmo_core/data/multimodal/__init__.py
@@ -30,6 +30,12 @@ from .mmfinereason import (
     MMFineReasonDatasetConfig,
     extract_answer_text,
 )
+from .ocr_caption_tars import (
+    OcrCaptionTarsDataset,
+    OcrCaptionTarsDatasetConfig,
+    TarShardIndex,
+)
+from .olmocr import OlmOcrMixDataset, OlmOcrMixDatasetConfig
 from .packing import pack_examples
 from .pixmo_cap import PixMoCapDataset, PixMoCapDatasetConfig
 from .pixmo_points import (
@@ -40,6 +46,12 @@ from .pixmo_points import (
     PixMoPointsDataset,
     PixMoPointsDatasetConfig,
 )
+from .pixmo_points_v2 import (
+    PixMoCountV2Dataset,
+    PixMoCountV2DatasetConfig,
+    PixMoPointsV2Dataset,
+    PixMoPointsV2DatasetConfig,
+)
 from .sequence_builder import (
     ATTEND_ALL_SUBSEGMENT_ID,
     build_branched_sequence,
@@ -48,7 +60,10 @@ from .sequence_builder import (
 from .paths import (
     ACADEMIC_DATASETS,
     MOLMO_DATA_DIR,
+    OE_ENCODER_DATA,
+    OLMOCR_MIX,
     PIXMO_DATASETS,
+    PIXMO_POINTS_V2,
     TORCH_DATASETS,
     TULU4_DATA,
 )
@@ -77,6 +92,15 @@ __all__ = [
     "PixMoCountDatasetConfig",
     "CoSynPointDataset",
     "CoSynPointDatasetConfig",
+    "PixMoPointsV2Dataset",
+    "PixMoPointsV2DatasetConfig",
+    "PixMoCountV2Dataset",
+    "PixMoCountV2DatasetConfig",
+    "OlmOcrMixDataset",
+    "OlmOcrMixDatasetConfig",
+    "OcrCaptionTarsDataset",
+    "OcrCaptionTarsDatasetConfig",
+    "TarShardIndex",
     "Tulu4Dataset",
     "Tulu4DatasetConfig",
     "AcademicDataset",
@@ -92,8 +116,11 @@ __all__ = [
     "SubMixture",
     "compute_flat_mixture_weights",
     "PIXMO_DATASETS",
+    "PIXMO_POINTS_V2",
     "TULU4_DATA",
     "ACADEMIC_DATASETS",
+    "OLMOCR_MIX",
+    "OE_ENCODER_DATA",
     "MOLMO_DATA_DIR",
     "TORCH_DATASETS",
     "MultimodalCollator",
--- a/src/olmo_core/data/multimodal/data_loader.py
+++ b/src/olmo_core/data/multimodal/data_loader.py
@@ -80,6 +80,10 @@ class MultimodalDataLoader(DataLoaderBase):
         if epoch is not None:
             self._epoch = epoch
         epoch = self._epoch if self._epoch is not None else 1
+        # Hand the epoch to a source that samples per example, as MixtureDataLoader does.
+        set_epoch = getattr(self.dataset, "set_epoch", None)
+        if callable(set_epoch):
+            set_epoch(epoch)
         order = np.arange(len(self.dataset))
         if self.shuffle:
             np.random.RandomState(self.seed + epoch).shuffle(order)
--- a/src/olmo_core/data/multimodal/message_sequence.py
+++ b/src/olmo_core/data/multimodal/message_sequence.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from typing import Dict, List, Optional, Sequence, Tuple
+from typing import Dict, List, Optional, Sequence, Tuple, Union
 
 import numpy as np
 
@@ -29,7 +29,7 @@ def encode_sft_example(
     p_high_res: float = 0.0,
     high_res_max_crops: int = 24,
     loss_token_weighting: str = "root_subsegments_root_tokens",
-    message_weight: Optional[MessageWeight] = None,
+    message_weight: Union[None, float, MessageWeight] = None,
     seed: int = 0,
     shuffle_rng: Optional[np.random.RandomState] = None,
 ) -> Dict[str, np.ndarray]:
--- a/src/olmo_core/data/multimodal/mixture_data_loader.py
+++ b/src/olmo_core/data/multimodal/mixture_data_loader.py
@@ -134,6 +134,14 @@ class MixtureDataLoader(DataLoaderBase):
         if epoch is not None:
             self._epoch = epoch
         epoch = self._epoch if self._epoch is not None else 1
+        # Sources that *sample* inside ``__getitem__`` (a few of an image's negatives, a render
+        # size) need the epoch, or every epoch redraws the same subset and the rest of each pool
+        # is never trained on: reshuffling references alone cannot rotate a within-example draw.
+        # See :class:`~olmo_core.data.multimodal.sft_common.EpochSeededExamples`.
+        for dataset in self.datasets:
+            set_epoch = getattr(dataset, "set_epoch", None)
+            if callable(set_epoch):
+                set_epoch(epoch)
         rng = np.random.RandomState(self.seed + epoch)
         # Number of example refs to draw. When packing, an epoch consumes ~all examples
         # (several per packed sequence), so draw a full epoch of examples; otherwise draw
--- a/src/olmo_core/data/multimodal/mixtures/ocr.py
+++ b/src/olmo_core/data/multimodal/mixtures/ocr.py
@@ -1,0 +1,137 @@
+"""The OCR source group for Molmo2 stage-1 (``Molmo2-Stage1.py --ocr_rate``).
+
+Twenty image -> free-text sources of three kinds, each a separate dataset sharing the group's
+rate (split by sqrt(size), mm_olmo's default ``root_size_factor``):
+
+* **page transcription** (style ``olmocr``): the four olmOCR-mix-1025 subsets
+  (:class:`~olmo_core.data.multimodal.olmocr.OlmOcrMixDatasetConfig`, rendered from PDFs), plus
+  the oe-encoder ``olmocr_v6_tars`` ``s2pdf`` / ``iabooks`` (pre-rendered JPEGs, re-transcribed).
+* **text-rich captions** (style ``ocr_caption``): synthetic charts / diagrams / documents /
+  graphics / tables, the Cambrian OCR-heavy subsets (arxivqa, ocr_vqa, screen_qa, llavar, oodvqa)
+  and TextCaps -- one dense natural-language caption per image.
+* **scene text** (style ``scene_text``): TextOCR, HierText, COCO-Text and UberText, whose target
+  is the text visible in the photo.
+
+``s2pdf`` and ``iabooks`` are the SAME pages as olmOCR-mix ``documents`` / ``books`` train
+(97.4% / 99.4% of their page ids, and every one of their documents; none of the eval pages), only
+rendered and transcribed by a different pipeline. They are registered so either rendering can be
+chosen, but :data:`DEFAULT_OCR_SOURCES` leaves them out so a page is not counted twice.
+
+TextCaps' ``caption`` is its five reference captions concatenated into one string (``n_refs``),
+which is what the tars ship; it stays in the default group as a caption source but is the one to
+drop first if that target form is unwanted.
+"""
+
+from __future__ import annotations
+
+import os
+from dataclasses import dataclass
+from typing import Dict, Tuple
+
+from olmo_core.data.multimodal.ocr_caption_tars import OcrCaptionTarsDatasetConfig
+from olmo_core.data.multimodal.olmocr import OlmOcrMixDatasetConfig
+from olmo_core.data.multimodal.paths import OE_ENCODER_DATA
+from olmo_core.exceptions import OLMoConfigurationError
+
+__all__ = [
+    "OcrTarSource",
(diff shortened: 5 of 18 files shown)


jason718 (commented):
Reviewed `a6d307f91`. The port itself looks careful — I checked the pointing-v2 field semantics against mm_olmo's `PixMoPointV2`, and the audit / `aux_*` handling, the negative sub-sampling knobs and the measured s2pdf/iabooks overlap all line up.

Keeping this high-level and to two things: **what's still missing against mm_olmo's stage-1 data**, and **six issues I'd want fixed before this trains anything**.

---

## 1. Missing vs mm_olmo's stage-1 pointing / counting / OCR

### Open: `figure_ocr` is only ~20% ported

The `text_rich_*` tar sources are the same *images* as mm_olmo's `figure_ocr` group, but the `*_v6_tars` build carries only the **`mid_level`** caption. mm_olmo's HF build (`molmo3_datasets/text_rich_caption/hf`) ships three levels and emits **all three as branches** (`text_rich_caption_datasets.py:173-183`).

Caption lengths in chars, two rows sampled per category:

| category | `high_level` | `mid_level` | `low_level` | mid as % of row total |
|---|---|---|---|---|
| chart | 145 / 227 | 645 / 788 | 2262 / 2051 | ~22% |
| doc | 149 / 162 | 900 / 823 | 2321 / 2970 | ~26% |
| table | 121 / 157 | 842 / 843 | 2171 / 4043 | ~19% |

For `chart/HTMLChartPipeline_area_0-0` the tar's `caption` field is byte-identical to `mid_level`. So the dropped `low_level` — consistently 2–4k chars, the dense read-out of everything on the page — is the most OCR-relevant part of the group, and it's absent.

**Net effect:** `--ocr_rate=0.15` with `DEFAULT_OCR_SOURCES` is not mm_olmo's OCR mixture. It's olmOCR-mix, plus a fifth of `figure_ocr`, plus this repo's extra Cambrian / TextCaps / scene-text sources.

I have a 3-level loader for this and would rather send it as a follow-on than grow this PR.

### Descoped on our side — please don't build these

`CocoTrain` and `PixMoPointV2`'s mask branches. Both are segmentation, and olmo-core has no segmentation path anywhere.

Worth recording *why* `CocoTrain` can stay unported, because your PR comment and my own first read both had it wrong: with `style="pixmo"` it emits **`pixmo_seg`** and its message dicts carry `bboxes` + `segmentations` (RLE) with **no points at all** (`academic_datasets.py:2175-2189`), and `pixmo_seg` is in `SEGMENTATION_STYLES` (`data_formatter.py:965-968`). It's a segmentation source, not detection-as-pointing.

### One consequence of that worth checking

mm_olmo's `_base_mixture` leaves `include_masks` at its class default of `True`, so its `_keep` also admits annotations through the masks path (`pixmo_datasets.py:1943-1945`), and `zero_points_as_segmentation=0.5` routes about half the zero-point annotations to segmentation.

Its filtered row count is therefore **larger** than your 220k. Because the v2 group splits *linearly* by size, that difference moves the weights inside the group — so it's worth confirming 220k is the denominator you want the linear split computed from, rather than one inherited from a config that assumed masks were on.

---

## 2. Critical — silently changes what the model trains on

These three produce no error. They're the ones I'd prioritise.

**1. Negatives never rotate across epochs** — `pixmo_points_v2.py:330`
*(credit to the Codex review for spotting this one)*

`example_rng(seed, i)` is mm_olmo's derivation with `epoch` pinned to 0 — its own docstring says "epoch 0" — and `MixtureDataLoader.reshuffle()` only permutes `(source, index)` references, so no epoch ever reaches the dataset. mm_olmo folds it into the seed (`dataset.py:68-75`).

With `p_paired_negatives=0.25`, that means **75% of each image's hard negatives are never trained on**, rather than being seen roughly once per four epochs — and the pointing group runs four. The sub-sampling knobs don't currently do what their docstrings say.

To be fair: the pin is pre-existing in `example_rng` and v1 shares it. But v1 had no negative pool to rotate, so this PR is where it starts changing the training distribution.

**2. Blank labels train silently** — `pixmo_points_v2.py:415`

`PixMoCountV2Dataset.keep` has none of the guards the points path has, and its `_keep_flat` is literally `np.ones(...)`. A blank `label` doesn't crash — it trains `"pointing:"` against an empty-named `<points>`. (`label=None` and `points=None` do crash, on `label.lower()` and the reshape respectively.)

**3. A zero-size source silently gets weight 0** — `Molmo2-Stage1.py:849`

`_size_fractions` divides by `frac.sum()` with no guard. All-zero sizes give `[nan nan nan]` and then `probabilities contain NaN`, but the quieter case is worse: one empty source inside a larger group gets weight 0 and is never sampled, with no warning. `mixture_weights._dataset_size_factor` already carries the `max(dataset_len, 1)` guard this copy drops.

---

## 3. Critical — aborts the run

**4. Null `points` crashes rows the index admitted** — `pixmo_points_v2.py:215`

`keep` calls `len(anno["points"])` unguarded, while `_keep_flat` deliberately tolerates a null `points` (mapping it to `-1` via `nan_to_num`). So a row with one good annotation plus one `points=None` passes the row filter and then raises `TypeError` in `format_row`. The loader absorbs it as a data error, but a systematic null pattern walks into `max_consecutive_data_errors=10`.

**5. Non-ASCII tar keys kill startup** — `ocr_caption_tars.py:152`

`np.bytes_` encodes str via the ASCII codec, so a single non-ASCII member key raises `UnicodeEncodeError`. This fires from `__init__`, i.e. inside `_build_mixture_sources` — where the loader's per-example error tolerance doesn't apply — so the job dies at startup *after* paying the full shard-scan cost. Keys here derive from filenames (iabooks, non-English scene text, TextCaps), so it's reachable.

**6. Expected-blank targets burn the error budget** — `ocr_caption_tars.py:304`

Blank captions, `<text></text>`, and truncate-to-no-loss-tokens all raise out of `__getitem__` instead of going through `sft_common.get_example_with_skip` (which exists for exactly this, and is used by `finevision` / `mmfinereason`). The scene-text sources genuinely contain images with no legible text, so blank targets are *expected data*, not corruption — 10 in a row inside one shard aborts the run. Note the new tests currently assert the raise, so they'd flip with the fix.

---

I also have a longer list of performance and design findings — shard indexing is O(n²) (measured 3.1 s for a 20k-sample shard, and it's GIL-bound so the thread pool can't help); the index cache never survives a Beaker job and isn't rank-0 gated; `--ocr_tars.dataset_path` is accepted, recorded in the saved config, and then ignored. None of them change what gets trained, so I've left them out here — happy to send them separately if useful.

jason718:
Superseded by #875. Roy has left the team, so the work continues there on `jasonr/stage1-pointing-v2-ocr`, which starts from this branch's head (`523c388f3`, including all the review fixes) and adds two commits: the SFT-stage prompt options are removed so the new sources are stage-1 only, and every new source is audited and guarded so only training splits are read. The review history here stays as the record for the six fixes.

PR #862, opened 2026-09-11
Add cu129 / torch-2.13 Beaker image and move GPU CI to it

AkshitaB:
## Summary

Adds a CUDA 12.9 / torch 2.13 Beaker image and moves GPU CI onto it. This lets `flash-linear-attention` be un-pinned back to 0.5.2 — 0.5.2 refuses its gated-delta-net backward on Hopper unless triton ≥ 3.7.1, which comes with torch 2.13.

## Image

- New `beaker-image-cu129` / `-cu129-rma` / `-cu129-sm80` Makefile targets: torch 2.13.0, CUDA 12.9.1, TransformerEngine 2.18. The CUDA-13 family and the default build are also on torch 2.13 now.
- The `sm80` variant builds for sm_80/90/100 on a CUDA devel base, so it runs on A100, H100, and B200, and imports TransformerEngine (the devel base provides the system libcudart TE loads at import).
- TE 2.18's NCCL expert-parallel extension is disabled at build (`NVTE_WITH_NCCL_EP=0`) — olmo-core doesn't use TE's NCCL-EP API, and it fails to compile against the bundled NCCL.
- `flash-linear-attention` un-pinned to 0.5.2; `dion` bumped to `e64832041` for torch-2.13 compatibility.
- The rowwise MXFP8 quant kernel's per-warp staging (`ROWWISE_MXFP8_Q_CHUNK_BLOCKS`) is reduced 128 → 32 so it fits sm_100's 48 KB static shared-memory cap — at 128 the kernel needs ~80 KB and nvlink rejects it on B200 ("uses too much shared data"). sm_90 is unaffected.

## CI

- GPU test matrix moved to the cu129 images: `sm80` is the universal image for the general jobs and the attention job; `rma` serves the **MoE kernels** job (needs NVSHMEM for the `symm_mem_vdev2d` EP kernels).
- Jobs target explicit `--cluster` lists rather than `--gpu-type` (jupiter/ceres = H100, saturn = A100, titan = B200), so scheduling only lands on clusters this budget can use. Every GPU job can run on B200; the attention and MoE-kernels jobs run on H100 + B200; the general jobs also on A100.
- `torch` is pinned `<2.14` and the CPU-CI setup action defaults to torch 2.13, so CI tests a torch the images actually ship (previously `.[all]`'s unbounded fla floor floated CPU CI up to 2.14).
- Dropped the separate torch-2.11 job (the universal image now covers all GPU types).

## Test gating for the new arch reality

- TE fused attention and FP8-e4m3 tests gated to compute capability ≥ 9 — they self-skip on A100 (sm_80).
- TE MXFP8 comparison gated to cc ≥ 10 (Blackwell).
- dion HSDP is skipped in tests on torch ≥ 2.13 (broken upstream two ways: a deterministic CPU-tensor→Triton error in the replica-only path, and a flaky ALLREDUCE hang in the sharded path). Since the default image is now torch 2.13, `Dion.build_parallelism_config` also raises a clear `RuntimeError` when HSDP is configured on torch ≥ 2.13, instead of erroring or hanging deep in the optimizer step.
- The flaky te-attn Ulysses CP tests (attention + transformer) are xfailed (`strict=False`) so those jobs stay green whether the test flakes or passes.

Tests on all GPU types:

| GPU  | cluster        | image                                             | link                                          |
|------|----------------|---------------------------------------------------|-----------------------------------------------|
| h100 | jupiter/ceres  | akshitab/olmo-core-tch2130cu129-rma-2026-09-11    | https://beaker.org/ex/01M2938RQ8KG44NX6TZC9XXR1Z |
| b200 | titan          | akshitab/olmo-core-tch2130cu129-rma-2026-09-11    | https://beaker.org/ex/01M2938M48E9WZ764VPGA1S8KG |
| a100 | saturn         | akshitab/olmo-core-tch2130cu129-sm80-2026-09-11   | https://beaker.org/ex/01M2938PMSXY5ZS67M6NZ9GDDZ |
| b300 | holmes         | akshitab/olmo-core-tch2130cu130-fa4-rma-2026-09-11 | https://beaker.org/ex/01M2945AZD1DTVPHS5NGK2GTE0 |


.github/workflows/main.yml
```diff
@@ -8,6 +8,8 @@ on:
   pull_request:
     branches:
       - main
+      # TEMP: run CI on PRs into the integration branch; revert before merge.
+      - akshitab/moe-v2-core
```
gregorybchris, line 12: Reminder to remove this

.github/workflows/main.yml
```diff
@@ -221,10 +207,10 @@ jobs:
             gpus: 2
             # The v1 dropless MoE tests use the grouped_gemm CUTLASS kernels, which have no
             # Blackwell (sm_100) build, so keep this job off b200.
-            gpu_types: --gpu-type h100 --gpu-type a100
-            # The v2 MoE GPU tests need TransformerEngine + torch 2.10, which this image
-            # lacks, so they run in the "Test MoE kernels" job below; exclude them here.
-            # (The fused block/model tests moved to src/test/nn/ddp, which this job's
+            clusters: --cluster ai2/jupiter --cluster ai2/ceres --cluster ai2/saturn --cluster ai2/titan
```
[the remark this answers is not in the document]
gregorybchris: This sounds right to me

gregorybchris (approved):

.github/workflows/main.yml
```diff
@@ -164,54 +166,42 @@ jobs:
...
-            image: akshitab/olmo-core-tch2110cu128-2026-08-28
+            image: &gpu_test_image akshitab/olmo-core-tch2130cu129-sm80-2026-09-11
             gpus: 2
-            gpu_types: --gpu-type h100
+            # The catch-all job runs a larger slice, so give it 30m on both the GH Actions step and
+            # the beaker run cap (other jobs default to 15m).
+            timeout_minutes: 30
+            task_timeout: 30m
+            clusters: --cluster ai2/jupiter --cluster ai2/ceres --cluster ai2/saturn --cluster ai2/titan
             run: |
```
abhishekraok, line 181: This job now runs `src/test/nn/ddp` (the new image has TE, so those tests no longer skip), which is also the bulk of the MoE-kernels job below. Run 34648263207 on B200 reached 58% before the 20m Beaker cap; ~35m at that pace, so 30m will still miss. Suggest adding `--ignore-glob='src/test/nn/ddp*'` here (the dropped torch-2.11 job had it) and reverting the timeout bumps.

pyproject.toml
```diff
@@ -72,19 +72,20 @@ fa4 = [
...
-# but no released torch bundles triton >= 3.7.1 (torch 2.10/2.11 pin 3.6.0, 2.12 pins 3.7.0) and
-# the CI/beaker image ships torch 2.10 / triton 3.6.0. Stay on 0.4.1 until the GPU image moves to a
-# torch that bundles triton >= 3.7.1 (or a validated forced-triton image), then match main's 0.5.2.
-fla = ["flash-linear-attention==0.4.1"]
+# NOTE: fla 0.5.2 requires triton >= 3.7.1 (a hard guard on Hopper). torch 2.13 bundles triton
+# 3.7.1 (and nccl 2.29.7 + a matching torchaudio), so it works on the CUDA-12.9 / torch-2.13 image
+# (see `beaker-image-cu129` in the Makefile). torch 2.10/2.11 ship triton 3.6.0 and torch 2.12 ships
+# 3.7.0, none of which satisfy the guard -- so the torch>=2.13 pin below enforces the floor. Since
+# `all` pulls in `fla`, this also makes `.[all]` require torch>=2.13.
+fla = ["flash-linear-attention==0.5.2", "torch>=2.13.0"]
```
abhishekraok: No ceiling here, so the CPU `Test` job's `uv pip install -e .[all]` upgrades the torch 2.10 CPU wheel the action just installed to `torch==2.14.0` from PyPI (visible in the job log: `- torch==2.10.0+cpu` / `+ torch==2.14.0`). CPU CI is therefore testing a torch no image ships and the action's `torch-version` input is dead. `torch>=2.13.0,<2.14` here, or make the action install 2.13 CPU and constrain the second install, would keep CPU and GPU CI on the same torch.

src/olmo_core/launch/beaker.py
```diff
@@ -139,17 +139,14 @@ class OLMoCoreBeakerImage(StrEnum):
     """
 
     # NOTE: when updating default images here, should also update images used in tests at .github/workflows/main.yml
-    stable = f"akshitab/olmo-core-tch{_DEFAULT_TORCH}cu{_DEFAULT_CUDA}-sm80-2026-08-25"
+    stable = f"akshitab/olmo-core-tch{_DEFAULT_TORCH}cu{_DEFAULT_CUDA}-sm80-2026-09-10"
```
abhishekraok: CI tests `akshitab/olmo-core-tch2130cu129-sm80-2026-09-11`; this is the 09-10 build. The NOTE above asks for these to match.

src/olmo_core/kernels/cuda/olmo_symm_mem_common.cuh
```diff
@@ -51,7 +51,7 @@
 #define ROWWISE_COMBINE_FUSED_VECS_PER_THREAD 16
 #define ROWWISE_WEIGHTED_PUT_ELEMS_PER_THREAD 32
 #define ROWWISE_MXFP8_BLOCK_SIZE 32
-#define ROWWISE_MXFP8_Q_CHUNK_BLOCKS 128
+#define ROWWISE_MXFP8_Q_CHUNK_BLOCKS 32
```
abhishekraok, line 58: This quarters `shared_q` per warp in `dispatchRowsPutScaledWeighted` (32*32 = 1 KiB vs 4 KiB) and multiplies the scale-group loop iterations by 4. The commit message is "reduce block size" and the PR description does not mention it. What forced it (ptxas 12.9 shared-memory limit on the sm_100 build?), and is there a before/after number for the rowwise EP path? If it is a real tuning change it probably wants its own PR.

src/Dockerfile
```diff
@@ -140,8 +140,12 @@ RUN if [ -n "${FLASH_ATTN_4_VERSION}" ]; then \
     fi
 
 # Install transformer-engine.
+# NVTE_WITH_NCCL_EP=0 disables TE's own NCCL expert-parallel extension (csrc/extensions/ep.cpp).
+# We don't use it (olmo-core has its own EP: symm_mem_vdev2d / nccl_rma_p2p), and on a devel/RMA
+# base it otherwise compiles against the system NCCL headers — TE >= 2.18 needs a newer NCCL
+# (ncclWindow_vidmem) than the image ships, breaking the build. Only TE's MXFP8 path is needed.
 ARG TE_VERSION=2.9
```
abhishekraok: Default is still 2.9 while the Makefile passes 2.18.0; harmless via `make`, but a bare `docker build` gets a TE that the new `NVTE_WITH_NCCL_EP=0` comment above does not describe.

abhishekraok (commented):
_Review written by Claude Code (Fable 5.1) on behalf of Abhishek Rao (@abhishekraok)._

Read the full diff and the CI history for the branch. The image/pin side looks right and CI is green everywhere except **Test (GPU)**, which has not passed on any commit of this branch. Findings, most important first; inline comments carry the detail for the first five.

1. **Test (GPU) does not fit its budget, and raising timeouts will not fix it.** Run 34648263207 (43f3fa9) landed on B200 and was killed by the 20m Beaker task-timeout at 58% of the 199 selected tests (`job has exceeded its timeout of 20m0s` on job 01M2953C8XJRBPR5KVXQCVQ8GY). At that pace the slice needs ~35m, past the new 30m too. The cause is that `src/test/nn/ddp` now executes in this job (the old sm80 image had no TE, so it skipped) *and* in the MoE-kernels job. Ignoring `nn/ddp` here restores the pre-PR split and should bring the job back under 15m.
2. **CPU CI now tests torch 2.14, not 2.13.** The `Test` job installs torch 2.10 CPU, then `.[all]` hits the new `torch>=2.13` floor (no ceiling) and replaces it: `- torch==2.10.0+cpu / + torch==2.14.0` (run 34648263207). The action's `torch-version` input is now dead and the CPU suite runs a torch no image ships.
3. **Dropping the temp trigger also dropped CI for this PR.** Since 5095699 the workflow no longer fires for PRs into `akshitab/moe-v2-core`, so the 30m bump has no run and none of the later commits will get one. Worth either re-targeting to `main` for a final run or verifying the Test (GPU) fix manually before merge.
4. **A kernel change is riding in an image PR.** `ROWWISE_MXFP8_Q_CHUNK_BLOCKS` 128 -> 32 (commit "reduce block size") quarters the per-warp staging buffer in the MXFP8 rowwise dispatch kernel and is not mentioned in the description. It looks like a ptxas 12.9 / sm_100 shared-memory workaround; it deserves a rationale and a perf number, or its own PR.
5. **`OLMoCoreBeakerImage.stable` (sm80 2026-09-10) is not the image CI tests (sm80 2026-09-11)**, despite the comment above it.
6. **The cu130 family has no CI coverage.** `stable_cu130` and the `olmo-ddp` preset now point at `tch2130cu130-fa4-rma-2026-09-11`, but every job runs cu129; the only validation of that image is the build-time import check. Since it is the B300 image, a manual holmes smoke run would be reassuring.
7. **Description is stale on placement.** It says bulk jobs keep `--gpu-type h100 --gpu-type a100` and attention is pinned to h100; the diff uses `--cluster` lists and attention ran on B200 (titan) in the last two runs. Likewise the "keep this job off b200" comment on Test MoE contradicts titan being in its list, and the job passed on B200 both times, so the comment is what is wrong.
8. Smaller: `xfail(strict=False)` on the TE Ulysses tests has no tracking issue, so a real regression there is invisible; Dion+HSDP now hard-fails on the default image (better than a hang, but worth a line in the description); stale defaults `TE_VERSION=2.9` in `src/Dockerfile` and `torch-version: '2.10.0'` in the setup action.

Of the earlier Codex comments, most are resolved by later commits; still open are the stable-tag mismatch (5), the B200 comment (7), and three low-priority ones (`stable_cu128` alias removed, no A100 entry in `run_ci_tests_on_images.py`, the Blackwell-unconditional xfail).

PR #863, opened 2026-09-11
Caleb/kda cute kernels

CalebOuellette:
This PR Cherry picks the KDA layer from MOE branch & adds new experiment kernels from Kernel-Fun packages. Need to both install the package and opt in via the flag to uses these kernels otherwise everything falls back to FLA.


src/olmo_core/nn/convolution.py
```diff
@@ -25,6 +25,7 @@ def __init__(
         dtype: torch.dtype | None = None,
         init_device: str = "cpu",
         activation: Literal["silu", "swish"] | None = "silu",
+        use_experimental_kernels: bool = False,
```
YashasSamaga, line 28: If `kernel-fun` is intended to be an unmaintained one-off package, should the details make their way into public API? Or do we see `use_experimental_kernels` to be generic enough that it's not committing to any specific package and just loosely means max performance.

YashasSamaga (approved):

PR #876, opened 2026-09-20
Stage 1: caption prompts are the style tag alone, matching test time

jason718:
Stage-1 caption training used a prompt form that neither the reference implementation trains on nor the eval sends. This makes training use the same form as test time: **the style tag alone**.

## The mismatch

| | Caption user turn |
|---|---|
| OLMo-core stage-1 training, before | `long_caption 63: Describe this image.` |
| mm_olmo stage-1 training | `long_caption 63:` |
| Stage-1 caption eval (olmo-eval `dense_caption_captioner`) | `long_caption 65:` |
| OLMo-core stage-1 training, **after** | `long_caption 63:` |

* mm_olmo's stage-1 scripts (`train_captioner.py`, `train_molmo3_stage1.py`) leave `DataFormatter.prompt_templates` at its `"none"` default. Under `"none"` a caption's question is the empty string, so the user turn is just the `style_and_length` tag.
* `PixMoCapDataset` appended an instruction sampled from `CAPTION_PROMPTS` / `TRANSCRIPT_PROMPTS` after the tag. The model therefore always saw a sentence in training and never sees one at test time.

## The change

In the conditioned form (`style_length_conditioning=True`, what `Molmo2-Stage1.py` runs), the user turn of both the caption and the transcript branch is now the tag and nothing else: `long_caption 63:` / `transcript 41:`, or the bare `long_caption:` the 10% of the time the length is dropped.

Unchanged: `fixed_prompt`, and `style_length_conditioning=False`, which still gives the un-tagged natural-language prompt. Pointing, counting and text-only sources are not touched.

One consequence to be aware of: the per-example RNG stream shifts by the single draw that used to pick the sentence, so the length bucket for a given index differs from earlier runs. The distribution is the same.

## Verification

* `src/test/data/multimodal/`: **158 passed** on CPU (torch 2.11). Local run, since CI does not cover PRs into `vision`.
* New test asserts that no conditioned prompt, over 24 examples and both branches, contains any pool sentence, and that the length bucket still appears. The existing conditioning test now matches the exact tag form with a regex.
* Rendered through the real code path on synthetic rows: `['long_caption 10:', 'transcript 12:']`, `['long_caption:', 'transcript 2:']`.
* `ruff` clean. **No training run yet**, so the effect on caption metrics is unmeasured. Any comparison against earlier stage-1 checkpoints should account for the prompt change.


jason718:
Folded into #875 at the owner's request, so the prompt change and the new data land together. The commit is cherry-picked there unchanged (`7b673ae0e` on `jasonr/stage1-pointing-v2-ocr`); the full multimodal data suite passes with it (267 tests, CPU).

